PR #149, opened 2014-06-30
CLI testing framework

pjd-nu:
This is a really simple test framework for cli.py that uses monkey patch to substitute out for the URL functions and just validates that requests get made to the proper URLs with the proper data being sent. If people think it's a good idea then it can be cleaned up and added.


zenhack:
Closing; see #367.

PR #251, opened 2014-09-10
Implement SNMP-based Cisco driver

gsilvis:
This is based on Zhaoliang's old code.

DON'T merge this until it's tested--- @zenhack and I are working on getting that done soon.


zenhack:
@gsilvis, do we want to just close this? This is for the older cisco switch we have upstairs (*not* the nexus), which I don't know that we really care about anymore, and I don't think this is really a good use of our time.

gsilvis:
Yeah, I agree.  We have history of it if we need it.

PR #297, opened 2014-10-10
Authentication

gsilvis:
Authentication/authorization backends belong in haas/auth_mechanisms .  It's
possible that the API of that plugin will have to be expanded to cover what
the CLI needs to do.

WIP


henn:
This was excellent work, @gsilvis.

Given the new driver interface as well as the new plugin model, though, I think it would require a bit of refactoring and so I'm closing this out.

Perhaps @zenhack can use some of this as a basis for his work in #491.

PR #319, opened 2014-10-27
Added power_cycle_nodes script

ryabouza:
Simple script that takes a project name and calls the CLI command to power cycle all of its nodes.

This PR was inteded to address the Trello backlog task ```trivial head node script to talk to HaaS API, power cycle machines...```, although now that I read it again I'm not sure that I actually addressed it.  Were we looking for something to actually run *on* the headnode?

I had to do some funky stuff to use the CLI module to do this.  Although, I don't know if there is a better option.

### DIFF
--- a/scripts/power_cycle_nodes
+++ b/scripts/power_cycle_nodes
@@ -1,0 +1,27 @@
+#!/usr/bin/env python
+
+# Copyright 2014 Massachusetts Open Cloud Contributors
+#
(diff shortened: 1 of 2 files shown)


henn:
Thanks for your work on this, @ryabouza!

We are working on a python client library as part of the work by @SahilTikale  in #498 and should transition to using that. Closing this out.

PR #320, opened 2014-10-27
Added register_layout script

ryabouza:
This script registers the assets listed in site-layout.json to the HaaS db.

### DIFF
--- a/scripts/register_layout
+++ b/scripts/register_layout
@@ -1,0 +1,37 @@
+#!/usr/bin/env python
+
+# Copyright 2014 Massachusetts Open Cloud Contributors
+#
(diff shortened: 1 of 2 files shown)


zenhack:
@gsilvis, do we still want to merge this? We have another example node-registration script in examples, but given that the deployment tests still use the same file format, this might be nice to have.

If so it doesn't look too bitrotted, and I think @ryabouza fixed the few problems with it. we should test obviously, but...

henn:
Thanks for your work on this, @ryabouza!

We are working on a python client library as part of the work by @SahilTikale in #498 and should transition to using that. Closing this out.

Issue #321, opened 2014-10-27
Fill out metadata in setup.py


zenhack:
We're going to want to package the haas and put it on pypi (and probably make rpms/debs). To that end, we should fill out extra metadata in setup.py - things like author{,_email} (probably something like mass open cloud contributors/moc-technical@googlegroups.com), description, etc, as well as the categories field, which we can use to indicate supported python versions, development stage of the project (alpha...).

This applies to moc-rest as well; really wish github let issues span multiple repos.

zenhack:
I've got some work on this on my branch `pypi`. Useful links for RPMs:

* https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/6/html/Deployment_Guide/sec-Yum_Repository.html
* https://fedoraproject.org/wiki/How_to_create_an_RPM_package

We'll want to find a place to actually host an RPM repo for HaaS; I think it'll be easier in the long term to be doing deployments from RPMs than the pip stuff we're doing now.

naved001:
Fixed by https://github.com/CCI-MOC/hil/pull/534

PR #324, opened 2014-10-28
Fill out metadata in setup.py

ryabouza:
This is in response to issue #321


setup.py
```diff
@@ -14,13 +14,35 @@
...
+def read(filename):
+    return open(os.path.join(os.path.dirname(__file__), filename)).read()
 
 requirements = [str(r.req) for r in parse_requirements('requirements.txt')]
 
 setup(name='haas',
       version='1.0',
+      description='HaaS is a low-level tool for reserving physical machines'
+                  ' and connecting them via isolated networks.',
+      long_description=read('README.rst'),
```
abhi10: Maybe we should have another document which would be more informative to the user or just remove the reference to readme.rst file.

henn:
@SahilTikale - Could you take this over and submit it?

henn:
This has been picked up in #534. Thanks for your help!

Issue #355, opened 2014-11-12
Need a list_projects call for admins & devs


henn:
Admins may want to list projects so that they can track down allocated nodes without manually looking through the DB.

Developers may want to look for a random node name to run commands on or may have forgotten what project name they may have used.

zenhack:
How about we just grab 02673c8 and add some tests? The rest of the pull request might require a bit more scrutiny, but I think that commit looks pretty solid.

//cc @henn, @gsilvis, @lokI8

zenhack:
This was resolved by #424; closing.

PR #362, opened 2014-11-17
Add a list_projects command to the API/CLI

ryabouza:
In response to issue #355, add a new API call to list all projects
and the nodes connected to them.

@henn, is this what you were looking for?

### DIFF
--- a/haas/api.py
+++ b/haas/api.py
@@ -661,6 +661,22 @@ def port_detach_nic(port):
     db.commit()
 
 
+@rest_call('GET', '/projects')
+def list_projects():
+    """List all projects and nodes that belong to them.
+
+    Returns a JSON array of objects representing a list of projects.
+
+    Example:  '[{"nodes": ["node1", "node2"], "name": "project1"},
+                {"nodes": ["node3", "node4"], "name": "project2"}]'
+    """
+    db = model.Session()
+    projects = db.query(model.Project).all()
+    projects = [{'name': p.label, 'nodes': [n.label for n in p.nodes]}
+                for p in projects]
+    return json.dumps(projects)
+
+
 @rest_call('GET', '/free_nodes')
 def list_free_nodes():
     """List all nodes not in any project.
--- a/haas/cli.py
+++ b/haas/cli.py
@@ -313,6 +313,12 @@ def port_detach_nic(port):
     do_post(url)
 
 @cmd
+def list_projects():
+    """List all projects and their nodes"""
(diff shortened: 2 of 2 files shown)


haas/api.py
```diff
@@ -661,6 +661,22 @@ def port_detach_nic(port):
...
+    """List all projects and nodes that belong to them.
+
+    Returns a JSON array of objects representing a list of projects.
+
+    Example:  '[{"nodes": ["node1", "node2"], "name": "project1"},
+                {"nodes": ["node3", "node4"], "name": "project2"}]'
+    """
+    db = model.Session()
+    projects = db.query(model.Project).all()
+    projects = [{'name': p.label, 'nodes': [n.label for n in p.nodes]}
```
haydarkarrar: We have another API call for getting the nodes of the project. I think returning only list of projects is sufficient.

haas/api.py
```diff
@@ -661,6 +661,22 @@ def port_detach_nic(port):
...
 
 
+@rest_call('GET', '/projects')
+def list_projects():
+    """List all projects and nodes that belong to them.
+
+    Returns a JSON array of objects representing a list of projects.
+
+    Example:  '[{"nodes": ["node1", "node2"], "name": "project1"},
+                {"nodes": ["node3", "node4"], "name": "project2"}]'
```
haydarkarrar: I don't think so. We can write another API call to get headnode of the project.
Example:
`/projects/<project_name>/headnode`

zenhack:
Re: @haydarkarrar's suggestions, I don't really feel very strongly -- I think we should either do this "information dump" style and give a detailed view of the projects, or reduce it to just the names, but I don't really care which. My temptation is to leave the decision to whoever steps up and implements it, but I want to get others to weigh in first. @gsilvis, @lokI8, thoughts?

zenhack:
Ah, just noticed, this is beeing worked on as part of @haydarkarrar's pr (#405).

gsilvis:
The simplest thing to do is have 'list_projects' just give a list of projects (admin only, once we have auth) and 'show_project' show all the details of a project (can be run by someone with access to that project).  We need the latter call anyways.

henn:
I like the idea of doing a "dump" of everything: nets, headnodes, nodes, etc. This will allow external systems (schedulers, GUIs, etc) to get the status of everything without needing to make a bunch of separate calls.

These calls would add overhead and increase latency.

gsilvis:
Superseded

Issue #354, opened 2014-11-12
Client should be able to get endpoint from env variable


henn:
Right now, the client obtains the endpoint to talk to from the config file in the current directory.

It should be able to get it from either an environmental variable or possibly the config file too.

PR #365, opened 2014-11-20
Improve configuration abilities.

henn:
-CLI client no longer requires a config file if HAAS_ENDPOINT env is set (#354)

-Config file can now be configured by the caller of config.load() or through
the environmental variable HAAS_CONFIG. If neither of those are set, it will
still look for "haas.cfg" in the current directory

-Throw an exception if no config file was found by default, except for the CLI
client (to allow things like getting help without a config file).  Otherwise,
users get unhelpful errors like:
    ConfigParser.NoSectionError: No section: 'database'

-Logging is now configured after config.load() is called to ensure consistency
between WSGI-called processes and CLI-initiated. This enables debug output
when run as a WSGI process that can help with problems like #360.

-Add testing for the config management

-Add a config.reset() ability for testing. See doc note if you'd like to use it
in regular code.

-Subsequent calls to config.load() will be ignored. See config.reset() note if
you'd like to reconfigure at runtime.


haas/config.py
```diff
@@ -17,19 +17,93 @@
...
 
     This must be called once at program startup; no configuration options will
-    be available until then.
+    be available until then. Once called, further calls to load() will silently
+    be ignored.
 
-    If the configuration file is not available, this function will simply load
-    an empty configuration (i.e. one with no options).
+    If the configuration file is not available, throw an exception unless
+    requireConfigFile was set to False. We generally (for server configs) need
```
zenhack: re: our conversation today; the few things that need more than just the endpoint really *are* special; I actually like the idea of just listing them explicitly.

henn:
I am splitting this work between into two pulls: one for the HAAS_ENDPOINT env (#414) and another for the rest of it.

PR #367, opened 2014-11-20
Cli test

abhi10:
This Pull Request  builds up on the original request  for a test suite for CLI commands.
Also has a small fix for argument mismatch in haas/cli.py.

### DIFF
--- a/tests/unit/cli.py
+++ b/tests/unit/cli.py
@@ -1,0 +1,299 @@
+# Copyright 2013-2014 Massachusetts Open Cloud Contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the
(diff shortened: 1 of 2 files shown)


tests/unit/cli.py
```diff
@@ -0,0 +1,299 @@
...
+from haas.cli import cfg
+import unittest
+
+config.load()
+
+
+print "LIST OF SECTIONS --->",cfg.sections()
+
+# Here's the fake HTTP infrastructure. Use monkeypatch to stash the method, url,
+# and data and return a fake status code, then return them when needed.
```
abhi10: agree, incorporated the same.

zenhack:
I'm closing this and #149; No one is working on them, and we have reworking the whole cli on the TODO list, which is going to break things.

...It would be really nice to add something though.

PR #373, opened 2014-12-02
Remove unused import.

zenhack:
We're using some things from subprocess in haas.model, but we import
those individually.


henn:
+1. Ideally, for future tiny changes like this I would prefer it be combined to amortize the pull request overhead.

lokI8:
+1, this is fine.

PR #377, opened 2014-12-05
Clean up/refactor haas.test_common

zenhack:
The maze of stacked decorators is gone; the new implementation
is much simpler.

As a bonus, The deployment tests seem to pass with the code coverage
tool on -- I don't know what was causing it to fail in the first place.


tests/deployment/deployment.py
```diff
@@ -21,28 +21,17 @@
...
-    # status code.  The error presented is:
-
-    #       'import site' failed; use -v for traceback
-    #       Traceback (most recent call last):
-    #           File "/usr/bin/virt-clone", line 25, in <module>
-    #           import virtinst.CloneManager as clmgr
-    #       ImportError: No module named virtinst.CloneManager
-
-    # The tests can be run successfully by commenting out 'addopts =
-    # --cov-report term-missing --cov haas' in setup.cfg.
```
gsilvis: Is this bug fixed?

zenhack: Per the commit message, supposedly, but I'll double check.

zenhack:
@gsilvis, I can confirm that the deployment tests pass with coverage on -- on this branch as well as master.

Also, I just noticed this is targeted at devel, which isn't realy in use... to bad you can't retarget the pr. Maybe just do the code review and when it's signed off just push it to master? Opening a new pr is an option too I guess.

zenhack:
In case it's not obvious amongst the rest of the stuff in master that got pulled in, I added another commit: 8786d3b, which fixes a small amount of bit-rot.

PR #388, opened 2015-01-08
Update HACKING.rst

SahilTikale:
Added extra line explaining that the haas repository has to be downloaded prior to starting any developmental work.


zenhack:
I'm going to close this without merging. Having a local checkout of the source code you're hacking on ought to be a given.

PR #390, opened 2015-01-09
Install Instructions and Init Script

patcable:
Hi!

I was setting up HaaS from a CentOS core install and added in some packages I needed to have to be able to compile stuff. Also, I hated the idea of having the HaaS network server in /etc/rc.local so I put an init script together.

Thoughts? Questions?


INSTALL.rst
```diff
@@ -308,13 +308,16 @@ You should also set apache to start on boot::
 
   sudo chkconfig httpd on
 
-The networking server may be started by running::
+The networking server may be started by running the following as the HaaS user::
 
   haas serve_networks &
```
zenhack: If we're adding a proper init script, this should probably become `service haas-network start` (and will need to be run as root, for the usual reason).

examples/init-scripts/haas-network-el6
```diff
@@ -0,0 +1,217 @@
...
+    if [ -f "$lockfile" ]; then
+        log_action_msg "$service dead but subsys locked"
+        return 2
+    fi
+
+    log_action_msg "$service is stopped"
+    return 3
+}
+
+# sometimes sensu components need a moment to shutdown, so
```
zenhack: I'm willing to bet I know what you used as a reference... I'll give the script a more thourough review when I get the chance, but for now, would you double check that there aren't any more vestigial bits left over from the sensu script? We should avoid having logic in here that only make sense for other daemons.

INSTALL.rst
patcable: Yeah, I was figuring that on reboot the service would take over and this would allow the person going through the install to at least see if serve_networks barfed at all, but in reality the service definition would be best. I'll add that a bit later.

patcable:
I used [pleaserun](https://github.com/jordansissel/pleaserun) to generate a cleaner init script file, and updated the documentation. 

If you're interested in running it yourself to generate an init file, running `pleaserun --user haas_user --group haas_user --chdir /var/lib/haas --name haas-network -p sysv -v lsb-3.1 /usr/bin/haas serve_networks` generated the file in this commit.

zenhack:
Nice! didn't know about pleaserun.

So far nothing else in the codebase actually cares about the name of the user that HaaS runs as -- it's just a convention of the docs. This changes that by hardcoding `haas_user` in the init script. I'm a little uncomfortable with that, but at the very least the docs need to be modified to be very clear about this.

What do other people think? Hardcoding a user is something distro packages tend to do (and often a sysvinit script is maintained there), but I kinda want to maintain a bit of separation between packaging and "upstream," even though we're both.

I also might want to shorten the user name to just `haas` if it's going to be part of the code -- `haas_user` was chosen for pedegogical purposes; half the point was to hint at the idea that it's probably something you can change.

I'd also like to see the above `pleaserun` command documented somewhere, maybe in a file `examples/init-scripts/README.rst`

//cc @gsilvis @henn

patcable:
I personally agree that `haas` seems like a better choice than `haas_user`.

My own two cents: lots of packages create their own user. puppet comes to mind, libvirt comes to mind -- I don't think it's a big deal for haas to create a user called `haas` if it requires one.

I can create a `README.rst` under `init-scripts` for documentation purposes sometime this week.

zenhack:
Quoting Patrick Cable (2015-01-12 23:35:53)
>    I personally agree that haas seems like a better choice than haas_user.
> 
>    My own two cents: lots of packages create their own user. puppet comes
>    to mind, libvirt comes to mind -- I don't think it's a big deal for
>    haas to create a user called haas if it requires one.

Yeah, I think that's the right approach -- Let's wait for others to 
comment, but my thinking is: change the username in the docs and the 
scripts to `haas`, and make it clear from the documentation that that 
really does need to be the name of the user -- the issue is that as-is 
it would be reasonable (and up until now correct) to infer from the docs 
that the username could be anything. We should avoid that pitfall. The 
rest is bikeshed.

>    I can create a README.rst under init-scripts for documentation 
>    purposes
>    sometime this week.

Great, I look forward to it.

patcable:
The user *could* be anything, but I added in some text to mention that the user `haas` is used throughout the documentation and elsewhere and it's easiest to just use `haas`.

zenhack:
I made one in-line comment, other than that this *looks* good to me. We still need to test it ourselves, and come up with a CLA for you to sign. @pns005, how close are we to having a CLA?

@gsilvis, @henn, @lokI8, any other comments?

patcable:
I'm going to close this as I need to work through some internal $work related things re: the CLA. I'll re-open soon.

Issue #394, opened 2015-02-19
Merge moc-rest into haas repo


henn:
Currently, installing HaaS requires a separate install step for the moc-rest module. There is some agreement within the team that merging moc-rest back into HaaS would simplify the installation and maintenance process and make certain development tasks (especially decorators) easier to implement.

Here are some links that might help with maintaining the history (if that's important):
* http://jasonkarns.com/blog/merge-two-git-repositories-into-one/
* https://www.kernel.org/pub/software/scm/git/docs/git-fast-import.html
* http://bpeirce.me/moving-one-git-repository-into-another.html

zenhack:
This has been done, closing.

PR #398, opened 2015-02-24
Merge moc-rest back into HaaS

zenhack:
Addresses #394.

The change isn't as involved as it looks, it's just that the diff includes the whole moc-rest history as well. commit 595b173 is the first one that's of any interest to this pull request.


henn:
+1, assuming someone's tested this with at least the HACKER instructions to ensure it works cleanly.

zenhack:
@henn, I double checked and the instructions work fine.

lokI8:
+1, looks good. merging

PR #403, opened 2015-03-24
Restapi docfix

zenhack:
We were still talking about form data -- it's been quite a while since we switched to JSON.

### DIFF
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -1,6 +1,6 @@
 
-* `{foo=bar, baz=quux}` denotes a JSON object (in the body of the
-  request).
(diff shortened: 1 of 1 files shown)


gsilvis:
Oh hey, we actually use JSON now!  +1

zenhack:
@henn, @lokI8, This also needs a second review. This one should be pretty quick.

zenhack:
Oh, worth noting: I also started expanding the documentation for other stuff in `rest_api.md`. I'd like to move in a direction where this file is a full reference, whereas now it documents arguments and urls and such, but you have to know the semantics already.

lokI8:
+1, merging. looks good

PR #405, opened 2015-03-24
Added project list and project's headnode API/CLI calls

haydarkarrar:


zenhack:
Thanks for taking the initiative on this!

A couple points:

1. You've got some tab characters in there; please use only spaces for indentation.
2. A project can have more than one headnode -- the fact that the variable name is singular is a bit confusing and probably left over from when this wasn't the case. My expectation is that `project.headnode` is going to give you a list, which isn't going to do what you want, though I've not played with that part of the code in a while. Which brings me to:
3. Please write some unit tests.

Thanks!

zenhack:
Incidentally, given the many-to-one headnode->project relationship, it probably makes more sense to have a list_project_headnodes and use the existing show_headnode for a detail view.

gsilvis:
Superseded

PR #407, opened 2015-03-30
Fix typo in headnode_detach_hnic

henn:


zenhack:
Nice catch! +1.

One of these days we should write some unit tests for the cli module...

zenhack:
@gsilvis, @lokI8, could one of you do a second review? Should be pretty quick.

gsilvis:
I've actually even SEEN this before.  Bluh.  +1

Issue #318, opened 2014-10-27
Make sleep time for async loop tunable


henn:
Per review comments on #296, the sleep() time in serve_networks() should be tunable. This should also be documented.

Izhmash:
How tunable should the sleep time be? My first thought is to pass the time as a parameter in serve_networks() but that might not be the right direction.

Izhmash:
@henn Should the solution from #323 that was merged into the devel branch be implemented in master?  It seems that is the accepted solution to this issue.

henn:
@lzhmash Nice find, though I think it only partially addresses the issue. The solution will need to check for configuration from the config file (see uses of `cfg` elsewhere in the code), then pass it to serve_networks() using the solution you found if the configurable was set.

naved001:
Fixed by #785

Issue #47, opened 2014-04-16
Need to be able to allow multiple VLANs/networks for a given node/nic/port


zenhack:
If we're going to be using the HaaS to deploy openstack on bare metal, need to be able to configure the network in such a way that the nodes can use multiple VLANs to communicate with eachother - Neutron makes use of this in one of the standard configurations.

The way to go about this is to set the ports to trunking, rather than access mode.

Probably step 1 is to convert the existing implementation from setting access mode (with a given VLAN), to trunking mode, with that VLAN as native (and no other VLANs assigned). From there we can start worrying about how we should expose the ability to added tagged VLANs.

henn:
Reopening, as we have hardware now that will require this feature.

henn:
FYI - the implementation will need to take into consideration [VLAN hopping attacks](https://en.wikipedia.org/wiki/VLAN_hopping).

henn:
Addressed by #455

PR #408, opened 2015-04-02
Vlan trunking

zenhack:
I'm starting to work on the vlan trunking. It's not ready yet, but the docs are there and I'd like commentary on the design.

//cc @gsilvis


docs/networks.md
```diff
@@ -40,3 +40,26 @@ Here are the illegal ones:
...
+VLANs.
+
+The common abstraction for this is called a "channel." A channel is a
+driver-specific identifier which describes how traffic will be tagged on
+the network. It is used by HaaS in the following ways:
+
+- The `node_connect_network` api call accepts an optional channel argument,
+  which determines how the network will be associated with the given
+  port. The headnode equivalents may support this eventually, but do not
+  as of yet.
```
gsilvis: Do we want individual drivers to be able to support or not support this, as they are able?

tests/unit/api.py
```diff
@@ -1335,7 +1335,7 @@ def test_show_headnode(self, db):
             'project': 'anvil-nextgen',
             'hnics': [
                 'eth0',
-                'wlan0',
+        'wlan0',
```
gsilvis: hm?

zenhack: I have no idea. I will fix.

docs/networks.md
```diff
@@ -40,3 +40,26 @@ Here are the illegal ones:
...
+
+  If this argument is left out, a sensible default will be chosen.
```
gsilvis: I would say that the default should be to put the network traffic on it, unmodified.  Being sure that you can simply listen to the ethernet cable for that traffic, whatever the driver is, seems very useful to me.

docs/networks.md
gsilvis: Clarification:  I'm talking about the special case of headnodes

docs/networks.md
zenhack: Quoting George Silvis, III (2015-04-02 13:27:31)
 
>    I would say that the default should be to put the network traffic on
>    it, unmodified. Being sure that you can simply listen to the ethernet
>    cable for that traffic, whatever the driver is, seems very useful to
>    me.

Yeah, I think "native" or equivalent is almost always going to be the 
right default. I want to keep this abstract though.

zenhack:
@gsilvis, I speced out the changes for `node_detach_network`. I think this will make it possible to find the right arguments to `apply_networking` by diffing this with the networks attached, but perhaps we should just change apply_networking to take the diffs directly, with None being delete all. What do you think?

docs/networks.md
henn: @gsilvis: I think we indeed should have an ability for network drivers to either support or not support channels. We should also include an ability for network drivers to express which features they support and have filed #410 to address that.

henn: @gsilvis: re headnodes, I think it's fine not to support channels for now since one can add unlimited hnics. Eventually, it might be nice to expose channels to headnodes, though I think before we talk about that we need to figure out what our long-term HN strategy.

docs/rest_api.md
```diff
@@ -97,3 +108,10 @@ Full Api spec:
             "nics": ["ipmi", "pxe", "public", ...]
         }
     [GET] /headnode/<headnode>
+
+    show_network <network> ->
+        {
+            "name": "my-net",
+            "access": "my-proj", # optional; absence means a public network.
+            "creator": "my-proj" # either a project or "admin"
```
henn: Is there a compelling reason to include "creator"? Once something is exported, it's difficult to take back later.

Perhaps additional information should be added for admins? (like VLAN id)

henn:
What would you say are some valid options for the channel argument? With VLANs at least, HaaS will be the one who picks the VLAN IDs for networks.

zenhack:
@henn, the idea was, once the rest of this was implemented, to have the show_network call contain a list of the valid channels. for a VLAN network this would just be something like `["native", "253"]` if, for example, the driver ended up choosing vlan 253 for the network.

We talked a bit about this this afternoon, and it needs work. In particular, for some drivers (e.g. vxlans), the list of valid channels may be effectively infinite, so listing them all is out of the question. Iirc, @patcable mentioned that he has a juniper switch that let's you use different vlan numbers on different ports -- my understanding is in that case each network would end up basically having the entire vlan space in the list.

henn:
Since all we support right now is VLANs, would it make sense to leaves channels till later and go forward with the rest of your proposed changes?

`node_connect_network`, instead of taking a "channels" argument could have a "native" flag. Users could figure out the VLAN id from `show_network` (or `show_project` :)

zenhack:
I wrote some docs to reflect the channels proposal as discussed today.

For posterity, @henn suggested an alternative API, which was very similar, but dealt with things at a less abstract level.

Differences:

* subtitute channel with VLAN id and don't have the separate discussion of driver-specific semantics.
* possibly disallow selection of a specific VLAN id, and just say "native" or "tagged."

@henn also suggested (in the context of the channel based driver), having a legal channels field on the nic, with similar semantics. He argued that we may run into situations where a node is on a switch which does not support tagging, in which case regardless of what the network thinks is okay, attaching to e.g. `vlan/260` isn't going to work.

I don't think the nic channel list is something that deeply affects the implementation now, or disturbs the rest of the API proposal significantly. I have no objection to adding it -- worst case, we never get any hardware with this restriction, and the driver can just *always* return `vlan/*`.

Did I miss anyting?

zenhack:
Oh, worth noting: I also started expanding the documentation for other stuff in rest_api.md. I'd like to move in a direction where this file is a full reference, whereas now it documents arguments and urls and such, but you have to know the semantics already.

henn:
This addresses #47.

@zenhack: Before merging the code changes, it might be good to verify the security of our implementation against the known VLAN hopping attacks documented on [this wikipedia page](https://en.wikipedia.org/wiki/VLAN_hopping).
Perhaps that would make a good project for one of the students?

zenhack:
@henn, probably a good intern task, if they've got the bandwidth. The double tagging exploit seems like something that depends on a bug in the switch though, so it's going to be hardware dependant, and I don't know how we go about working around it if it exists -- not having native VLANs isn't really an option.

zenhack:
Update: the double header attack in the wiki article makes more sense to me now; it looks like it can happen if one switch is tagged on the shared port and the other isn't. In that case, the mitigation measures for both of these are things the haas doesn't actually touch itself; the sysadmin is currently responsible for making that change. We should verify that we aren't perturbing the negotiation bits with what we're doing, and perhaps document this as a concern when configuring the network, but this is out of the scope of what HaaS actually does itself.

At some point we may want the complex vlan driver to know about the connections between switches, beyond just trusting that they exist, but for now, that's something that's done manually.

zenhack:
What do people think about the contents of `show_network`'s output? Do we want to be including ownership/access information?

zenhack:
This is officially ready for review. I'll go over the structure with y'all on Tuesday, as planned.

henn:
Since we are rewriting the async network code, please address #318

henn:
Also, on the network action queue, we still have the problem that a large number of requests on a busy server could perpetually not complete because of DB transactions on related items failing to commit.

Imagine there being a busy server, with a request being added every second. Every time the server tries to process the queue, it takes 3 seconds. When the server is done, it will try to execute db.commit(), which will fail because at the top it used .all(), which is now stale. Each successive time it runs will take even longer and be more likely to fail.

This could be solved if we handled one transaction at a time instead of batching (`.first()` instead of `.all()` combined with an ordering). Handling one at a time would also have the advantage of keeping an error on one of the operations from blocking all of the others.

henn:
FYI - we decided that we need to have an architectural discussion over what we want the serialization guarantees of HaaS to be

zenhack:
@gsilvis, @lokI8 and I talked about the serialization issues a bit on IRC today. 70c2aad2 was pretty clearly indented to cause strict ordering (though see #420). I don't have strong opinions about the way we should do this, but I think changing it is orthoginal to this pull request, and given that @henn was clearly correct about the intent of the original implementation, I think we should not disturb that. I think we should also fix #420 while we're at it.

zenhack:
It's worth noting that restoring the old semantics is easy --- we just add a modification date to the attachments table and order by that.

gsilvis:
I think #420 should be fixed independently.  Because of that, I would argue that this PR should add the modification date, so it doesn't change the process of fixing #420---it keeps the same "semantics" of ordering the database read, and then throwing the order out.

zenhack:
If others agree, I'll submit a separate pr that fixes #420, then pull it back into this branch once it's merged.

zenhack:
While I'm thinking of it --- I seem to recall having a conversation with @gsilvis about relaxing the one-action-per-nic to one-action-per-channel or somesuch, and deciding that there was enough subtlty to defer it. I think the original reason for restricting to one action per nic was that otherwise we'd either have to look at the whole queue to decide if the action was compatible with all of the pending actions (which seemed a bit overly tricky) or do the saftey checks at application time, rather than request time.

zenhack:
@henn, @gsilvis and I had a discussion about the serialization issue. We agreed to re-separate the newtork operations and attachments, and keep serializing things, with the same basic strategy.

zenhack:
Just a note, I pulled the networking actions back out into their own table like we talked about. I still need to run the deployment tests and fix any bugs they uncover, which I'll do tomorrow.

zenhack:
re: #420, I'd like to get this merged first, since it's likely going to have merge conflicts with the fix for that.

henn:
Regarding what to do if the user doesn't indicate a specific channel (ie - connect_network/node_connect_network), would it make more sense to explicitly say that we will only connect something as native if they don't specify a channel?

I think that would be closer to expected behavior for a user not trying to user channels.

docs/apidesc.md
```diff
@@ -66,8 +66,8 @@ one exception is NICs, where the label is unique only on a per-node basis.
     project_connect_node        <project_label> <node_label>
     project_detach_node         <project_label> <node_label>
  
-    node_connect_network        <node_label> <nic_label> <network_label>
-    node_detach_network         <node_label> <nic_label>
+    node_connect_network        <node_label> <nic_label> <network_label> [ <channel> ]
+    node_detach_network         <node_label> <nic_label> <network_label>
```
henn: Could we make network_label optional here? In the (common?) case of not using channels, it would just detach the native network and be close to how many users would think of it. It could throw an error if more than one network were connected.

zenhack: I'm not totally sure how I feel about this. @gsilvis, what do you think?

zenhack:
@henn, per the changes to rest_api.md, we are defaulting to native/untagged in `node_connect_network`, and this is documented.

docs/rest_api.md
```diff
@@ -2,12 +2,139 @@
...
+    {
+        "network": <network>,
+        "channel": <channel> (Optional)
+    }
+
+Connect the network named `<network>` to `<nic>` on `<channel>`.
+
+`<channel>` should be a legal channel identifier specified by the output 
+of `show_network`, above. If `<channel>` is omitted, the driver will choose
+a default, typically some form of "untagged."
```
henn: Could we make the default behavior explicit (like "untagged")?

For something with potential security implications, I'd like to leave the default defined rather than implied.

docs/rest_api.md
```diff
@@ -2,12 +2,139 @@
...
+
+Possible errors:
+
+* 409, if:
+  * The current project does not control `<node>`.
+  * The current project does not have access to `<network>`.
+  * There is already a pending network operation on `<nic>`.
+  * `<network>` is already attached to `<nic>` (possibly on a different channel).
+  * The channel identifier is not legal for this network.
+
```
henn: + the user does not have access to the project that the node is assigned to

examples/deployment.cfg
```diff
@@ -23,9 +23,16 @@ libvirt_endpoint = qemu:///system
...
+# A comma separated list of ports which may be manually configured to be trunked
+# on HaaS's vlans. The deployment tests need this information when verifying
+# that their configuration changes had the expected effect.
+#
+# Ports may be set up manually this way for a few different reasons, including:
+#
+# * The port is used to give headnodes access to the network
+# * The port is used to bridge networks to another switch (as used by the
+#   complex vlan driver).
+trunk_ports = gi1/0/4, gi1/0/5
```
henn: What is this used for anymore?

haas/api.py
```diff
@@ -120,11 +122,14 @@ def project_detach_node(project, node):
     node = _must_find(db, model.Node, node)
     if node not in project.nodes:
         raise NotFoundError("Node not in project")
+    attachment = db.query(model.NetworkAttachment)\
+        .filter(model.Nic.owner == node,
+                model.NetworkAttachment.nic_id == model.Nic.id).first()
```
henn: Could we use exists() or count() here?

docs/rest_api.md
zenhack: Quoting henn (2015-05-22 17:31:07)

> > +`<channel>` should be a legal channel identifier specified by the 
> > output
> > +of `show_network`, above. If `<channel>` is omitted, the driver will choose
> > +a default, typically some form of "untagged."
> 
>    Could we make the default behavior explicit (like "untagged")?

We can only make it so explicit while keeping the docs independent of 
the particulars of a given channel type. This *is* explicitly document 
as being `vlan/native` for the vlan case further down in the file.

I could change the wording to make it sound like less of a suggestion I 
suppose; is 'which should be some form of "untagged"' any better?

docs/rest_api.md
zenhack: Quoting henn (2015-05-22 17:32:25)

>      * the user does not have access to the project that the node is
>        assigned to

Users *are not* part of the core HaaS; they're auth-backend specific (in 
theory; we need to actually add auth and move the user model into one of 
the backends). The notion of "current project" implies you've already 
somehow proven you're authorized to access it.

examples/deployment.cfg
zenhack: Quoting henn (2015-05-22 17:36:14)

> > +# * The port is used to give headnodes access to the network
> > +# * The port is used to bridge networks to another switch (as used by the
> > +#   complex vlan driver).
> > +trunk_ports = gi1/0/4, gi1/0/5
> 
>    What is this used for anymore?

Good call, it isn't --- got obsoleted when I changed the way we were 
querying the switch. I'll take it out.

haas/api.py
zenhack: Quoting henn (2015-05-22 19:07:17)

>    Could we use exists() or count() here?

I don't see why not. I'll change it.

docs/apidesc.md
gsilvis: Making it optional could help make it backwards-compatible.  Some ways of doing this are:  remove the native VLAN; remove /everything/; remove the native VLAN if it's the only one [as @henn suggested].  I prefer the first---it's most parallel with how we're doing node_connect_network.

zenhack: Quoting George Silvis, III (2015-05-23 00:44:43)

>    Making it optional could help make it backwards-compatible. Some 
>    ways
>    of doing this are: remove the native VLAN; remove /everything/; remove
>    the native VLAN if it's the only one [as [2]@henn suggested]. I prefer
>    the first---it's most parallel with how we're doing
>    node_connect_network.

I think I agree -- I suppose if a user is blindly ignoring the optional 
field, this works as expected; they'll never set any other channel. I'm 
fine with making it optional.

docs/rest_api.md
henn: Interesting point. So whomever ends up adding authentication will need to update the docs with the appropriate status codes.

henn:
Will the null_vlan.py driver still work with get_default_channel()/other channel stuff?

zenhack:
Quoting henn (2015-05-28 19:12:20)

>    Will the null_vlan.py driver still work with
>    get_default_channel()/other channel stuff?

It will; the interface hasn't changed enough to disturb what it actually 
does (which isn't too surprising, since that's not much).

henn:
Re: null_vlan.py, I was wondering if this code (from `node_connect_network()`) would throw an exception if a channel weren't passed to the null_vlan driver:

```python
driver = _get_driver()
if channel is None:
    channel = driver.get_default_channel(db)
```

This might be a case where expressing driver capabilities (ala #410) could come in handy.

zenhack:
That's why the if statement is there --- the driver needs a channel passed, but it's optional in the api call, so if it's not provided, we fill it in with the driver default.

re capabilities: A driver which wants to support only VLANs with no trunking support could always just accept only one possible channel (denoting untagged). We do need some way to deal with different underlying network types, but that's not dealt with with this patch. Have a look at the driver refactoring proposal; it touches on this.

henn:
@zenhack: Are you saying that the code above wouldn't try to call the null_vlan driver's `get_default_channel()` if the user didn't pass in a channel?

zenhack:
@henn the opposite; if they supply a channel it will just use that instead of calling the method.

docs/rest_api.md
zenhack: Correct -- probably will be a blanket 401 Unauthorized for every call, but yeah, we'll have to document it when we add it.

henn:
@zenhack: For this PR, I think we either need to implement [driver capabilities](https://github.com/CCI-MOC/haas/issues/410) ala:
```python
driver = _get_driver()
if 'CHANNELS' in driver.capabilities and channel is None:
    channel = driver.get_default_channel(db)
```
or implement all of the necessary channel interfaces in every driver.

Otherwise, the server will crash when doing operations like trying to connect a network and not passing an explicit channel when using the null_vlan.py driver.

I'd be happy to take a first pass at the capabilities mechanism since it seems to be popping up a lot.

zenhack:
Quoting henn (2015-05-29 11:23:48)
>    or implement all of the necessary channel interfaces in every 
>    driver.

I did this (with the exception of the nexus, which is done on a 
different branch, and is its own can of worms).  Please actually read 
the code carefully.  The null driver doesn't need to change because even 
after the interface changes, apply_networking still takes a dict (the 
values are different), and that's as closely as the null driver looks at 
it (it doesn't actually *do* anything after all).

zenhack:
superseded by #455, closing.

PR #421, opened 2015-05-04
New dev example

kamfonik:
an example haas.cfg file for developers that includes the 'recursive' section


kamfonik:
oops, wrong place again, sorry everyone

PR #423, opened 2015-05-06
List project headnodes

zenhack:
We said a while ago that we wanted this, and now one of our early adopters at MIT has specifically asked about it.


gsilvis:
This looks good to me +1

lokI8:
+1

Issue #389, opened 2015-01-09
HaaS CLI should have a 'version' command


gsilvis:
It would print the version.

SahilTikale:
A fix for this is generated under #425

zenhack:
@SahilTikale, I'm assigning this to you, since you've already got a patch out for it. Apoorve asked me to give it to him, did you two coordinate something I don't know about? In any case, this should make it a little more clear what's going on.

SahilTikale:
No, we did not coordinate anything related to this call. I am working on this and it is fine that you have assigned it to me. In future if I and Apoorve work on any part of HaaS together we will update you accordingly.

naved001:
@shwsun and I would like to work on this issue.

SahilTikale:
It requires changing the way we have implemented cli.
Wait for sometime. Once client library is over, you can start working on
this.

--Sahil

shwsun:
@SahilTikale Yes, I just talked to Naved this morning, he can't start to work on this issue unless #617 and the click enhancement is done.

xuhang57:
@naved001 I can work on this. Currently, HIL version is 0.2 right? Should I hardcode the version or there is a better way of handling this through automation?

xuhang57:
@SahilTikale Can I take this? ;)

naved001:
I am actually not sure myself how to do this because the HIL CLI and API server would be running on different machines. Which would mean getting the version from the hil server through an API? or would the CLI fetch it from wherever it was installed. What if we are using different  versions of the hil for cli and api server.

I don't have a clear thought regarding this. But you could come up with some proposal.

Also you wouldn't hardcode it, you would probably fetch it from `setup.py` 
This PR might be relevant. https://github.com/CCI-MOC/hil/pull/905

xuhang57:
hmm, let's talk in person first then. I thought the cli and api are always come together.Or at least, the cli is calling the api so maybe I am wrong.

so we need to set the definition of the version of HIL cli ? And is that the same as the version release number on github?

The only way that i know is to hardcode the release version in the code base. Whenever you do a version upgrade, you need to make sure it is changed.

Or, if hil is a python package(I know it is), we can manage the version through pip so it doesn't have to be hardcoded. But we need to play it with pip and set up the version. Do you have the access to do that? 

@zenhack any thoughts?

zenhack:
Either or both of the following would be reasonable:

[...]

Issue #316, opened 2014-10-27
pip.req trickery breaks python setup.py bdist_rpm


zenhack:
I was playing with packaging our stuff, and hit this; it looks like it's trying to call setup.py recursively from somewhere other than the root of the tree, and since that means requirements.txt isn't there, the trick we're pulling to avoid duplicating the dependency list between requirements.txt and setup.py causes it to fail.

I wish we could share bugs between repos; this affects moc-rest too.

zenhack:
We don't have requirements.txt anymore, so this is no longer relevant; I'm closing.

PR #425, opened 2015-05-14
Add version CLI call

SahilTikale:
Created an API and CLI call for fetching current version of HaaS.


SahilTikale:
This is a fix for issue #389

zenhack:
I think we want the version command to just spit out the version based on the cli itself, rather than adding an API call. I actually think having an API is a bad idea -- the most we want to expose there is the verison of the API, but not the software.

setup.py
```diff
@@ -13,9 +13,16 @@
 # governing permissions and limitations under the License.
 
 from setuptools import setup, find_packages
+import os
+
+v = open('haas/version.py')
+exec(v.read())
```
zenhack: Two things:

1. We should be careful about how we're locating the file; some of the setup.py operations run things from other directories, so using a relative path could be perilous. See also #316 (which I *thought* was fixed...). I don't fully understand what needs to happen for this to be safe.
2. is the `exec` necessary? Is there a reason we couldn't just import it?

@gsilvis, pulling this from one place is going to necessarily involve a bit of trickery, can you weigh in on the right way to do this?

gsilvis: @SahilTikale spent some time looking up how people generally do this in Python, and he says that this is actually one of the common methods.  @SahilTikale, could you link some sources you found?

SahilTikale: Response to "1. We should be careful about how we're locating the file;"
I looked around on the Internet for best practices and realized that the most practiced method is to have setup.py in the root folder of the project. Hence, the relative path should work just fine. 
https://packaging.python.org/en/latest/distributing.html#setup-py

Response to "2. is the exec necessary? Is there a reason we couldn't just import it?"
Importing the package in the setup.py may invoke a race condition and circular dependency both of which can be a problem for new users trying to install the package. Circular dependency arises because the package will try to import all other modules, many of which may not have been installed. 

I came across some posts on stackoverflow, like this one:
http://stackoverflow.com/questions/2058802/how-can-i-get-the-version-defined-in-setup-py-setuptools-in-my-package
Thus to avoid the potential issues of race condition and also make the code compatible across python 2.X and 3.X I implemented the solution as submitted.

gsilvis:
We should discuss the meta-issue (do we even want this API call) in person

zenhack:
Looks good to me, +1.

henn:
How would folks feel about making it `haas --version` instead of `haas version`?

The former I would expect to give me the local version while from the latter I would expect the server's.

zenhack:
Quoting henn (2015-05-28 19:17:21)

>    How would folks feel about making it haas --version instead of haas
>    version?
> 
>    The former I would expect to give me the local version while from the
>    latter I would expect the server's.

Fine conceptually, but we'd have to do some refactoring to make that 
work. Really we ought to move over the cli to something based on the 
`argparse` module. There are a lot of limitations with the stuff I 
rigged up way back (when I was still learning the language), and this is 
a solved problem.

My take: leave it as `version` for now, and put it on a bucket list of 
things to do when we refactor the cli.

zenhack:
Going to put waiting on reviewer back on this; this is more something for us to discuss than for @SahilTikale to fix.

zenhack:
Un+1ing this, since it's seen some further changes that we should talk about.

SahilTikale:
This also fixes #431

zenhack:
The right approach here is probably to use sub-commands:

https://docs.python.org/dev/library/argparse.html#sub-commands

The current approach has the kind of awkwardness I was worried about with an incremental implementation --- particularly with the help message, which looks awkward.

zenhack:
Also, try to keep the first line on your commit messages short --- you'll notice that github cut it off in a weird spot. See also:

http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

haas/cli.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2013-2014 Massachusetts Open Cloud Contributors
+''# Copyright 2013-2014 Massachusetts Open Cloud Contributors
```
henn: ??

henn:
Should we close this?

SahilTikale:
I guess we should close this. Too much has changed since then. And also we wish to do CLI differently than using argparse.

PR #426, opened 2015-05-14
Fix a few bugs in the nexus driver.

zenhack:
Clearly this hadn't been tested, but the problems were fairly shallow.

It's also missing an implementation of `get_switch_vlans`, but since the trunking support removes that function, I'm not going to bother. I'll submit a separate pr (based off of the vlan trunking stuff) that converts this over to trunking.


henn:
Assuming you've tested this against a real nexus, +1

zenhack:
I have indeed.

zenhack:
The driver is totally rewritten in #455, I'm just going to close this.

PR #429, opened 2015-05-18
port number from command line

apoorvemohan:
changes for a port being passed as a command line argument. please review.

### DIFF
--- a/haas/cli.py
+++ b/haas/cli.py
@@ -97,7 +97,7 @@ def serve(port):
         port = schema.And(schema.Use(int), lambda n: MIN_PORT_NUMBER <= n <= MAX_PORT_NUMBER).validate(port)
     except schema.SchemaError:
 	sys.exit('Error: Invaid port. Must be in the range 1-65535.')
-    except as e:
+    except Exception as e:
 	sys.exit('Unxpected Error!!! \n %s' % e)
 
     """Start the HaaS API server"""
--- a/haas/cli.py
+++ b/haas/cli.py
@@ -27,6 +27,8 @@ from functools import wraps
 
 command_dict = {}
 usage_dict = {}
+MIN_PORT_NUMBER = 0
+MAX_PORT_NUMBER = 65525
 
 def cmd(f):
     """A decorator for CLI commands.
@@ -90,6 +92,17 @@ def do_delete(url):
 
 @cmd
 def serve(port):
+    isNotValidPort = False
+    if port.isdigit():
+        port = int(port)
+        if (port < MIN_PORT_NUMBER) or (port > MAX_PORT_NUMBER):
+            isNotValidPort = True
+    else:
+	isNotValidPort = True
+
+    if isNotValidPort:
+	sys.exit('Error: Invaid port. Must be in the range 0-65535.')
+
     """Start the HaaS API server"""
     if cfg.has_option('devel', 'debug'):
         debug = cfg.getboolean('devel', 'debug')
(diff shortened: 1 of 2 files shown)


zenhack:
Looks good to me, +1.

Ideally we'd have it choose a default, and be overridable, but that's going to require some refactoring of the cli tool.

@gsilvis, @lokI8, @henn, can one of you do a second review?

apoorvemohan:
exception handling for invalid port number committed

haas/cli.py
```diff
@@ -27,6 +27,8 @@
 
 command_dict = {}
 usage_dict = {}
+MIN_PORT_NUMBER = 0
```
zenhack: Port 0 has special meaning:

http://lxr.free-electrons.com/source/net/ipv4/inet_connection_sock.c?v=3.18#L89

Start with port 1.

haas/cli.py
```diff
@@ -27,6 +27,8 @@
...
+MAX_PORT_NUMBER = 65525
```
zenhack: How about `2**16 - 1`? makes it a bit more clear where the number is coming from.

haas/cli.py
```diff
@@ -89,7 +91,18 @@ def do_delete(url):
...
 @cmd
-def serve():
+def serve(port):
+    isNotValidPort = False
+    if port.isdigit():
+        port = int(port)
+        if (port < MIN_PORT_NUMBER) or (port > MAX_PORT_NUMBER):
+            isNotValidPort = True
+    else:
+	isNotValidPort = True
```
zenhack: Check out the schema library (we're using it elsehwere in the code) --- I bet you can get this down to a (readable) one-liner.

haas/cli.py
zenhack: Ah, had another look. You're using strictly less than in the check so this works actually, but in my mind "MIN" implies an inclusive range, so the name is a bit confusing.

apoorvemohan:
changes committed

haas/cli.py
```diff
@@ -89,7 +92,12 @@ def do_delete(url):
     return check_status_code(requests.delete(url))
 
 @cmd
-def serve():
+def serve(port):
+    try:
+        port = schema.And(schema.Use(int), lambda n: MIN_PORT_NUMBER <= n <= MAX_PORT_NUMBER).validate(port)
+    except:
```
zenhack: Always specify the actual error you're expecting --- in this case `schema.SchemaError`. If some error you're not expecting occurs, you want to hear about it (and not necessarily handle it the same way).

haas/cli.py
```diff
@@ -89,7 +92,14 @@ def do_delete(url):
     return check_status_code(requests.delete(url))
 
 @cmd
-def serve():
+def serve(port):
+    try:
+        port = schema.And(schema.Use(int), lambda n: MIN_PORT_NUMBER <= n <= MAX_PORT_NUMBER).validate(port)
+    except schema.SchemaError:
+	sys.exit('Error: Invaid port. Must be in the range 1-65535.')
+    except as e:
```
zenhack: You need a type here; `except Exception as e` is probably what you right. As it stands, this is a syntax error.

zenhack:
Okay, this seems to be working. @gsilvis, this *does* break backwards compatibility, but I'd like to just not worry about it --- it's in a dev tool and I don't want to imply any promises unless we need to. Thoughts? I'm otherwise happy with this, +1

lokI8:
+1, merging. Ian's right that we need optional arguments in the CLI

PR #430, opened 2015-05-28
Extensions

zenhack:
We should talk about when/how we want to merge the refactoring work I'm doing, but this is pretty self-contained.

This is based on the proposal I wrote at <https://github.com/CCI-MOC/haas/wiki/network-driver-model-refactor>

This also includes some slight cleanup of some of the test support code.

### DIFF
--- a/docs/extensions.rst
+++ b/docs/extensions.rst
@@ -1,0 +1,20 @@
+HaaS supports a simple extension mechanism, to allow external plugins
+to implement things we don't want in the HaaS core. One obvious example
+of this is drivers.
+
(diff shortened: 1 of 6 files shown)


gsilvis:
The code looks good.  I think we should wait to merge until we see the code that depends on it, IMO.

zenhack:
After talking to @gsilvis, I just created a branch on the main repo called `network-refactor`, which I'll send prs related to this refactoring, so we can review them incrementally before the whole thing's ready. I'm going to close this and open another pr directed at that branch.

PR #434, opened 2015-06-08
Extensions

zenhack:
Superceeds #430.


gsilvis:
This looks good right now.  I still would like to see it used in practice, to check that it's really the right API for this.

gsilvis:
Just noticed/was told that this is not aimed at master.  +1, then

haas/test_common.py
```diff
@@ -45,20 +45,36 @@ def releaseDB(db):
...
+    ``config_dict`` should be a dictionary mapping section names (strings)
+    to dictionaries mapping option names within a section (again, strings)
+    to their values.
+    """
+    config_clear()
+    for section in config_dict.keys():
+        cfg.add_section(section)
+        for option in config_dict[section].keys():
+            cfg.set(section, option, config_dict[section][option])
+
```
gsilvis: This function seems almost the same as ``config_merge`` in #437 .  Same purpose, almost the same implementation.  Should the code be shared?

zenhack: Ack, it totally should be shared. I probably wrote this one first and forgot about it --- just dealt with the merge on another branch, now I know why it was so big...

This one isn't going into master anyway; if we decide what to do with #437, I can merge it back into this.

gsilvis:
I'm assuming you don't want to merge this until we get #437 merged?

zenhack:
@gsilvis probably makes sense, yeah.

zenhack:
subsumed by #455; closing

Issue #428, opened 2015-05-18
Get HaaS working with PostgreSQL


zenhack:
We need support for a production ready DBMS (sqlite isn't even thread-safe). Postgres should be pretty easy to get working; *in theory* it should work already, but is untested so there are probably going to be a few snags to fix.

apoorvemohan:
I have completed the changes but I need to test it. I see a tests directory in the haas code. Can anyone  guide me how to use them.

apoorvemohan:
Tested on Centos 6.6. Will create a pull request after testing on Centos 7 (on PRB cluster).

zenhack:
This has been fixed in master; closing.

PR #435, opened 2015-06-09
Add postgresql support (addresses #428)

apoorvemohan:
Postgresql support for HaaS. Docs updated.


HACKING.rst
```diff
@@ -1,4 +1,35 @@
-The first time you start working in the repository, set up a clean test
+A) MOC supports both sqlite and postgresql. If you are using sqlite, you
```
zenhack: HaaS. The docs shouldn't be talking about the MOC here; it isn't relevant to what the software is capable of.

HACKING.rst
```diff
@@ -1,4 +1,35 @@
...
+		service postgresql initdb
+		chkconfig postgresql on
+		serivce postgresql start
+		
+	Setup permission to allow postgresql to be accessed by other applicaitons::
+		
+		vi /var/lib/pgsql/data/pg_hba.conf
+		
+		By default postgresql uses IDENT based authentication. All you have 
+		to do is allow username and passowrd based authentication for your
```
zenhack: Why is this indented so far?

HACKING.rst
```diff
@@ -1,4 +1,35 @@
...
+		file. E.g. local all all peer -> local all all trust    
+			   host all 127.0.0.1/32 ident -> host all 127.0.0.1/32 trust
+
+	After editing this restart the postgresql server::
+
+			service postgresql restart
+				
+	A default user named 'postgres' is created. Set its password and create a database named 'haas' as follows::
+
+		passwd postgres
```
zenhack: This looks wrong to me; why do you need to set a password for the postgres user?

zenhack:
There are lots of little things here I don't understand the purpose of. Let's talk about it tomorrow.

HACKING.rst
apoorvemohan: my bad.

HACKING.rst
apoorvemohan: what looks wrong? setting postgres user password? I made changes to access a database through a user. so for security I am giving it a password.

zenhack: I'm fairly certain the default behavior is to not have a password in the sense that there is nothing a user can type at the password prompt to log in as that user, not that no authentication is required. Typically, you have to run `su - postgres` as root to get a shell as the postgres user.

haas/test_common.py
```diff
@@ -37,13 +37,15 @@ def network_create_simple(network, project):
...
     return Session()
 
 def releaseDB(db):
     """Do we need to do anything here to release resources?"""
-    pass
-
+    db.close_all()
+    # According to the documentation, we shouldn't need the Session().bind, but this
+    # breaks without it.
+    Base.metadata.drop_all(Session().bind)
```
zenhack: Does this work with sqlite as well? (it looks like it should, but please confirm).

tests/unit/drivers/complex_vlan.py
```diff
@@ -43,6 +43,9 @@ def config_initialize():
                         '{"name":"1", "switch":"test"}, ' \
                         '{"name":"2", "switch":"test"}]')
             cfg.set('driver complex_vlan', 'trunk_ports', '[]')
+            cfg.add_section('database')
+            cfg.set('database', 'sqlite', 'sqlite:///:memory:')
+            cfg.set('database', 'postgres', 'postgresql://postgres:postgres@localhost/haas')
```
zenhack: Please do not hard code the postgres uri; until we can properly parametrize this (which we'll want at some point) I'd like to keep sqlite as the test database, so people can easily run the test suite on their laptops.

INSTALL.rst
```diff
@@ -21,13 +21,20 @@ Then, the rest of the packages can be installed via:
...
-    yum install libvirt bridge-utils ipmitool telnet sqlite httpd mod_wsgi python-pip qemu-kvm python-virtinst
+    yum install libvirt bridge-utils ipmitool telnet httpd mod_wsgi python-pip qemu-kvm python-virtinst
 
 In addition, HaaS depends on a number of python libraries. Many of these are
 available as RPMs as well, but we recommend installing them with pip, since
 this will install the versions that HaaS has been tested with.  This is done
 automatically by the instructions below.
 
+HaaS supports both Sqlite and Postgresql. To install HaaS using Postgresql you need to setup
+Postgresql. You can find the details in HACKING.rst. But if you want to install HaaS using
```
zenhack: It probably makes more sense to have the instructions for the production-oriented database here, and put a reference to it in HACKING than the other way around.

zenhack:
The docs you wrote have some formatting issues; have a look at github's rich text diff.

Also, rather than using e.g. `A)` and `B)`, please use restructured text's standard notation for section headings or lists, depending on what you're going for.

INSTALL.rst
apoorvemohan: Sure.

tests/unit/drivers/complex_vlan.py
apoorvemohan: "If somebody wants to test HaaS code with postgresql, they should edit this information everywhere." - Should I add this information in some doc file?

zenhack: I would definitely add something to the effect that this is a missing feature (patches welcome). I'm not sure I want to be specifically advising people to go and change the hardcoded value every time they want to run the tests; I'd like to just put this near the top of the TODO list and have it solved soon.

zenhack:
One final thing that bugs me about the structure of this a bit: the docs re: setting up the DB are asymmetric, rather than having a thing that says "skip this section if..." I would just have two sections, one for postgres and one for sqlite, and mention above them both that the user should follow the instructions for whichever database they wish to use.

INSTALL.rst
```diff
@@ -40,7 +47,7 @@ Disable SELinux
 ---------------
 
 The setup described below runs into problems with SELinux related to the sqlite
-database. For now the recommended solution is to simply disable SELinux. In
+database and apache. For now the recommended solution is to simply disable SELinux. In
```
zenhack: I don't think we fully understand exactly what the problem SELinux is having is, and I'd rather not mis-document it. With sqlite it was pretty clear what the specific problem was, it's less immediately obvious what's causing it on a postgres setup.

haas/test_common.py
apoorvemohan: Unit tests were successful with sqlite - reconfirmed.

INSTALL.rst
```diff
@@ -21,13 +21,20 @@ Then, the rest of the packages can be installed via:
 
 ::
 
-    yum install libvirt bridge-utils ipmitool telnet sqlite httpd mod_wsgi python-pip qemu-kvm python-virtinst
+    yum install libvirt bridge-utils ipmitool telnet httpd mod_wsgi python-pip qemu-kvm python-virtinst
```
zenhack: We're pulling in `python-virtinst` for the `virt-clone` in CentOS 7 this has been renamed `virt-install`. I just ran into that on the test environment you set up.

INSTALL.rst
apoorvemohan: If a user is not using SQLite he may skip this step, so wouldn't it be better to mention it (even if we are not sure about the problem)? I think how I edited it may not be the best way. Could you please suggest how to mention this?

zenhack: Quoting apoorvemohan (2015-06-10 09:58:59)

>    If a user is not using SQLite he may skip this step

We hit this with something when using postgres as well, though it wasn't 
clear exactly what the issue was. It doesn't seem to work without 
disabling selinux regardless?

apoorvemohan:
committed all changes except the one to use Testsuite.cfg for testing (waiting for #437 to be merged)

INSTALL.rst
```diff
@@ -11,22 +11,49 @@ Prerequisite Software
...
+A default user named 'postgres' is created. Create a database as follows::
+
+    psql -h localhost -U postgres      //Command to enter the psql (postgresql interactive terminal) console
+    create database <db-name>;         //Command to create the database (alway succeeded by a semicolon)
+    \q                                 //Command to quit the database
+
+
+Below are the steps to install and setup SQLite::
+
+    yum -y install sqlite
```
zenhack: sqlite is a dependency of python, so this isn't necessary --- it will have been pulled in automatically.

INSTALL.rst
```diff
@@ -42,9 +69,7 @@ Disable SELinux
 The setup described below runs into problems with SELinux related to the sqlite
 database. For now the recommended solution is to simply disable SELinux. In
 future releases, we will support and recommend the use of SELinux with another
-DBMS, such as MySQL.
-
-::
+DBMS, such as MySQL.::
```
zenhack: I would take out the details about exactly what's causing the error, and just say it doesn't work with SELinux, and refer to bug #456

haas/test_common.py
```diff
@@ -16,7 +16,7 @@
 from haas.model import *
 # XXX: This function has an underscore so that we don't import it elsewhere.
 # But... we need it here.  Oops.
-from haas.model import _on_virt_uri
+#from haas.model import _on_virt_uri
```
zenhack: Rather than commenting this out, just delete it.

haas/test_common.py
```diff
@@ -75,11 +77,16 @@ def database_only(f):
...
         cfg.add_section('general')
-        cfg.set('general', 'driver', 'null')
+        cfg.set('general', 'driver', 'null_vlan')
         cfg.add_section('devel')
         cfg.set('devel', 'dry_run', True)
         cfg.add_section('headnode')
         cfg.set('headnode', 'base_imgs', 'base-headnode, img1, img2, img3, img4')
+	cfg.add_section('database')
+	cfg.set('database', 'sqlite', 'sqlite:///:memory:')
+	cfg.set('database', 'postgres', 'postgresql://postgres:postgres@localhost/haas')
```
zenhack: You've got these two lines in several places. This is analogous to:

     x = 1
     x = 2

...which is silly for obvious reasons.

tests/unit/drivers/complex_vlan.py
zenhack: This is actually configurable on the branch I just submitted a pr for (#455).

zenhack:
There are some conflicts that need fixing before we can merge this. can you please address those?

zenhack:
I worked through fixing the conflicts with @apoorvemohan and this looks good to me, but I think others should be the ones to review it since I have my fingers in it a bit too.

@gsilvis, @lokI8, @henn?

INSTALL.rst
```diff
@@ -39,12 +54,9 @@ The HaaS software can then be installed by running:
...
 
-The setup described below runs into problems with SELinux related to the sqlite
-database. For now the recommended solution is to simply disable SELinux. In
-future releases, we will support and recommend the use of SELinux with another
-DBMS, such as MySQL.
-
-::
+The setup described in this document runs into problems with SELinux. In the
+future, we hope to ship a set of SELinux security policies with HaaS, but for
+now the solution is to disable SELinux::
```
henn: If disabling SELinux was for sqlite, do deployers still need to disable it for Postgres?

HACKING.rst
```diff
@@ -28,6 +31,10 @@ you may need to run::
 
   pip install importlib
 
+You may get an error 'psycopg2 package not found' when you do 'haas init_db' 
+in the next step if you are using PostgreSQL database. You may need to run::
+
+  pip install psycopg2
```
henn: Rather than having an explicit pip command, perhaps we should just place the requirement in setup.py? It wouldn't hurt anything to have the extra package installed for sqlite users.

haas/model.py
```diff
@@ -243,14 +242,14 @@ class Network(Model):
     # The project to which the network belongs, or None if the network was
     # created by the administrator.  This field determines who can delete a
     # network.
-    creator_id = Column(String,ForeignKey('project.id'))
+    creator_id = Column(Integer,ForeignKey('project.id'))
```
henn: Rather than specifying the "id" column type in this and the other local calls you changed, let's just remove the type and let SqlAlchemy [infer it from the foreignkey](http://docs.sqlalchemy.org/en/rel_0_9/core/metadata.html#sqlalchemy.schema.Column.params.type_).

This helps us not to repeat ourselves, making it easier if we want to move to a biginteger primary key (#479) or something else.

INSTALL.rst
```diff
@@ -11,24 +11,39 @@ Prerequisite Software
...
 
     yum install epel-release
 
-Then, the rest of the packages can be installed via:
-
-::
+Then, the rest of the packages can be installed via::
 
-    yum install libvirt bridge-utils ipmitool telnet sqlite httpd mod_wsgi python-pip qemu-kvm python-virtinst
+    yum install libvirt bridge-utils ipmitool telnet httpd mod_wsgi python-pip qemu-kvm python-virtinst virt-install python-psycopg2
```
henn: Above you recommended installing psycopg2 via pip.

Either way is fine (install via the OS may be better since it will be updated with the system packages), but we should be consistent in which one we recommend.

INSTALL.rst
zenhack: We hit another SELinux issue, I'm not sure it was even the database. But yes, even with postgres it still doesn't work

HACKING.rst
zenhack: Thing is, if in the future we add sort for more databases (e.g. mysql), this will become "install the library for your database," which won't necessarily be the same for everyone.

INSTALL.rst
zenhack: Good catch. Eventually I'd like us to be doing everything (including haas itself) via distro packages, but right now installing it via pip would be more consistent with how we're handling the rest of the python dependencies. We should definitely remove one of these though.

zenhack: @henn, actually, the pip reccomendation is in `HACKING`, which is a different use case.

haas/model.py
zenhack: Works for me. Since @apoorvemohan is unlikely to get to this until he gets back, I'm going to open up a new pr from my branch with this extra change.

zenhack:
Superseded in favor of #481, closing.

PR #436, opened 2015-06-09
Don't import private function _on_virt_uri

zenhack:
There's actually a fairly straightforward way of doing this without that
function...

### DIFF
--- a/haas/test_common.py
+++ b/haas/test_common.py
@@ -14,12 +14,10 @@
 
 from functools import wraps
 from haas.model import *
-# XXX: This function has an underscore so that we don't import it elsewhere.
-# But... we need it here.  Oops.
-from haas.model import _on_virt_uri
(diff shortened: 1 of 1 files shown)


gsilvis:
This is definitely better.  I kind of wish we didn't have to catch that particular exception here (maybe the function should catch that exception and rethrow it as something else?), but this is still definitely better. +1

zenhack:
Yeah, I'm not thrilled either, but it was an obvious easy improvement.

lokI8:
Definitely an improvement. I agree, in that I'm not terribly happy that we're not waiting to see if it ends up working, we're just catching the exception and moving on... but you're right, still an improvement.

+1 and merging, but would be nice if we can maybe add a time to wait before trying the call in the tear-down process? Don't know how impactful this is to the testing suite, but it would be gross to leave around state...

PR #437, opened 2015-06-09
Testsuite.cfg

zenhack:
I haven't gotten a chance to run the deployment tests yet; would like to do that before we merge.

@apoorvemohan, this will be useful for your postgres stuff.


zenhack:
It should be noted that this builds on top of #436; that should be merged first.

haas/test_common.py
```diff
@@ -14,12 +14,80 @@
...
+
+    ``config_dict`` should be a dictionary mapping section names (strings)
+    to dictionaries mapping option names within a section (again, strings)
+    to their values. If the value of a section, or option is None, that
+    section or option is removed. Otherwise, the section is created if it
+    does not exist, and any options are set to the specified values.
+    """
+    for section in config_dict.keys():
+        if config_dict[section] is None:
+            cfg.remove_section(section)
```
gsilvis: Is it safe to remove a non-existent section?

haas/test_common.py
```diff
@@ -14,12 +14,80 @@
...
+        else:
+            if not cfg.has_section(section):
+                cfg.add_section(section)
+            for option in config_dict[section].keys():
+                if config_dict[section][option] is None:
+                    config.remove_option(section, option)
```
gsilvis: Same question as 6 lines above

haas/test_common.py
zenhack: Seems to be.

gsilvis: It's odd that they throw an exception for creating an existing section but not one for deleting a non-existent one.  Okay, though

haas/test_common.py
zenhack: Same answer.

haas/test_common.py
zenhack: Deleting returns `False`, so they are signifying it somehow. The inconsistency is weird though.

gsilvis:
This looks okay to me.  +1

zenhack:
Subsumed by #455, and I may have done some rebasing, so don't want to risk merging these and causing conflicts. Closing.

PR #439, opened 2015-06-16
Use iproute2 in scripts/create_bridges

### LINKED ISSUE #433: Use iproute2 commands in create_bridges (not net-tools)
The `create_bridges` script uses `ifconfig` from the net-tools package. More and more distros are not installing net-tools by default, in favor of iproute2 (net-tools has been unmainted for some time). We should move over to using the `ip` command that's part of iproute2

gsilvis:
Fixes #433 .  I've run the commands manually and they have the same effect, but we should run a full test of that on a real HaaS setup.  Once I do that, I'll add a commit changing the list of dependencies in the docs.

### DIFF
--- a/scripts/create_bridges
+++ b/scripts/create_bridges
@@ -21,16 +21,24 @@ from haas.drivers.driver_tools.vlan import get_vlan_list
 
 config.load()
 
+def ip_link(args):
+    return call(['ip', 'link'] + args)
+
 def create_bridge(vlan_no):
     vlan_no = str(vlan_no)
     trunk_nic = cfg.get('headnode', 'trunk_nic')
     bridge = 'br-vlan%s' % vlan_no
     vlan_nic = '%s.%s' % (trunk_nic, vlan_no)
-    call(['brctl', 'addbr', bridge])
-    call(['vconfig', 'add', trunk_nic, vlan_no])
-    call(['brctl', 'addif', bridge, vlan_nic])
-    call(['ifconfig', bridge, 'up', 'promisc'])
-    call(['ifconfig', vlan_nic, 'up', 'promisc'])
+    ip_link(['add', 'type', 'bridge', 'name', bridge])
+    ip_link(['set', 'promisc', 'on', 'dev', bridge])
+    ip_link(['set', 'up', 'dev', bridge])
+    ip_link(['add', 'type', 'vlan',
+             'link', trunk_nic,
+             'name', vlan_nic,
+             'id', vlan_no])
+    ip_link(['set', 'promisc', 'on', 'dev', vlan_nic])
+    ip_link(['set', 'up', 'dev', vlan_nic])
+    ip_link(['set', 'dev', vlan_nic, 'master', bridge])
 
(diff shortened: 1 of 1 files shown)


zenhack:
Looks good. Once you've done the testing/docs bit I'll take another look and probably +1.

scripts/create_bridges
```diff
@@ -21,16 +21,24 @@ from haas.drivers.driver_tools.vlan import get_vlan_list
...
     trunk_nic = cfg.get('headnode', 'trunk_nic')
     bridge = 'br-vlan%s' % vlan_no
     vlan_nic = '%s.%s' % (trunk_nic, vlan_no)
-    call(['brctl', 'addbr', bridge])
-    call(['vconfig', 'add', trunk_nic, vlan_no])
-    call(['brctl', 'addif', bridge, vlan_nic])
-    call(['ifconfig', bridge, 'up', 'promisc'])
-    call(['ifconfig', vlan_nic, 'up', 'promisc'])
+    ip_link(['add', 'type', 'bridge', 'name', bridge])
+    ip_link(['set', 'promisc', 'on', 'dev', bridge])
```
henn: Why enable promiscuous mode?

scripts/create_bridges
```diff
@@ -21,16 +21,24 @@ from haas.drivers.driver_tools.vlan import get_vlan_list
...
+    ip_link(['set', 'up', 'dev', bridge])
+    ip_link(['add', 'type', 'vlan',
+             'link', trunk_nic,
+             'name', vlan_nic,
+             'id', vlan_no])
+    ip_link(['set', 'promisc', 'on', 'dev', vlan_nic])
```
henn: Same promiscuous mode question

scripts/create_bridges
gsilvis: Required for traffic to go through to machines that aren't the host

henn: > Required for traffic to go through to machines that aren't the host

Are you sure we need it; that proxy ARP doesn't provide the necessary
functionality?

The current commands seem to work so I'd be surprised if we need to
enable promiscuous mode.

gsilvis: The current commands also enable promiscuous mode

henn:
:+1:

zenhack:
So the point of this was to remove dependencies on things like bridge-utils (which aren't a big deal, but hey, it's fewer packages to install). Maybe remove the references in INSTALL.rst?

Other than that, I'd like to see this tested on hardware, but it looks great.

henn:
@gsilvis - would you have some time in the next couple weeks to address @zenhack's comment? If not, could you let us know so we can transfer this PR?

Thanks!

gsilvis:
I recommend transferring---the doc update won't be hard, but I don't have time to test this exhaustively right now.

henn:
@SahilTikale is going to convert this startup script to use systemd.

Given our effort to move to a VPN service (#559), these scripts will likely need to be changed in more substantive ways anyhow. Thus, I am closing this unless we see a need.

PR #441, opened 2015-06-17
Decorators to fixtures

zenhack:
This pulls the decorators apart so we can apply them more flexibly --- right now we can't easily add an extra setup step for a test

This patch isn't as big as it looks; it pulls in the stuff from #434 and #437, so it's sharing some commits.


zenhack:
Superseded by #455

PR #442, opened 2015-06-17
Network pools

zenhack:
This factors the vlan allocation stuff out into extensions that are pluggable; we can use something else for rHaaS (probably create a network upstream and use the name of the network).

Some of the driver related tests have been half-heartedly forced into place; next up is the switch refactoring, which will replace large parts of those anyway, and  mesh better with this.


zenhack:
This builds on top of #441, so it's even bigger.

henn:
Thanks for your great work!

One slight modification to the syntax to consider would be specifying the plugin in the section name instead of a special `[extensions]` section. I don't think it would affect the current code too much, and would have a few advantages:

1. Determining which config section pertains to which plugin would be really easy. In the current model, the user would need to look up somewhere out of band the section name that pertains to a given plugin.

2. It avoids namespace collisions. Say two people wrote cisco device drivers (say the python modules were zenhack.cisco and henn.cisco), but both named the the config sections something more generic (like "cisco").

3. It allows a module to be used twice without having a config name conflict (assuming it's designed for that). Perhaps one of the options for a cisco driver were the IP address of the router. Being able to configure the section twice would allow me to create 2 instances.

An example of this new method would be:
```
[ext:haas.ext.network.cisco5500]
ip = 1.2.3.4
username = admin
password = admin

[ext:haas.ext.auth.openstack]
keystone_ip = 5.6.7.8
```

zenhack:
superseded by #455, closing.

zenhack:
@henn, I feel like we talked about this when the design was being proposed.

1. (I'd forgotten to document this, #455 is updated with the changes now) The strategy here is that we just config section that's the same as the extension module.

2. same response as (1); this could only happen if the extensions were in violation of that convention.

3. It is in fact not designed for that. #455 moves switch objects to the database, not the config file.

Let's move any further discussion on this to #455.

PR #444, opened 2015-06-18
Fix #443

### LINKED ISSUE #443: node_delete doesn't check if the node has nics
If it does, sqlalchemy throws an exception (because the `owner_id` column in `Nic` is not nullable).

We should check this and throw a `BlockedError` (which I think is most consistent with the rest of HaaS' behavior.

zenhack:
Includes a test for the bug.


gsilvis:
+1

lokI8:
+1, merging

PR #452, opened 2015-07-08
Remove erroneous information about apache config.

### LINKED ISSUE #440: INSTALL.rst suggests using AllowEncodedSlashes NoDecode on newer versions of apache, but this breaks things
you end up sending things like gi1%2F0%2F13 to the switch. we should update the docs to correct this

zenhack:
Resolves #440.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -277,11 +277,7 @@ safe to replace it.)
 strings that should be escaped (see `issue 361 <https://github.com/CCI-MOC/haas/issues/360>`_). By default, Apache `Doesn't
 allow <https://stackoverflow.com/questions/4390436/need-to-allow-encoded-slashes-on-apache>`_
 this due to security concerns. ``AllowEncodedSlashes On`` enables the passing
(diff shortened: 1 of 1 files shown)


zenhack:
//cc @gsilvis

gsilvis:
Do we fully understand the situation here?  I know we added that to make something work.  Is there no setup where it's required?

zenhack:
We never actually tried it with a more recent apache until I set it up on the Cisco gear. @henn just thought would be better because it was slightly more conservative. Unfortunately it turns out you need to decode the slashes for this to work, so the documentation was strictly incorrect.

gsilvis:
Okay.  +1

lokI8:
Okay. I remember this being a security concern to include slashes, but we need them to match switch configs... should we make an issue so that we remember this is a potential hole? Otherwise +1

zenhack:
It can be a security issue in other contexts, e.g. if the argument in question is an path name, someone might pass something like `../../etc/passwd` to break out of your document root. in this case that's not true, so having slashes in the input isn't dangerous.

lokI8:
Fair enough, that makes sense. +1, merging

Issue #46, opened 2014-04-15
Use more secure mechanism to talk to dell switches


zenhack:
The managed switches supposedly support public key auth (ssh?); we be doing this instead of e.g. telnet (in the dell driver).

henn:
Switching to using ssh should be relatively straightforward.

zenhack:
Should be a design consideration when addressing #744

naved001:
Fixed by #922 and #907

PR #455, opened 2015-07-15
Network refactoring work and drivertization

zenhack:
Holy crap it's ready for review.


henn:
Wonderful!

Were you planning to do a code walkthrough? I started reading through it, but might wait if you were planning to do a walkthrough anyway.

zenhack:
@lokI8, @gsilvis would you find it useful for me to do a code walkthrough? It's a big patch, and there are a handful of different things in here.

docs/network-drivers.md
```diff
@@ -1,47 +1,127 @@
...
-to the HaaS headnode host.
+This document describes the network driver model, primarily from a 
+system administrator's standpoint, though there are some details useful 
+for developers writing tools which interact with the API as well.
+
+# Overview
+
+There are two classes of network-related drivers: network allocators and 
+switches. Switches are straightforward to understand; they provide 
+support for a particular hardware switch. Network work allocators are a
```
lokI8: Network work allocator? seems like you doubled a word =)

docs/network-drivers.md
```diff
@@ -1,47 +1,127 @@
...
+### switch_register_port
+
+Like the powerconnect driver, the Nexus driver accepts port names of the 
+same format accepted by the underlying switch, in this case (e.g.) 
+``ethernet 1/42``. The same concerns about validation apply.
+
+## Using multiple switches
+
+Networks may span multiple switches. No special configuration of HaaS 
+itself is required; just register each switch as normal and things shoul
```
lokI8: should*

docs/rest_api.md
```diff
@@ -63,6 +185,74 @@ Full Api spec:
...
+
+Possible Errors:
+
+* 409, if a switch named `<switch>` already exists.
+
+### switch_delete
+
+`DELETE /switch/<switch>`
+
+Delete the switch named `<switch>.
```
lokI8: extra '

haas/api.py
```diff
@@ -529,56 +565,111 @@ def network_delete(network):
...
+          has access to the network. Otherwise, the network is public.
+    """
+    db = model.Session()
+    allocator = get_network_allocator()
 
+    network = _must_find(db, model.Network, network)
+    return json.dumps({
+        'name': network.label,
+        'channels': allocator.legal_channels_for(db, network.network_id),
+    })
```
gsilvis: This seems to be missing the 'creator' and 'access' fields

haas/api.py
```diff
@@ -529,56 +565,111 @@ def network_delete(network):
...
+    cls = concrete_class_for(model.Switch, type)
+    if cls is None:
+        raise BadArgumentError('%r is not a valid switch type.' % type)
+    cls.validate(kwargs)
+    obj = cls(**kwargs)
+    obj.label = switch
+    obj.type = type
+
+    db.add(obj)
+    db.commit()
```
gsilvis: It's a little unfortunate that we're shadowing the Python built-in 'type' in this function, but I don't see any good way of avoiding that

haas/deferred.py
```diff
@@ -51,22 +52,33 @@ def apply_networking():
...
                 'Not modifying NIC %s; NIC is not on a port.' %
                 nic.label)
 
-    # Apply them
-    driver_name = config.cfg.get('general', 'driver')
-    driver = importlib.import_module('haas.drivers.' + driver_name)
-    driver.apply_networking(network_map)
+    # Close all of our sessions:
+    for session in switch_sessions.values():
+        session.disconnect()
```
gsilvis: At some point, we might want to make it so that we can keep these sessions open between calls of `apply_networking`, but that'll take some refactoring

haas/ext/network_allocators/null.py
```diff
@@ -0,0 +1,49 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS
+# IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied.  See the License for the specific language
+# governing permissions and limitations under the License.
+"""A null network allocator.  Network IDs are random and arbitrary.
+
+A null switch driver.  Network IDs are random and arbitrary.  Applying the
+network state does nothing. "null" is the only channel id.
+"""
```
gsilvis: Repeated text in the docstring

zenhack: And worse than that, it's talking about switches! I'll fix.

haas/deferred.py
zenhack: I'd also want to test what happens when you leave a session for one of those sessions open for a very long time. But yeah, I think that's down the line.

haas/api.py
zenhack: I feel like at some point there'd been a bit of debate as to what we actually wanted to list here, and came to this? I don't really recall. leaving it out means it's easier to change later. At some point we should take advantage of the fact that we can have option/structure arguments to make that whole array of options more intuitive.

haas/api.py
zenhack: yeah... me neither.

zenhack:
@gsilvis, @lokI8, I believe I've addressed all of your comments thus far.

docs/driver-model.rst
```diff
@@ -0,0 +1,24 @@
...
+superclass, e.g. ``Switch`` or ``OBM``, and the drivers themselves are
+subclasses of that superclass (typically defined from within extensions). Under
+most curcumstances, SQLAlchemy makes this "just work"; when fetching the object
+from the database, the right subclass will be instantiated, and everything will
+"just work." However, when *creating* an object for the first time, HaaS must be
+told which driver to use (is the switch a powerconnect or a nexus?) For these
+cases, the api calls in question (node_register, switch_register...) will expect
+a type field to be provided. The module ``haas.class_resolver`` provides
+facilities for finding the appropriate subclass based on the contents of this
+type field.
```
henn: It would be good to reference an example from here.

henn:
[x-posted from #442 with modifications]

One slight modification I would like to submit for consideration for the wonderful new plugin extension functionality is to specify the plugin in the section name instead of the special `[extensions]` section. I don't think it would affect the current code too much, and would have a few advantages:

1. Conciseness: I believe it would only ever save space in the config file. Currently, if something has configuration it needs to be specified in both  `[extensions]` and `[the.extension.name]`. If there were no configuration needed then one could simply specify an empty `[the.extension.name]` section.

2. Specifying the extension name twice violates the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle.
 
3. Determining which config section pertains to which plugin would be really easy. In the current model, the user would need to look up somewhere out of band the section name that pertains to a given plugin, or rely on an unenforced convention of using the module name.

4. It potentially allows a module to be used twice without having a config name conflict (assuming it's designed for that). Perhaps one of the options for a cisco driver were the IP address of the router. Being able to configure the section twice with two separate IPs could allow me to create 2 instances.

An example of this new method would be:
```
[haas.ext.network.cisco5500]
ip = 1.2.3.4
username = admin
password = admin

[haas.ext.network.cisco5500]
ip = 5.6.7.8
username = admin
password = admin

[haas.ext.auth.openstack]
keystone_ip = 5.6.7.8
```

To make the parsing really easy, we could prepend the section name like so: `[ext:haas.ext.network.cisco5500]`

zenhack:
re: (3), the convention is just as unenforced with the model you propose; nothing stops a plugin from just reading it's config from elsewhere in the file --- you're still relying on convention.
re: (4) I'm fairly sure the ConfigParser module doesn't support multiple copies of the same section anyway, and you can't double-import a module, so this is moot.

If we did this, I'd want to change the section names for the non-extension parts of the config so that they're properly namespaced, but I otherwise don't have an objection.

zenhack:
Also re: (4) I have a pretty strong philosophical objection to double-loading plugins.

As I've mentioned before, the much of the point of this patch was to move state re:switches *out* of the config file and into the database.

gsilvis:
I actually prefer explicit lists of plugins to load, here, for mostly similar reasons to what @zenhack said.

gsilvis:
I think all of my issues have now been addressed.  I'll check this again this evening and then if it still looks good I'll +1 it.

docs/network-drivers.md
```diff
@@ -1,47 +1,127 @@
...
+driver-dependant, below are the details for each of these switches.
+
+## Powerconnect driver
+
+### switch_register
+
+To register a Dell Powerconnect switch, the ``"type"`` field of the
+request body must have a value of::
+
+    http://schema.massopencloud.org/haas/switches/powerconnect55xx
```
henn: We should version these URLs.

zenhack: how about `http://schema.massopencloud.org/haas/v0/switches/powerconnect55xx` ?

docs/network-drivers.md
```diff
@@ -1,47 +1,127 @@
...
+
+In addition, it requires three extra fields: ``"username"``,
+``"hostname"``, and ``"password"``, which provide the necessary
```
henn: Currently, I consider it a feature (defense in depth) that one doesn't need to tell the API server anything about a switch's credentials, nor does the API server need to have access to the switch's management interface.

While the additional configurability of having switches configured via the API is nice, it would also be desirable to maintain the defense in depth now possible by hiding the the switch's credentials from the API server.

What do folks think about this? Possible solutions include:
* Use SQL permissions to make credential columns write-only for the API server (and hope a compromised server doesn't record it somewhere)
* Keep credentials in the config file
* Use encryption

docs/network-drivers.md
```diff
@@ -1,47 +1,127 @@
...
+
+    [haas.ext.network_allocators.vlan_pool]
+    vlans = 300-500, 700-750
+
+# Network allocator drivers
+
+The only network allocator shipped with HaaS that is of interest to
+users (there are others useful for development purposes) is the VLAN
+allocator. The name of the extension is
+``haas.ext.network_allocators.vlan_pool``, and it requires a single
```
henn: Have you given thought to how we'll handle divergent vlan_pools managed by a single HaaS instance (like I'm told junipers can work)? Configure separate vlan_pools via the API?

docs/network-drivers.md
```diff
@@ -1,47 +1,127 @@
...
+
+    http://schema.massopencloud.org/haas/switches/nexus
+
+The nexus driver requires the same additional fields as the powerconnect
+driver, plus an additional field "dummy_vlan", which should be a JSON
+number corresponding to an unused VLAN id on the switch. This VLAN
+should be deactivated (and thus no traffic should flow across it ever).
+This exists to get around an implementation problem related to disabling
+the native VLAN.
+
```
henn: Would it be possible to include an example command to accomplish this?

docs/network-drivers.md
```diff
@@ -1,47 +1,127 @@
...
+### switch_register_port
+
+Like the powerconnect driver, the Nexus driver accepts port names of the
+same format accepted by the underlying switch, in this case (e.g.)
+``ethernet 1/42``. The same concerns about validation apply.
```
henn: Just a suggestion which you can take or leave: it might be worth rewording this section without the powerconnect references, since the reader may skip straight to this section. As we add drivers (even 1 or 2 more), it could become cumbersome to both manage and navigate references to other drivers.

docs/network-drivers.md
zenhack: Getting the switch info *out* of the config file and *into* the database was the **whole point** of this work. Why did you not bring this up when you reviewed the proposal a month and a half ago?

Addressing the bullet points in order:

* The SQL permissions thing might be interesting to explore at some point, but I don't it think needs to be a blocker. It's also probably mostly a matter of researching/documenting and probably orthogonal to this stuff.
* As mentioned, not having the switch info stored in the config file was the whole point of this patch.
* This is vague. What did you have in mind?

In addition to being able to read database tables, they need to actually have IP connectivity to the switch management console. We already as a general policy don't allow management interfaces on untrusted networks. This is not to say that enhancements in this regard aren't a good extra layer (e.g. #46), but to be clear, we are not strictly dependent on the secrecy of that information for security.

The concerns you bring up also apply to the IPMI credentials, which have been in the database for as long as we've supported IPMI.

docs/network-drivers.md
zenhack: Probably worth doing, yeah. I'll get to it.

docs/network-drivers.md
zenhack: I think this is a good thing to keep in mind when/if we do add more drivers. @gsilvis, do you think it's important to do this now?

docs/network-drivers.md
zenhack: Not sure what you mean by divergent?

My understanding/thinking on the juniper thing: internally the vlans/networks/whathaveyou just have arbitrary names, and can be connected to any port tagged as anything. In this case there would be no restrictions on what "channel" you could use.

When I talked about it with @gsilvis we decided the right approach was probably to just write another network allocator driver.

zenhack:
re: extension config file sections --- I'll point out that as it stands, the only extension that even *has* config options is the `vlan_pool` allocator (you used switches in your examples, but those actually store their information in the db). @gsilvis and I have also vaguely talked about whether having the list of VLANs in the config file in the long term (vs. adding them via the api) was the right thing; there's a vague feeling of "no," which would actually mean there would be *no* extension-specific config options in the main repo. I anticipate that the ability to have extra config options will not be heavily used.

docs/network-drivers.md
gsilvis: Encryption doesn't help much here.  HaaS would need the key to decrypt the passwords, so someone with access to the system would have that too.

docs/network-drivers.md
gsilvis: I think this version is fine for now.  I would vote for fixing it either in a separate pull request right after this, or the next time we add a driver or rework this section in general.

docs/network-drivers.md
henn: > Encryption doesn't help much here. HaaS would need the key to decrypt
> the passwords, so someone with access to the system would have that too.
>
Regarding the encryption option, the administrator could have a separate
haas.cfg on the Network Action Server which contained either a symmetric
key or an asymmetric private key. The latter would be similar to
OpenStack services (like Nova and Cinder instances) having their own
identities, per our past discussions.

The threat I'm trying to protect against is a compromise to the API
server. Without switch credentials, the API server (or DB server for
that matter) would be limited to making requests to attach/detach
networks. These requests might be separately logged, leaving an audit
trail. Also prevented would be more malicious operations, like creating
a mirror port.

zenhack: They could also do anything they want to the nodes' IPMI controllers.

I'm disinclined to spent too much time on this; separate api/network daemon servers are still something we can do in theory, but have never tried nor felt the need to devote the resources to. I think any of these solutions is going to involve a substantial amount of work, and it will only have any impact at all in a use case that we haven't actually had need for.

gsilvis:
With that, have a +1

zenhack:
We just had a conversation in person about a number of these issues, and resolved some things. For the purposes of documentation:

* The network allocator design/juniper concerns: We came to the conclusion that, at a minimum this doesn't move us away from where we want to go there. The cases we don't have a clear picture for require about the same amount of design and modification as they did before.
* We're keeping the extension mechanism as-is. Most of the arguments for and against are already in the comments above.
* We're adding an issue (#460) to implement the database permissions solution @henn suggested (though not doing so as part of this pr). A couple other solutions were proposed, but rejected:
  * encrypting the credentials with a key the API server doesn't have (but the network action server does). We decided that this didn't hit a good complexity to payoff ratio.
  * @henn suggested (and still thinks it's a good idea, though I disagree and @gsilvis had misgivings as well), having the credentials stored in a file on the network action server, and the database just reference that file. There are some subtle issues here about what the API server can actually do in this case; it can spoof some information that the network daemon looks at, and implementing something like this would require careful desgin. I'm also of the opinion that this is taking us in the wrong direction: we've been trying to move stuff like this *out* of the config file, not in, and actually making use of this will add administrative complexity.

zenhack:
I think I've addressed everything that's been brought up. @gsilvis has already +1'd the latest version. @henn?

zenhack:
For reference, the original proposal is here:

https://github.com/CCI-MOC/haas/wiki/network-driver-model-refactor

Some of the details have changed, but it is still a very good overview.

docs/driver-model.rst
```diff
@@ -0,0 +1,24 @@
...
+Each type of object that supports different drivers has a top-level superclass,
+e.g. ``Switch`` or ``OBM``, and the drivers themselves are subclasses of that
+superclass (typically defined from within extensions). Under most
+curcumstances, SQLAlchemy makes this "just work"; when fetching the object from
+the database, the right subclass will be automatically used. However, when
+*creating* an object for the first time, HaaS must be told which driver to use
+(is the switch a powerconnect or a nexus?). For these cases, the api calls in
+question (node_register, switch_register...) will expect a type field to be
+provided. The module ``haas.class_resolver`` provides facilities for finding
+the appropriate subclass based on the contents of this type field.
```
henn: Could you reference one of the classes you implemented as an example from here?

docs/network-drivers.md
henn: thanks!

haas/api.py
```diff
@@ -258,34 +275,57 @@ def node_connect_network(node, nic, network):
...
+        raise BlockedError("The channel is already in use on the nic.")
+
+    if not allocator.is_legal_channel_for(db, channel, network.network_id):
+        raise BadArgumentError("Channel %r, is not legal for this network." %
+                               channel)
+
+    db.add(model.NetworkingAction(nic=nic,
+                                  new_network=network,
+                                  channel=channel))
     db.commit()
```
henn: The (new) documentation says node_connect/detach_network returns 202 Accepted. Does this do that?

haas/api.py
```diff
@@ -258,34 +275,57 @@ def node_connect_network(node, nic, network):
...
-
-    db.add(model.NetworkingAction(nic, None))
+    attachment = db.query(model.NetworkAttachment)\
+        .filter_by(nic=nic, network=network).first()
+    if attachment is None:
+        raise BadArgumentError("The network is not attached to the nic.")
+    db.add(model.NetworkingAction(nic=nic,
+                                  channel=attachment.channel,
+                                  new_network=None))
     db.commit()
```
henn: The (new) documentation says node_connect/detach_network returns 202 Accepted. Does this do that?

haas/api.py
```diff
@@ -258,34 +275,57 @@ def node_connect_network(node, nic, network):
...
+        channel = allocator.get_default_channel(db)
+
     if nic.current_action:
         raise BlockedError("A networking operation is already active on the nic.")
 
     if (network.access is not None) and (network.access is not project):
         raise ProjectMismatchError("Project does not have access to given network.")
 
-    db.add(model.NetworkingAction(nic, network))
+    if _have_attachment(db, nic, model.NetworkAttachment.network == network):
```
henn: Could we do this inline `_have_attachment()`, since it is only used twice? This would be friendlier to those new to the project.

If we want to be a little more compact, perhaps we could `import NetworkAttachment`? We could also use [exists()](docs.sqlalchemy.org/en/rel_0_9/orm/query.html?highlight=exists#sqlalchemy.orm.query.Query.exists) which is a little shorter and could be more efficient on the server.

docs/rest_api.md
```diff
@@ -63,6 +185,74 @@ Full Api spec:
...
+
+### switch_register_port
+
+`PUT /switch/<switch>/port/<port>`
+
+Register a port `<port>` on `<switch>`.
+
+The permissable values of `<port>`, and their meanings, are switch
+specific; see the documentation for the apropriate driver for more
+information.
```
henn: Are there special values that shouldn't be used here? Like '/', '&', ...?

examples/haas.cfg
```diff
@@ -0,0 +1,93 @@
...
+# use the database myDatabase:
+# uri = postgresql://alice:secret@dbhost:5432/myDatabase
+
+[devel]
+# A handful of functions which touch external state (start vms, communicate with
+# swithces...) can be supressed for development purposes. When this is done, the
+# calls to the function will be logged, but not actually executed. If the option
+# ``dry_run`` is present (regardless of its value), this functionality is
+# enabled:
+dry_run=
```
henn: Should this be commented out for our "real example"?

haas/api.py
```diff
@@ -258,34 +275,57 @@ def node_connect_network(node, nic, network):
 
     project = node.project
 
+    allocator = get_network_allocator()
+    if channel is None:
+        channel = allocator.get_default_channel(db)
+
```
henn: More of a stylistic comment (take it or leave it), but it might make sense to move this closer to where the 2 other channel-related error checks are so that the logic dealing with 'channel' is close together and easy to see.

haas/ext/network_allocators/vlan_pool.py
```diff
@@ -0,0 +1,82 @@
...
+    return returnee
+
+
+class VlanAllocator(NetworkAllocator):
+    """A allocator of VLANs. The interface is as specified in ``NetworkAllocator``."""
+
+    def get_new_network_id(self, db):
+        vlan = db.query(Vlan).filter_by(available=True).first()
+        if not vlan:
+            return None
```
henn: Stylistic (take or leave): is it better to return `None` or throw an exception in this case? Could there ever be a `get_new_network_id()` that legitimately returned `None`?

haas/api.py
```diff
@@ -612,7 +708,8 @@ def port_detach_nic(port):
     """
     db = model.Session()
 
-    port = _must_find(db, model.Port, port)
+    switch = _must_find(db, model.Switch, switch)
+    port = _must_find_n(db, switch, model.Port, port)
```
henn: Should we check `nic.current_action` ?

haas/api.py
```diff
@@ -258,34 +275,57 @@ def node_connect_network(node, nic, network):
...
-    db.add(model.NetworkingAction(nic, network))
+    if _have_attachment(db, nic, model.NetworkAttachment.network == network):
+        raise BlockedError("The network is already attached to the nic.")
+    if _have_attachment(db, nic, model.NetworkAttachment.channel == channel):
+        raise BlockedError("The channel is already in use on the nic.")
+
+    if not allocator.is_legal_channel_for(db, channel, network.network_id):
+        raise BadArgumentError("Channel %r, is not legal for this network." %
+                               channel)
+
```
henn: It's fine if you want to defer this (you aren't introducing the bug), but we aren't checking to see if there is a port connected before passing the request along.

haas/deferred.py
```diff
@@ -51,22 +52,33 @@ def apply_networking():
...
 
-    # Then perform the database change and delete them
+    # Then perform the database changes and delete them
     for action in actions:
-        action.nic.network = action.new_network
+        if action.new_network is None:
+            db.query(model.NetworkAttachment)\
+                .filter_by(nic=action.nic, channel=action.channel)\
+                .delete()
```
henn: Could we make this line `.one().delete()` ?

An extra error check won't hurt and might catch something.

haas/ext/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""Extensions which are maintained in-tree.
+
+All public in this package are extensions. They could, in concept, be
```
henn: Could you clarify "All public"?

haas/ext/network_allocators/null.py
```diff
@@ -0,0 +1,48 @@
...
+        return str(uuid.uuid1())
+
+    def free_network_id(self, db, net_id):
+        pass
+
+    def populate(self, db):
+        pass
+
+    def legal_channels_for(self, db, net_id):
+        return ["null"]
```
henn: To provide better compatibility with scripts, I'm wondering if this should return `vlan/native` or if that would better belong in a null_vlan.py driver?

henn:
I'm done reviewing. Let me know if you have any questions.

Regarding the credential storage concern, if we can agree on some general design in #470 then I'm happy.

haas/api.py
zenhack: Good catch, it was supposed to be doing that all along, so I'd just assumed it did. will fix.

haas/api.py
zenhack: It was  a lot harder to understand the purpose, and to spot the difference between the two without the descriptive name (and the docstring). @gsilvis, opinion?

docs/rest_api.md
zenhack: I'll add a note to the top about needing to escape special characters.

examples/haas.cfg
zenhack: This example isn't "real" (or for any particular use case) anymore --- just a reference for all of the options. I don't think it really matters whether any of the options are commented out or not.

haas/api.py
zenhack: This isn't actually checking for an error --- None just indicates that it should use the default channel (one was not provided by the user).

haas/ext/network_allocators/vlan_pool.py
zenhack: network ids are strings by definition, so `None` is safe as a sentinel. The only place this is actually called is in `network_create`, and I think I like the explicit check there. @gsilvis, thoughts?

haas/api.py
zenhack: We should --- good catch.

haas/api.py
zenhack: Might as well fix it now. good catch.

haas/deferred.py
zenhack: I think `one()` actually fetches the object, whereas `delete` is a method on the query; I don't think that will work at all (the object doesn't have a delete method), and even if it does it will needlessly execute an extra query.

haas/api.py
zenhack: Actually, there are other constraints: what the nic is attached to networks, for example? I think the thing to check is that the node is free; that will ensure everything else is already detached.

haas/api.py
zenhack: Started working on this, and found there was actually a test that checks for the behavior you're pointing out. Talked to @gsilvis, I think the reasoning at the time was that you should never actually have nics in the DB that aren't connected to the switch, and that this was an administrator error. The driver writes a warning to the log and just noops. I'm not sure I still think this is the right behaviour, but let's not add unrelated api changes to this patch.

haas/ext/network_allocators/null.py
zenhack: This doesn't cause any compatibility problems. The mock driver actually doesn't care what the channel identifiers are at all, nor does any client-side code; they just pass the channels through without inspecting them.

haas/api.py
gsilvis: I think erring on the side of too many functions is a good idea.  If it's only used in this function, and we don't think that will have to change, maybe it could be defined right there?

haas/api.py
gsilvis: Oops, lost track of this comment.  The docstring says that it returns the fields, so we need to change something.

haas/api.py
zenhack: Added the check for a non-free node, and I've updated the `api-docs` branch (which there's a pr out for) with the new error accordingly.

zenhack:
@henn, I believe I've either responded to or addressed all of your comments, please have another look.

@gsilvis, you should probably skim over the last couple changes. I'm going to remove the waiting on +2 until you've done so.

haas/api.py
zenhack: Fair enough. Okay if I change the docs, rather than the code?

haas/api.py
zenhack: I'm okay with moving the definition inside the scope of this function --- that will make it easier to find. If we end up wanting it elsewhere, we can always move it again.

haas/api.py
gsilvis: The fields really do seem essential to me---creator tells you if you can delete it, access tells you if it's a public network.  If I've forgotten that information, there's no other way to find it out.  So, I'd prefer to modify the code here.

(I don't remember any conversation where we talked about what needed to be returned here)

haas/ext/network_allocators/vlan_pool.py
gsilvis: No strong opinion---this seems fine to me

haas/api.py
zenhack: Fair enough, I'll change it. We'll have to figure out what to do with
the sentinels when/if we address #463, but we can cross that bridge when
we come to it.

zenhack: Just noticed; The docs also don't mention the channels field (which is in the code); I'll add that as well.

zenhack: Fixed.

gsilvis:
This looks good to me.  +1 again, waiting on @henn's +2

haas/api.py
henn: True. It was more about collecting the channel logic together so that it's easier to follow the channel logic flow.

haas/api.py
henn: Ok. Filed #474

henn:
Getting closer!

Two outstanding things:

1. The comment in driver-model.rst
2. Moving _have_attachment() into node_connect_network() or inlining it

zenhack:
Just pushed (2); thought I'd already done so. (1) has been addressed for a while, but I wrote it right below your comment, so it didn't cause the diff to become "outdated."

henn:
+2; merged

PR #457, opened 2015-07-16
Add -I lanplus to ipmitool invocations

zenhack:
I snuck this in to the hacky branch of mine we're running on the cisco gear; we need to make sure this gets into master before we deploy it there.

### DIFF
--- a/haas/model.py
+++ b/haas/model.py
@@ -148,11 +148,16 @@ class Node(Model):
     def _ipmitool(self, args):
         """Invoke ipmitool with the right host/pass etc. for this node.
 
+        Also adds ``-I lanplus``, which is needed for machines which don't
+        accept the older ``lan`` wire protocol. ``lanplus`` is IPMI version 2
+        only.
+
         `args` - A list of any additional arguments to pass to ipmitool.
 
         Returns the exit status of ipmitool.
         """
         status = call(['ipmitool',
+            '-I', 'lanplus',
             '-U', self.ipmi_user,
             '-P', self.ipmi_pass,
             '-H', self.ipmi_host] + args)


gsilvis:
@zenhack and I talked about this in person.  Our belief is that this will break setups without IPMI 2.0---but currently the 'console_log' functions require IPMI 2.0 anyways, so we aren't losing support for anything.  With that logic, +1

henn:
Could we make this tunable? I'm fine with it being the default.
We still have some IPMI 1.5 boxes that could be used for testing.

zenhack:
I just checked upstairs to confirm my memory; we don't even have enough outlets to plug the 1.5 machines in. Right now all of the outlets are in use and only one of those blades is plugged in. It clearly hasn't been used recently, since the OBM is daisy-chained through 6 other nodes, none of which have power (and therefore it can't actually be reached by the haas master).

I'd actually rather explicitly say no to supporting old hardware unless we think someone is really going to want it. If nothing else, it's extra code to maintain, and I'm doubtful that it will carry it's weight.

In any case, if someone really needs it, once the OBMs have been drivertized, they can add it themselves. I'd rather not maintain something *in core* that's this much of an edge case.

@gsilvis, @lokI8 ?

zenhack:
@henn, @gsilvis and I talked abou the question of whether or not to keep supporting IPMI 1.x (see #457), and basically decided "not now, but maybe later." @lokI8, can you take a pass at this? @henn is really busy with the paper and I don't want to suck up more of his time than I already have before the deadline.

henn:
Just sent @zenhack a PR for some docstring preferences. If that's good with everyone then :+1:

zenhack:
@henn, I actualy really don't like the phrasing in the docstring you added, and I think we should document more prominently that we only support IPMI 2; Let's add that somewhere easier to find (administrators need to care, after all), and then just mention in the docstring that that function also forces the new transport. I think ad a maximum we should add a reference to #475 to the high-level docs; let's not put feature requests in the source.

zenhack:
...It occurs to me that I don't think we really *have* a document where this kind of thing would be placed. Weren't we going to write a guide for site-planning or some such? I remember talking about it.

If nothing else, part of @SahilTikale and @apoorvemohan's drivertization work will involve documenting what each of the drivers does, and so a reference could go there, but I don't know if there's a clear high-level place to put that sort of thing in the meantime?

henn:
> should document more prominently that we only support IPMI 2

A good place to do this might be the [deployment guide](https://github.com/CCI-MOC/haas/blob/master/docs/testing.md).

> let's not put feature requests in the source.

I think it's helpful for future HaaS developers who are unaware of this conversation to know the reason for a random flag.

henn:
@zenhack - did you want to document the IPMI v2+  restriction in the deployment guide?

I'm kindof wondering if we should add a **requirements** section to the README that had this type of stuff.

zenhack:
I made a comment on #480; let's deal with the docs there and just merge this.

gsilvis:
+1 and merging

PR #458, opened 2015-07-17
Database migration script infrastructure.

zenhack:
This builds on top of #455, so it looks bigger than it is (really just the last two commits).

The branch `ucs-migrations` on my fork has a throwaway migration script for the NEU gear, as a reference.


haas/api.py
```diff
@@ -258,34 +275,55 @@ def node_connect_network(node, nic, network):
...
 
 @rest_call('POST', '/node/<node>/nic/<nic>/detach_network')
-def node_detach_network(node, nic):
-    """Detach a physical nic from any network it's on.
+def node_detach_network(node, nic, network):
+    """Detach network ``network`` from physical nic ``nic``.
 
     Raises ProjectMismatchError if the node is not in a project.
+
+    Raises BlockedError if there is already a pending network action.
```
gsilvis: This also raises an error if the network is not attached

zenhack: This is now addressed on the branch on #455

zenhack:
@gsilvis can you have a look at this? just the last two commits are revelevant, the rest is #455.

gsilvis:
I don't know much about alembic.  Could you give me (and another potential reviewer) a short in-person explanation of what's going on here?

zenhack:
I added a document linking to the alembic docs and adding a few extra notes. It may also be useful to look over some actual migration scripts (since this pr doesn't include any).

Here's the one I'm planning on using for the UCS gear (it isn't included since it's migrating from a version that never existed upstream):

https://github.com/zenhack/haas/blob/ucs-migration/alembic/versions/8db1c5a9bbe_.py

I also have a number of these in a repo for a side project of mine. These are generally simpler than the above (and a bit cleaner as well):

https://github.com/zenhack/iron-blogger2/tree/master/alembic/versions

zenhack:
@gsilvis, @lokI8, @henn, can we get this reviewed?

henn:
I like this in principle, since it'll help folks move forward without having to reinstall the whole thing. A few comments:

1. Could you split docs/database-migrations.md into two files:
 * One oriented towards developers, having the instructions necessary for generating the alembic migration scripts, testing them and checking them in (maybe HACKING.rst)
 * One for administrators that gives the instructions for migrating their DB. Maybe a docs/upgrades.md or maybe INSTALL.rst.
1. If you make a new doc file, please link to it from the README so it's easy for folks to find them.
1. Would it make sense to have some sort of test associated with the alembic functionality?

Thanks

zenhack:
@henn, I've addressed (1) & (2) (though see the note in the last commit --- It's getting late, I'll finish that up tomorrow).

re: tests, while it's always nice to have some (and I'm often wishing our test suite was better than it is), I'm actually not sure I think it's a great use of time here. Most of what I've added (besides the docs) is just stock scaffolding set up by `alembic init`, and I don't expect the rest to change very much (my experience with other projects is that it doesn't; writing a comprehensive test suite for the alembic scaffolding itself is probably close to same level of utility as unit testing setup.py). It may be worth doing unit tests for particular migration scripts, but right now we don't have any. @gsilvis, opinion?

alembic/env.py
```diff
@@ -0,0 +1,79 @@
+from __future__ import with_statement
```
gsilvis: Copyright header

zenhack: Quoting George Silvis, III (2015-08-18 16:43:51)

>    Copyright header

Talked about this in person, but... Most of this is just the output of
`alembic init`? Don't know how that works; I only did some small tweaks
to get it to look at haas.cfg for some stuff. I wish I'd done two
commits so it was clear from the history what they were --- oh well.

gsilvis: From what I'm seeing in the FSF's guidelines, I think we should put a basic copy right line (year, 'massachusetts open cloud contributors'), and then a notice that the code was largely autogenerated, to make it clear.  The FSF doesn't list any exception to license-headers for boilerplate files; only one for files under 10 lines.

henn:
re: tests, what about a test that took a pre-binary sqlite DB from a particular version, then checked to see if alembic can bring it up to date with no errors?

The tricky part would then be making exceptions for when we are not guaranteeing compatibility (like perhaps with the OBM work).

alembic/env.py
zenhack: Okay, added the header.

zenhack:
Quoting henn (2015-08-19 11:57:52)

>    re: tests, what about a test that took a pre-binary sqlite DB from a
>    particular version, then checked to see if alembic can bring it up to
>    date with no errors?

Not a bad idea, though I'd make two minor tweaks:

1. Use a textual dump, instead of a binary.
2. Per the docs, we're not actually requiring things to work with
   SQLite, so we should use one of the databases we recommend for
   production.

>    The tricky part would then be making exceptions for when we are not
>    guaranteeing compatibility (like perhaps with the OBM work).

I think this is going to have to be done on a case by cases basis.

It's not clear to me that this idea implies anything added to this
patch; just ideas for things to do when we actually add a script to the
repo?

henn:
> It's not clear to me that this idea implies anything added to this patch; just ideas for things to do when we actually add a script to the repo?

I was thinking that this PR could include a test that verified all alembic files could be applied (maybe we could include a no-op alembic file for now?) as part of the infrastructure, so that when people start including actual changes they'll be tested.

gsilvis:
This looks good to me now.  +1

zenhack:
@henn, I think given that we're not sure how it will generalize, it's worth waiting. We also don't have any migration scripts to test the test infrastructure with.

henn:
> @henn <https://github.com/henn>, I think given that we're not sure how
> it will generalize, it's worth waiting. We also don't have any
> migration scripts to test the test infrastructure with.
>

While I'm usually a big fan of not doing work till it is needed, I think
it's preferable that we a) don't want the first person who happens to
make DB changes to have to write the testing infrastructure and b) don't
want this code to bit-rot from lack of use.

That being said, it may not matter until we have a release, since that's
when we'll need to support upgrades. If that's the case, maybe we can
close this and open a bug that we'll need to solve after the next release.

zenhack:
I don't think we necessarily have any plans to *make* a release. Our one deployment is going to be running master for the forseeable future, and db migration scripts are an important part of how we'll deal with that. I linked above to a script that's been written that *doesn't* make sense to merge, only because it's migrating from something that's never been in mainline. The only practical use of even bumping the version number for our purposes that I can see is to make rpm happy when we start using that for deployment, in which case we'll be doing releases *very* frequently.

I'd like to get this in soon; it's the last thing we're waiting on for the upgrading the gear we have now.

henn:
For the reasons stated above (not wanting the next DB-changer to have to write the test harness and not wanting this code to bit-rot), it would be good if we could have some sort of test infrastructure to accompany this, even if there aren't any DB-changes to apply at the moment.

Re: sqlite vs. postgres, this isn't a sticking point for me. I think using sqlite could have benefits since every developer could test alembic scripts quickly during development without needing to install/configure postgres. Using sqlite would also not preclude adding postgres deployment tests later.

zenhack:
The trouble with SQLite (as mentioned in the docs I wrote) is that it's ALTER TABLE statement is quite limited, so it actually makes writing some kinds of scripts a pain.

How about this: I'll rename the Headnode's dirty column to "mutable," which is  a less confusing name, and write a script and tests around that?

henn:
Coupled with a deployment test, that sounds perfect.
Thanks

gsilvis:
I believe this is the next one we were looking at merging---- @zenhack, it has merge conflicts now.

zenhack:
This is an earlier pass at migrations, and it's somewhat half-baked. The current stuff is in #556; I'm just going to close this.

PR #462, opened 2015-07-28
Begin work on client library

zenhack:
This isn't ready to merge, but I'm worried about forgetting about it, so I want a pr out for it. Feel free to comment on the general design, but no rush.


henn:
Thank you for your code! This functionality is being incorporated as #498.

Would appreciate any comments there!

PR #468, opened 2015-08-03
flesh out API documentation

zenhack:
Everything in rest_api.md is filled out now. There's still some more to do --- I'd like to get rid of the function names, and apidesc.md is still there and pretty much unchanged (i.e. very bitrotted), but I think this is a good self-contained change.

This builds on top of #455, so that should be merged first.

### DIFF
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -351,6 +351,43 @@ Possible Errors:
     * networks
     * headnodes
 
+### project_connect_node
+
+`POST /project/<project>/connect_node`
+
+Request body:
+
+    {
+        "node": <node>
+    }
+
+Reserve the node named `<node>` for use by `<project>`. The node must be
+free.
+
+Possible errors:
+
+* 404, if the node or project does not exist.
+* 409, if the node is not free.
+
+### project_detach_node
+
+`POST /project/<project>/detach_node`
+
+    {
+        "node": <node>
+    }
+
+Return `<node>` to the free pool. `<node>` must belong to the project
+`<project>`. It must not be attached to any networks, or have any
+pending network actions.
+
+Possible errors:
+
+* 404, if `<node>` or `<project>` does not exist, or `<node>` does not
+  belong to `<project>`.
+* 409, if the node is attached to any networks, or has pending network
+  actions.
+
 ### list_projects
 
 `GET /projects`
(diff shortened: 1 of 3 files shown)


haas/api.py
```diff
@@ -602,20 +712,26 @@ def port_connect_nic(port, node, nic):
...
     """
     db = model.Session()
 
-    port = _must_find(db, model.Port, port)
+    switch = _must_find(db, model.Switch, switch)
+    port = _must_find_n(db, switch, model.Port, port)
 
     if port.nic is None:
         raise NotFoundError(port.label + " not attached")
+    if port.nic.owner.project is not None:
```
henn: I can envision scenarios where one might want to change the switch port of an allocated node. For instance, if a port/cable is dead, it could be moved over to another open port.

zenhack: This seemed simpler than enumerating all of the things you need to check if the node is in use. It's also more consistent with the way the rest of HaaS does things. At a minimum, you'd have to make sure there aren't any networks connected (or if you wanted to do the migration transparently, know to reconnect them somehow after the move. That sounds complex). I don't think it's worth it. @gsilvis?

docs/rest_api.md
```diff
@@ -1,8 +1,63 @@
+This file documents the HaaS REST API in detail.
+
+# How to read
+
+Each possible API call had a entry below containing:
```
henn: s/had/has/

zenhack: Quoting henn (2015-08-18 16:36:23)

>    s/had/has/

And further more `s/had a/has an/`. Fixed.

docs/rest_api.md
```diff
@@ -131,59 +230,376 @@ Possible Errors:
...
+        "base_img": <base_img>
+    }
+
+Create a headnode owned by project `<project>`, cloned from base image
+`<base_img>`. `<base_img>` must be one of the installed base images.
+
+Possible errors:
+
+* 409, if a headnode named `<headnode>` already exists
+* 404, if the project does not exist.
```
gsilvis: 404, if the `<base_img>` does not exist

gsilvis:
Suggestion: We can point out that any API call will fail with 404 if you reference a non-existent resource in a non-creation API call.

gsilvis:
+1

lokI8:
Looks good. Thanks for the documentation Ian =) +1

Issue #453, opened 2015-07-10
There needs to be a 'list_networks' command


gsilvis:
The exact semantics are up to debate, but there really should be /something/.  As it is, there's no way of finding the information out, so I'm calling this a bug.

zenhack:
Quoting George Silvis, III (2015-07-10 15:49:03)

>    The exact semantics are up to debate, but there really should be
>    /something/. As it is, there's no way of finding the information out,
>    so I'm calling this a bug.

Yeah, I hit this earlier today. Would also be nice to be able to query 
what a (h)nic is attached to.

-Ian

lokI8:
Or a physical nic. Would be nice to tell if one of the networks got detached from one of my nodes without going into the switch.

zenhack:
Quoting Jonathan Bell (2015-07-15 14:36:59)

>    Or a physical nic. Would be nice to tell if one of the networks got
>    detached from one of my nodes without going into the switch.

I had been thinking in terms of a user figuring out what's connected to 
what network, but yeah, it would also be useful for administrators to be 
able to query what's attached physically.

henn:
I just hit this- if you want to disconnect a NIC from some existing network, you have to know which network it is currently connected to satisfy `node_detach_network`. There is currently no way to do this without peeking in the DB or listing all of a project's networks and trying each possible one.

What do folks think about adding this information to `show_node`?

Instead of:

    $ haas show_node foo
    {"nics": [{"macaddr": "UNKNOWN", "label": "eth0"}], "name": "foo", "free": false}
we could have

    $ haas show_node foo
    {"nics": [{"macaddr": "UNKNOWN", "label": "eth0", "networks": [{"pxe", "vlan/native"}]}], "name": "foo", "free": false}

zenhack:
Quoting henn (2015-07-23 01:20:33)

>    What do folks think about adding this information to show_node?
> 
>    Instead of:
> $ haas show_node foo
> {"nics": [{"macaddr": "UNKNOWN", "label": "eth0"}], "name": "foo", "free": false
> }
> 
>    we could have
> $ haas show_node foo
> {"nics": [{"macaddr": "UNKNOWN", "label": "eth0", "networks": [{"pxe", "vlan/nat
> ive"}]}], "name": "foo", "free": false}

This looks pretty reasonable to me (except of course the fact that 
you've got a syntax error :P).

-Ian

henn:
[...]

PR #469, opened 2015-08-03
Add network info to show_node and add documentation for REST API

zenhack:
This modifies the show_node api call as suggested by @henn's comments in #453.


zenhack:
This includes #455 and #468, so those should be merged first.

haas/api.py
```diff
@@ -822,6 +813,9 @@ def show_node(nodename):
         'free': node.project_id is None,
         'nics': [{'label': n.label,
                   'macaddr': n.mac_addr,
+                  'networks': dict([(attachment.channel,
```
henn: For human parsing, it might be helpful to return everything sorted.

zenhack: It's a dictionary --- it's inherently unordered.

henn: I was thinking of something like https://stackoverflow.com/questions/9001509/how-can-i-sort-a-dictionary-by-key

zenhack: So I played around with this and it impircally seems to work, but I have a strong objection to this: There's nothing in the json module's interface that says it respects the ordering, and it isn't required to --- json objects are unordered. We'd be relying on undocumented implementation details, and we should *never* do that.

henn: I understand it's not required to. As a human interactive user, it's just easier for me to work with a list that is sorted; especially in operations where I think something is connected a certain way and I just want to verify that.

henn: Looks like [json.dump()](https://docs.python.org/2.7/library/json.html#json.dump) supports this with `sort_keys=True`!

zenhack:
@henn, I've addressed your comment.

henn:
:+1: (which is `:+1:`)

zenhack:
@henn, we decided to leave the waiting on reviewer label on a pr even after adding the waiting on +2 (the reasoning being that github doesn't give you a logical or, so removing it makes searching for things that need review more annoying). Just fyi.

gsilvis:
We definitely want to merge something like this, I think... but it has merge conflicts, and the diff looks MUCH larger than it needs to be.  I think one of the other PRs this depended on got changed somehow, so this'll take some untangling.

henn:
This works was picked up in #537. Thank you for your contribution!

PR #477, opened 2015-08-12
Add additional startup code required by #455

zenhack:
We weren't loading extensions (or calling our sanity-check routines), so
calling populate() in `haas init_db` was throwing an exception.

Since we obviously need some tests for basic stuff like this, this patch
adds a few that try to verify that the server-side commands work
correctly.

### DIFF
--- a/haas/cli.py
+++ b/haas/cli.py
@@ -13,7 +13,7 @@
 # governing permissions and limitations under the License.
 
 """This module implements the HaaS command line tool."""
-from haas import config
+from haas import config, server
 from haas.config import cfg
 
 import inspect
@@ -108,14 +108,7 @@ def serve(port):
     # We need to import api here so that the functions within it get registered
     # (via `rest_call`), though we don't use it directly:
     from haas import model, api, rest
-    model.init_db()
-    # Stop all orphan console logging processes on startup
-    db = model.Session()
-    nodes = db.query(model.Node).all()
-    for node in nodes:
-        node.stop_console()
-        node.delete_console()
-    # Start server
+    server.api_server_init()
     rest.serve(port, debug=debug)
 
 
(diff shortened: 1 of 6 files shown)


haas/cli.py
```diff
@@ -136,6 +131,8 @@ def serve_networks():
 def init_db():
     """Initialize the database"""
     from haas import model
+    server.register_drivers()
```
henn: Should we just call server.api_server_init() and add a `stop_consoles=True` default which gets overridden here?

haas/cli.py
```diff
@@ -124,6 +117,8 @@ def serve_networks():
     """Start the HaaS networking server"""
     from haas import model, deferred
     from time import sleep
+    server.register_drivers()
```
henn: Should we just call server.api_server_init() and add a stop_consoles=True default which gets overridden here?

henn:
Other than my comments, looks good.

Have you run the full deployment tests with this?

haas/cli.py
zenhack: Might be reasonable --- we could also pass through the create parameter to the init_db function, which would let us use it everywhere (and maybe rename it to just server_init, since it would no longer only work in the api server).

@gsilvis, what do you think?

zenhack:
Re: tests, the cli module previously had *zero* test coverage --- the deployment tests don't interact with any of the code modified by this patch (nor do any of the other tests, save the ones introduced in this patch).

zenhack:
I made the change to `api_server_init` that @henn suggested. I've also merged in the stuff on #478, since I needed 668b3d9; This may complicate the order in which to merge these, since now this pr has commits on either side of that, but at least it's linear...

henn:
+1

gsilvis:
+1 and merging

PR #478, opened 2015-08-12
Dont call api twice

zenhack:
D'oh

The test suite didn't catch this because it mostly invokes the api calls directly; doesn't go through the http handler.

This is built on top of #477; that should be merged first.

### DIFF
--- a/haas/cli.py
+++ b/haas/cli.py
@@ -13,7 +13,7 @@
 # governing permissions and limitations under the License.
 
 """This module implements the HaaS command line tool."""
-from haas import config
+from haas import config, server
 from haas.config import cfg
 
 import inspect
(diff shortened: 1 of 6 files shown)


zenhack:
@henn, added some docs.

@gsilvis, can you look at this as well?

lokI8:
So I'm guessing something like this https://github.com/giampaolo/psutil isn't worth including for the one run_for_seconds test?

zenhack:
Yeah, that strikes me as overkill --- we don't really need to collect a bunch of statistics, this is just "make sure it doesn't just explode when  you run the command."

henn:
+1

gsilvis:
What's the current interaction between this test and the coverage tool?

PR #480, opened 2015-08-18
Add requirements section to README

henn:
@zenhack, @gsilvis , @okrieg - this is a start to a requirements section for HaaS users.

Thought it would be helpful to stipulate up front what is needed to get an environment going. Suggestions/improvements welcome.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -59,16 +59,13 @@ Required software/hardware for running a production HaaS include:
 
 * Database: a Postgres database server. Sqlite works but is not recommended for production.
 
-Optional components
--------------------
-
-For optional IPMI proxy functionality:
+For IPMI proxy functionality:
 
 * Network access from the HaaS service node to the IPMI interfaces of node under management
 * Nodes that support IPMI v2+
 * A recent version of ipmitool installed on the HaaS service node
 
-For optional headnode functionality:
+For headnode functionality:
 
 * A recent Linux version for the HaaS service node that has libvirt with KVM installed
 * Some number of VM templates
--- a/README.rst
+++ b/README.rst
@@ -36,6 +36,44 @@ administrator must populate initially.
 #. Log in to the headnode, set up a PXE server, reboot the nodes, and deploy an
    operating system on them via the network.
 
+Requirements
+============
+
+Required software/hardware for running a production HaaS include:
+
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -36,6 +36,44 @@ administrator must populate initially.
...
+  * Access to:
+
+    * The Internet or intranet (a way for users to connect to the HaaS service)
+    * The administrative telnet IP on the managed switches
+
+  * Currently only CentOS and RHEL 7.x have been tested, though any node that otherwise meets these requirements should function.
+
+* Database: a Postgres database server. Sqlite works but is not recommended for production.
+
+Optional components
```
zenhack: Do we really want to flag these as optional? You can't currently add nodes without giving IPMI info, and HaaS isn't terribly useful without at least the ability to reboot the nodes...

Headnodes are a bit less essential, but it's a bit awkward given that you can't take the api calls away.

henn: I guess I see these as optional, but the current API doesn't necessarily reflect that. Would you prefer merging in the current optional section?

Re: IPMI, in certain trusted environments, users may have direct access to IPMI anyhow and not require HaaS. Perhaps to make the point, with the new OBM drivers we could include a "null" driver?

Re: Headnodes, until we've completed separating this functionality into its own service we could add a headnodes=False config option that would cause the calls to throw an exception.

zenhack: re: headnodes, I actually think you could get away with just not definining any base images. The api calls would still be there, but there would be no way for a user to actually make a headnode. Still, it's a bit of a hack.

I'd like to just not have any of this tagged as optional for now. We'll update it when @SahilTikale's OBM stuff is in, and we can do so again when we rip out the headnodes.

I think the null OBM driver is a good idea; We seem to be gradually replacing `@no_dry_run` with that sort of thing anyway. We'll also have the recursive OBM driver at some point.

henn:
I believe I have addressed the concerns raised. Appreciate any additional feedback

SahilTikale:
@henn Read the documentation. 
It has two sections Requirements and Documentation.
Documentation seems empty. Is there something else that is supposed to come there. 

Otherwise it looks good for merging.

SahilTikale:
+1

PR #483, opened 2015-08-19
Console rework

zenhack:
There are a couple things that should be done still:

* We need to describe the api in the public docs. #468, has much of the rest fleshed out, so it would perhaps be nice to add it on to that.
* We should document the restriction on what wsgi servers will work, which probably belongs in the documentation that @henn is writing as part of #480.


gsilvis:
@zenhack and I had a conversation a little while ago about this.  Right now, this eats an entire thread for the duration of the time the user is listening to the console.  Since we typically run apache with a fixed number of threads (in testing, exactly one!), this won't work.  We're not yet sure how to fix this problem.

zenhack:
The only thing I can think of is to use greenlets for the request handlers, which we may want to do at some point anyway. This is also important functionality; we have users who want it now. I'm inclined to just:

1. Open an issue for figuring out how to do this with greenlets.
2. Change INSTALL.rst to reccommend a high thread count.
3. Merge it and fix it later.

I am open to better ideas though. @henn, have any thoughts?

henn:
Overall, I think we should prioritize this in terms of our other HaaS tasks.

For now, if we're going to go with this patch I'd suggest we change the installation instructions to specify single-threaded worker processes instead of having worker threads (something like `processes=10 threads=1`, based on [the documentation](https://code.google.com/p/modwsgi/wiki/ProcessesAndThreading)). This should make it so that a single process will only ever run a single task, thus a block in one thread won't affect another. On the technical front, [this stackoverflow](https://stackoverflow.com/questions/3551364/how-do-i-handle-blocking-io-in-mod-wsgi-django) discusses the same problem.

For the workloads we're supporting now, having worker processes instead of worker threads shouldn't have a measurable impact.

zenhack:
I'd be surprised if `time.sleep` didn't release the GIL; it should block *other* threads, so just cranking that number up should work fine as well. I'm not picky though --- when/if we decide we need to sit down to really optimize, I think greenlets are the right way to go

zenhack:
Fixed the merge conflicts (very minor). If others come to a consensus on what to suggest in the docs (high thread count, processes, both...?) I'll tweak the doc and push, should take no time.

@gsilvis, @lokI8, @okrieg, opinions? "don't care" is an acceptable answer, let's us know you've had a chance to voice things.

SahilTikale:
@henn 
We should get this PR in after the FLASK PR
@zenhack do you have any suggestions about the merge conflicts ?

zenhack:
@SahilTikale, I think I still have the keystone auth plugin in the pipeline of stuff I agreed to do. I can come up with an estimate for you guys if you want, and maybe do that after.

My suspicion is that there won't be any deep issues fixing the conflicts; the iirc this is mostly orthonginal to the other changes I made. I would have to look to be sure, however.

//cc @henn, @okrieg, @pns005, @pjd-nu

matsuuran:
@henn @SahilTikale 
Are we still waiting on the flask PR to be merged?  I looked and couldn't see an open PR referencing flask, so just want to see if we can resolve the conflicts and merge this PR

@zenhack 
Was there an offline discussion about using greenlets vs increasing the number of processes?  I checked #487 and it looked like we are hoping on putting greenlets in 0.3.  Just want to make sure I'm tracking

zenhack:
The flask pr was merged a while ago.

re: 0.3, the version numbers don't actually mean anything; we don't have a process for deciding when to do a release, or when what sort of version bump (0.2.1 vs 0.3 vs 1.0) is appropriate, or what the version number actually indicates. So 0.3 doesn't actually convey any information whatsoever about timeframe. I've seen a tendency to just tag things with milestones without discussing it with the team at all, so having a version milestone doesn't even indicate that the devs agree it should be done *at all*. I wrote a longer piece about the situation on another issue a few weeks back; don't have time to dig it up right now.

Issue #431, opened 2015-05-29
use argparse for the cli; make it more flexible


zenhack:
Per my comments in #425, the cli code is somewhat brittle; there's a standard library module `argparse` that we should be using instead of the stuff that I hacked together a year and a half ago.

SahilTikale:
Added support for same, submitted the pull request today  #425

zenhack:
Just a thought: I've been playing with [click][1] lately, and it's pretty nice -- more pleasant to use than argparse IMO. Flask also has some integration with it, though I'm not sure we actually want to use that. Perhaps it's worth considering using that instead of argparse.

[1]: http://click.pocoo.org

zenhack:
In particular, it's handling of subcommands is much nicer.

henn:
FYI - once we have argparse (or something functionally equivalent), we can do stuff like accept the username/password on the command line or have it [automatically read from a file](https://docs.python.org/2/library/argparse.html#fromfile-prefix-chars).

zenhack:
Re: dealing with the CLI, I think one thing we'd talked about is waiting until we had a client library in place (see #617), and building the new CLI on top of that. @henn, what are your thoughts?

shwsun:
My $0.02 is to merge the cli library we are working on first, then talk about this one. We also have to figure out whether or not implementing cli with click is our top priority.

matsuuran:
I wanted to do it because @henn asked me to do #631 and ideally it would just use an optional param.  I'll do whatever though

zenhack:
Quoting (Jethro) Shuwen Sun (2016-09-07 16:48:22)

>    My $0.02 is to merge the cli library we are working on first, then talk
>    about this one. We also have to figure out whether or not implementing
>    cli with click is our top priority.

I think the questions to be asking are:

1. How high priority is 631?
2. Are there any other high-priority issues that depend on this?
3. How close are we to actually being able to merge the client library?
   My memory is "not very."
4. How much work will we save by doing the client library first, rather
   than having to glue the CLI to it later?

shwsun:
Of course, another option is to let @matsuuran start with the new one first, but making sure that it will not break the cli library we will merge later.

@SahilTikale is preempted by the NSDI / BMI work so it might be a little hard for him to finish the work quick. I will bring this up in tmr's hil meeting.

// cc @henn

SahilTikale:
There was hick up with client library ( mostly difference of opinion) which has been settled.
[...]

PR #484, opened 2015-08-19
OBM support for IPMI implemented.

SahilTikale:
Following the 'drivertization' model, added OBM support for nodes.
Changes are as follows:
1. Necessary support is added to support various OBM types in the HaaS database schema
2. Current types of OBM inlcude ipmi, haas(as an obm) and a lego robot driver (for variety of OBM driver)
3. Nodes can be registered using the ipmi information using the api call
4. Example of how to do that is included under examples/obm/insertNodes.command
5. power_cycle method moved from Node to OBM and a specific type implemented under ipmi
6. To ease the development process, automated scripts to start/stop haas servers is included
	under 'examples/devel/' directory


haas/ext/obm/ipmi.py
```diff
@@ -0,0 +1,82 @@
+# Copyright 2013-2014 Massachusetts Open Cloud Contributors
```
zenhack: Should update the end year to 2015

haas/ext/obm/ipmi.py
```diff
@@ -0,0 +1,82 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS
+# IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language
+# governing permissions and limitations under the License.
+
+"""
+A obm driver for implementing lego robot based controls for 
+arm based servers on in a DC at North Pole :-)
```
zenhack: I assume this was a mistake; this is in the IPMI driver.

haas/ext/obm/ipmi.py
```diff
@@ -0,0 +1,82 @@
...
+class Ipmi(Obm):
+    id = Column(Integer, ForeignKey('obm.id'), primary_key=True)
+    ipmi_host = Column(String, nullable=False) 
+    ipmi_user = Column(String, nullable=False)
+    ipmi_password = Column(String, nullable=False)
+
+    api_name = 'http://schema.massopencloud.org/haas/v0/obm/ipmi'
+    __mapper_args__ = {
+        'polymorphic_identity': api_name,
+        }
```
zenhack: Put a blank line after this to separate it from the function below.

haas/ext/obm/ipmi.py
```diff
@@ -0,0 +1,82 @@
...
+    def validate(kwargs):
+        schema.Schema({
+            'type': Ipmi.api_name,
+            'ipmi_host': basestring,
+            'ipmi_user': basestring,
+            'ipmi_password': basestring,
+            }).validate(kwargs)
+
+    def _ipmitool(self, args):
+        """ Invoke ipmitool with the right host/pass etc. for this code.
```
zenhack: s/code/node/

Also, scan over [PEP 257][1] and adjust the formatting of your docstrings accordingly; we follow the conventions described there.

[1]: https://www.python.org/dev/peps/pep-0257/

haas/ext/obm/ipmi.py
```diff
@@ -0,0 +1,82 @@
...
+        `args`- A list of any additional arguments to pass the ipmitool. 
+        Returns the exit status of ipmitool.
+
+        Note: Includes the ``-I lanplus`` flag, available only in IPMI v2+.
+        This is needed for machines which do not accept the older.
```
zenhack: This doesn't look like a complete sentence to me.

haas/ext/switches/switch_haas.py
```diff
@@ -0,0 +1,33 @@
...
+"""A switch driver for recursive HaaS
+It will run the necessary commands on the bHaaS to provide the switch 
+related configuration activities. 
+
+"""
+from sqlalchemy import Column, String, Integer, ForeignKey
+from haas.model import Switch
+
+
+class Switch_haas(Switch):
```
zenhack: You can probably leave the word "switch" out of both the name of the class and the file; it's obvious from the directory/parent class that it's a switch. Maybe call it "recursive."

Also, I'd prefer not to merge this into the repo until it's actually fleshed out; right now you don't have an implementation here.

haas/model.py
```diff
@@ -56,6 +56,9 @@ def init_db(create=False, uri=None):
     Session.configure(bind=engine)
     if create:
         get_network_allocator().populate(Session())
+        #x = get_network_allocator()
+        #print x
+        #x.populate(Session())
```
zenhack: Don't leave stuff commented out when you commit; just delete it.

haas/model.py
```diff
@@ -122,62 +125,12 @@ class Node(Model):
...
-        if self._ipmitool(['chassis', 'power', 'cycle']) == 0:
-            return
-        if self._ipmitool(['chassis', 'power', 'on']) == 0:
-            # power cycle will fail if the machine isn't running, so let's
-            # just turn it on in that case. This way we can save power by
-            # turning things off without breaking the HaaS.
-            return
-        # If it still doesn't work, then it's a real error:
-        raise OBMError('Could not power cycle node %s' % self.label)
+    # The node is initially registered with no nics; see the Nic class.
```
zenhack: Having a wayward comment here doesn't really make sense. I don't think we really need it.

Do the unit tests pass? I feel like at the very least the ones in `tests/unit/model.py` *shouldn't* since they depend on `__init__` having positional arguments. I think removing `__init__` is perfectly sensible, but I suspect there are things that need to be adjusted.

haas/model.py
```diff
@@ -201,6 +154,7 @@ def start_console(self):
     # because we are not interested in the ouput of this command.
     @no_dry_run
     def stop_console(self):
+        return
```
zenhack: So, while I did suggest just doing this for now, I'm not really keen on merging it in like this. We could take a couple of approaches. The console stuff is being reworked in #483, so I don't think it makes sense to spend too much time messing with it on your end.

Perhaps the solution is to get that merged and then adjust this accordingly? We can still review the rest of it in the meantime.

haas/model.py
```diff
@@ -227,6 +182,31 @@ def get_console(self):
     def get_console_log_filename(self):
         return '/var/run/haas_console_logs/%s.log' % self.ipmi_host
 
+class Obm(AnonModel):
+    """ Obm superclass supporting various drivers 
+        related to out of band management of servers """
+    type = Column(String, nullable=False)
+
+    __mapper_args__ = {
+            'polymorphic_identity': 'new_obmType',
```
zenhack: You don't actually need `'polymorphic_identity'` here, because you're never going to be instantiating `Obm` itself, just subclasses. Please remove it.

haas/model.py
```diff
@@ -227,6 +182,31 @@ def get_console(self):
...
+            'polymorphic_on': type
+            }
+    @staticmethod
+    def validate(kwargs):
```
zenhack: This is pretty clearly copypasta from the Switch class. If it contained actual functionality I'd suggest a mixin, but since it's just docs It might make more sense to just put the docs somewhere else and refer to them in both places.

@gsilvis, @henn, thoughts?

haas/model.py
```diff
@@ -227,6 +182,31 @@ def get_console(self):
...
+        Raise a ``schema.SchemaError`` if the  ``kwargs`` is invalid. 
+        Note well: This is a *static* method; it will be invoked on the class.
+        """
+        assert False, "Subclasses MUST override the validate method "
+
+    def power_cycle(self):
+        """
+        Depending on which sub-class is associated with a given node,
+        The exact method of power cycle will differ Implements the power cycle method specific to 
+        each 'TYPE' of obm pertaining to the respective node
```
zenhack: You don't really need to explain the whole inheritance concept here; just explain what the power_cycle method is supposed to do. Just mention in the class's docstring the fact that it's meant to be subclassesd, and which methods need to be overwritten (validate and power_cycle). Have a look at `Switch` for a reference.

haas/model.py
SahilTikale: I do not follow what is the context here ?

--Sahil

haas/ext/obm/ipmi.py
SahilTikale: fixed it in all the obm drivers.

haas/ext/obm/ipmi.py
SahilTikale: yup. I had corrupted my ipmi driver while trying to fix it. So recreated it using the lego.py template. Missed out on changing the docstring. 

Corrected it.

haas/ext/obm/ipmi.py
SahilTikale: done

haas/ext/obm/ipmi.py
SahilTikale: noted. fixed the doc-strings as per the guidelines.

haas/ext/switches/switch_haas.py
SahilTikale: Since tables are created in the HaaS database using the class names, I used the term 'switch' to distinguish base HaaS used as OBM vs Switch.

haas/model.py
SahilTikale: deleted the stuff. Moved the 'Console' related logic from Node to ipmi and made requisite stub methods in the OBM superclass

haas/model.py
SahilTikale: I moved the console related code to ipmi.py under obm. Did necessary changes to server.py and api.py

haas/ext/switches/switch_haas.py
zenhack: Fair enough. I opened #485; once we do that we can change the names. We'll leave it alone for now.

zenhack:
The tests are failing because you've changed the arguments that need to be passed to a number of things. They shouldn't be, please fix them.

docs/registerSwitch_addPorts.commands
```diff
@@ -0,0 +1,30 @@
+Commands to register a switch with HaaS
```
henn: We should move these examples to somewhere easier to find. Perhaps place it in `docs/USING.rst` or `examples/curl_commands` ?

haas/ext/obm/haas_obm.py
```diff
@@ -0,0 +1,34 @@
+# Copyright 2013-2015 Massachusetts Open Cloud Contributors
```
henn: Are we doing recursive HaaS with this submit, or just OBM drivertization? If the latter, then this should be removed.

haas/ext/obm/haas_obm.py
```diff
@@ -0,0 +1,34 @@
...
+""" An obm driver for recursive HaaS.
+It invokes Base HaaS to execute various OBM actions on nodes 
+"""
+
+from sqlalchemy import Column, String, Integer, ForeignKey
+
+from haas.model import Obm
+
+
+class Haas_obm(Obm):
```
henn: Are we doing recursive HaaS with this submit, or just OBM drivertization? If the latter, then this should be removed.

docs/registerSwitch_addPorts.commands
SahilTikale: moved to USING.rst under haas/docs as discussed

henn:
As @SahilTikale and I discussed a little bit ago, we need to work on getting all of the tests to pass for this to make it into upstream HaaS.

Part of that will be adjusting the CLI to function with the new node_register. Since it could be different for each OBM type, we'll need to do something (like add a --type flag). This should be done in conjunction with #431, since that will entail redoing the CLI interface.

@SahilTikale, could you take a look at which tests aren't passing and categorize what work will need to be done to get them passing?

zenhack:
@henn, there aren't any tests that actually exercise the CLI (except the ones I wrote for launching the severs/init_db), so `node_register` isn't going to cause failing tests because of the CLI. (The tests that are there do need to pass, of course). `switch_register` has the same problem (and has already been merged). I'd rather not block all use of structure/optional/varying data in the API on the CLI refactoring (especially since there's a dependency on the client library). I'd be comfortable just taking out those CLI functions and documenting that a few operations aren't supported via the CLI.

docs/USING.rst
```diff
@@ -50,3 +50,41 @@ to hard-disk booting the installed system.
 This is, as the filepath states, merely an example of how you might deploy to
 physical nodes.  Existing deployment systems such as Canonical's MAAS have also
 been run succesfully.
+
+Using CURL Commands
+====================
```
zenhack: Remember, this isn't actually plain text, it's RST. Have a look at the rich text view of this file and you'll see that the formatting on the code samples below doesn't look right; please fix. Github lets you edit/commit files on the website, and it has a preview option, so it's useful for stuff like this.

docs/USING.rst
```diff
@@ -50,3 +50,41 @@ to hard-disk booting the installed system.
...
+ipmi_username:-      ipmiUser4node-02
+ipmi_password:-      ipmiPass4node-02
+
+
+
+For nodes using IPMI use the following api call:
+
+curl -X PUT http://127.0.0.1:5001/node/dummyNoderHaaS-02 -d '
+> {"obm": { "type": "http://schema.massopencloud.org/haas/v0/obm/ipmi",
+> "ipmi_host": "ipmiHost4node-02",
```
zenhack: Nitpick: now that the ipmi fields are encapsulated in their own object which is marked as being of type "ipmi," it's a bit redundant to have `ipmi_user`, `ipmi_password`, `ipmi_host`; you could just have `user`, `password`, `host`.

docs/registerSwitch_addPorts.commands
zenhack: Any reason not to move the rest of the information in this file to USING, and just delete it?

docs/USING.rst
SahilTikale: I left those as they are because they affect how columns-names are created in the database. 
It is good to have a unique (sometimes verbose) column names than common names. It makes life a little easier while debugging.

zenhack: The column names are already scoped by the table though, which is already called IPMI. It strikes me as a bit too much redundancy.

SahilTikale: @henn also raise this point, so I guess there are more people who think that variable names should be simple. I agree and have made the necessary changes. Committed the changes.

docs/USING.rst
```diff
@@ -50,3 +50,70 @@ to hard-disk booting the installed system.
...
 physical nodes.  Existing deployment systems such as Canonical's MAAS have also
 been run succesfully.
+
+Using CURL Commands
+====================
+
+Included herwith are some examples about interacting with HaaS API using the curl 
+utility.
+
+** Example for registering a Node which use IPMI **
```
zenhack: These still aren't showing up as headings; Maybe you need to get rid of the spaces between the stars and the text?

SahilTikale:
Is there something needed to push this upstream

henn:
@SahilTikale Where are we with this one? Could you update the labels?

Thanks

SahilTikale:
This branch can be closed after #526 is accepted and merged with upstream. 

#526 represents the same work using the latest HaaS code-base

henn:
Closing this, as this work is continuing under #526

PR #489, opened 2015-08-27
Add a node_power_off API call.

zenhack:
Might be nice to save some power.

//cc @henn @gsilvis @lokI8

### DIFF
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -13,14 +13,14 @@ Each possible API call has an entry below containing:
   not return any data, in which case this is omitted.
 * A list of possible errors.
 
-In addition to the error codes listed for each API call, HaaS may 
+In addition to the error codes listed for each API call, HaaS may
 return:
 
-* 400 if something is wrong with the request (e.g. malformed request 
+* 400 if something is wrong with the request (e.g. malformed request
   body)
-* 401 if the user does not have permission to execute the supplied 
+* 401 if the user does not have permission to execute the supplied
   request.
-* 404 if the api call references an object that does not exist 
+* 404 if the api call references an object that does not exist
   (obviously, this is acceptable for calls that create the resource)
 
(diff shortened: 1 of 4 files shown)


gsilvis:
What will this return if it's already off?

lokI8:
Apparently  "raise OBMError('Could not power off node %s', self.label)"

gsilvis:
I mostly meant that we should document that.  (It might also be nice to say something like 'the machine is already off' if possible.)

zenhack:
Quoting George Silvis, III (2015-08-28 14:41:11)

>    I mostly meant that we should document that. (It might also be nice to
>    say something like 'the machine is already off' if possible.)

ipmitool seems to return successfully regardless of whether the machine
is already off. I can add something to this effect to the docs.

zenhack:
Docs update.

gsilvis:
+1

gsilvis:
Worth noting:  this change will have to be integrated into #484

zenhack:
Yep. I mentioned that to @SahilTikale already.

henn:
I fully agree that we should have this internally, and potentially run
it automatically when a node is free'd to consume less power/cooling.

That said, I'm not sure how customers would use it for allocated
machines. If someone is paying for a node, when would they want it to be
powered off? Ideally, they would free it and not use it anymore.

We should be careful when adding something to the API, since it will be
difficult to take away later.

lokI8:
I definitely think this is beneficial, seeing as admins and users have access to the same cli... 

It would be nice if this was included in free node, but this is what we have for now, and we don't want to keep machines running. 

I'd like to approve, code looks good, but feel awkward going against Jay

henn:
My objection wasn't strong and consensus seems to be to include this, so merging.

Also - we'll need to add tests for this once we have IPMI tests (see #524).

PR #491, opened 2015-08-28
Initial authentication/authorization (basic)

zenhack:
This is a work in progress, but I thought I ought to submit it so you guys can pick it up in my absence.

Best of luck.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -257,6 +257,43 @@ location specified in ``haas.cfg``) and initialize its tables::
 
   haas init_db
 
+
+Authentication and Authorization
+--------------------------------
+
+HaaS includes a pluggable architecture for authentication and authorization.
+HaaS ships with two authentication backends. One uses HTTP basic auth, with
+usernames and passwords stored in the haas database. The other is a "null"
+backend, which does no authentication or authorization checks. This can be
+useful for testing and experimentation but *should not* be used in production.
+You must enable exactly one auth backend.
+
+Database Backend
+^^^^^^^^^^^^^^^^
+
+To enable the database backend, make sure the **[extensions]** section of
+``haas.cfg`` contains::
+
+  haas.ext.auth.database =
+
+Then initialize the database as described above. You will need to add an
+initial user with administrative privileges to the database in order to
+bootstrap the system. You can do this by running the following command from
+within the directory containing the server's ``haas.cfg``::
+
+  haas create_admin_user <username> <password>
+
+You can then create additional users via the HTTP API. You may want to
+subsequently delete the initial user; this can also be done via the API.
+
+Null Backend
+^^^^^^^^^^^^
+
+To enable the null backend, make sure **[extensions]** contains::
+
+  haas.ext.auth.null =
+
+
 Running the Server under Apache
 -------------------------------
 
--- a/docs/extensions.rst
+++ b/docs/extensions.rst
@@ -25,12 +25,17 @@ section with the same name as the module, e.g.:
     [some_3rd_party.haas.drivers.obm.robotic_power_button_pusher]
     push_duration = 3 seconds
 
-Extensions should not make use any part of the HaaS source tree that do not
-explicitly invite it (e.g. everything by default is *Private*). These currently
-include (see the docstrings in the specific components for details):
+Extensions should not make use any part of the HaaS source tree that does not
+explicitly invite it (i.e. everything by default is *Private*). Components
+which may be used from extensions will explicitly say so in their
+documentation, (and describe in detail how they may be used).
+Extension-approved components currently include:
 
 * Most of haas.network_allocator
+* haas.auth
 * From haas.model:
     * AnonModel
     * Model
     * Switch
+
+See the docstrings for each component for details.
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -11,6 +11,11 @@ Each possible API call has an entry below containing:
 * A human readable description of the semantics of the call
 * A summary of the response body for a successful request. Many calls do
   not return any data, in which case this is omitted.
+* Any authorization requirements, which could include:
+  * Administrative access
+  * Access to a particular project or
+  * No special access
+ In general, administrative access is sufficient to perform any action.
 * A list of possible errors.
 
 In addition to the error codes listed for each API call, HaaS may
@@ -21,7 +26,7 @@ return:
 * 401 if the user does not have permission to execute the supplied
   request.
 * 404 if the api call references an object that does not exist
-  (obviously, this is acceptable for calls that create the resource)
+  (obviously, this is acceptable for calls that create the resource).
 
 Below is an example.
 
@@ -51,6 +56,10 @@ Response Body (on success):
         "numbers": [1,2,3]
     }
 
+Authorization requirements:
+
+* No special access.
+
 Possible errors:
 
 * 418, if `<thing>` is a teapot.
@@ -62,31 +71,10 @@ Possible errors:
 * `{"foo": <bar>, "baz": <quux>}` denotes a JSON object (in the body of
   the request).
 
-# Full API Specification
+# Core API Specification
 
-## Users
-
-### user_create
-
-`PUT /user/<username>`
-
-Request body:
-
-    {
-        "password": <plaintext-password>
-    }
-
-Create a new user whose password is `<plaintext-password>`.
-
-Possible errors:
-
-* 409, if the user already exists
-
-### user_delete
-
-`DELETE /user/<username>`
-
-Delete the user whose username is `<username>`
+API calls provided by the HaaS core. These are present in all
+installations.
 
 ## Networks
 
@@ -103,7 +91,13 @@ Request Body:
     }
 
 Create a network. For the semantics of each of the fields, see
-`docs/network.md`.
+[docs/networks.md](./networks.md).
+
+Authorization requirements:
+
+* If net_id is `''` and creator and access are the same project, then
+  access to that project is required.
+* Otherwise, administrative access is required.
 
 Possible errors:
 
@@ -117,6 +111,11 @@ Possible errors:
 Delete a network. The user's project must be the creator of the network,
 and the network must not be connected to any nodes or headnodes.
 Finally, there may not be any pending actions involving the network.
+
+Authorization requirements:
+
+* If the creator is a project, access to that project is required.
+* Otherwise, administrative access is required.
 
 Possible Errors:
 
@@ -152,6 +151,11 @@ Response body (on success):
         "creator": <project or "admin">,
         "access": <project with access to the network> (Optional)
     }
+
+Authorization requirements:
+
+* If the network is public, no special access is required.
+* Otherwise, access to the project specified by `"access"` is required.
 
 #### Channel Formats
 
@@ -192,6 +196,12 @@ operation to be preformed. Each nic may have no more than one pending
 network operation; an attempt to queue a second action will result in an
 error.
 
+Authorization requirements:
+
+* Access to the project to which `<node>` is assigned.
+* Either `<network>` must be public, or its `"access"` field must name
+  the project to which `<node>` is assigned.
+
 Possible errors:
 
 * 409, if:
@@ -218,6 +228,10 @@ API call returns a status code of 202 Accepted, and queues the network
 operation to be preformed. Each nic may have no more than one pending
 network operation; an attempt to queue a second action will result in an
 error.
+
+Authorization requirements:
+
+* Access to the project to which `<node>` is assigned.
 
 Possible Errors:
 
@@ -247,6 +261,10 @@ example provided in USING.rst
 
 Register a node named `<node>` with the database.
 
+Authorization requirements:
+
+* Administrative access.
+
 Possible errors:
 
 * 409, if a node with the name `<node>` already exists
@@ -256,6 +274,10 @@ Possible errors:
 `DELETE /node/<node>`
 
 Delete the node named `<node>` from the database.
+
+Authorization requirements:
+
+* Administrative access.
 
 ### node_register_nic
 
@@ -271,6 +293,10 @@ Register a nic named `<nic>` belonging to `<node>`. `<mac_addr>` should
 be the nic's mac address. This isn't used by HaaS itself, but is useful
 for users trying to configure their nodes.
 
+Authorization requirements:
+
+* Administrative access.
+
 Possible errors:
 
 * 409 if `<node>` already has a nic named `<nic>`
@@ -280,6 +306,10 @@ Possible errors:
 `DELETE /node/<node>/nic/<nic>`
 
 Delete the nic named `<nic>` and belonging to `<node>`.
+
+Authorization requirements:
+
+* Administrative access.
 
 ### node_power_cycle
 
@@ -295,6 +325,10 @@ PXE. If the node is powered off, this turns it on.
 Power off the node named `<node>`. If the node is already powered off,
 this will have no effect.
 
+Authorization requirements:
+
+* Access to the project to which `<node>` is assigned (if any) or administrative access.
+
 ### list_free_nodes
 
 `GET /free_nodes`
@@ -309,6 +343,10 @@ Response body:
         ...
     ]
 
+Authorization requirements:
+
+* No special access
+
 ### list_project_nodes
 
 `GET /project/<project>/nodes`
@@ -322,6 +360,10 @@ Response body:
         "node-2",
         ...
     ]
+
+Authorization requirements:
+
+* Access to `<project>` or administrative access
 
 ### show_node
 
@@ -343,13 +385,18 @@ The object will have at least the following fields:
 
 Response body:
 
-	{"name": "node1",
+    {"name": "node1",
 	 "project": "project1",
          "nics": [{"label": "nic1", "macaddr": "01:23:45:67:89", "networks": {"vlan/native": "pxe", "vlan/235": "storage"}},
                        {"label": "nic2", "macaddr": "12:34:56:78:90", "networks":{"vlan/native": "public"}}]
 	}
-    
-   
+
+Authorization requirements:
+
+* If the node is free, no special access is required.
+* Otherwise, access to the project to which `<node>` is assigned is
+  required.
+
 ## Projects
 
 ### project_create
@@ -357,6 +404,10 @@ Response body:
 `PUT /project/<project>`
 
 Create a project named `<project>`
+
+Authorization requirements:
+
+* Administrative access.
 
 Possible Errors:
 
@@ -367,6 +418,10 @@ Possible Errors:
 `DELETE /project/<project>`
 
 Delete the project named `<project>`
+
+Authorization requirements:
+
+* Administrative access.
 
 Possible Errors:
 
@@ -390,6 +445,10 @@ Request body:
 Reserve the node named `<node>` for use by `<project>`. The node must be
 free.
 
+Authorization requirements:
+
+* Access to `<project>` or administrative access.
+
 Possible errors:
 
 * 404, if the node or project does not exist.
@@ -406,6 +465,10 @@ Possible errors:
 Return `<node>` to the free pool. `<node>` must belong to the project
 `<project>`. It must not be attached to any networks, or have any
 pending network actions.
+
+Authorization requirements:
+
+* Access to `<project>` or administrative access.
 
 * 409, if the node is attached to any networks, or has pending network
   actions.
@@ -424,6 +487,10 @@ Response body:
         ...
     ]
 
+Authorization requirements:
+
+* Administrative access.
+
 ## Headnodes
 
 ### headnode_create
@@ -440,6 +507,10 @@ Request body:
 Create a headnode owned by project `<project>`, cloned from base image
 `<base_img>`. `<base_img>` must be one of the installed base images.
 
+Authorization requirements:
+
+* Access to `<project>` or administrative access
+
 Possible errors:
 
 * 409, if a headnode named `<headnode>` already exists
@@ -450,6 +521,10 @@ Possible errors:
 
 Delete the headnode named `<headnode>`.
 
+Authorization requirements:
+
+* Access to the project which owns `<headnode>` or administrative access.
+
 ### headnode_start
 
 `POST /headnode/<headnode>/start`
@@ -458,6 +533,10 @@ Start (power on) the headnode. Note that once a headnode has been
 started, it cannot be modified (adding/removing hnics, changing
 networks), only deleted --- even if it is stopped.
 
+Authorization requirements:
+
+* Access to the project which owns `<headnode>` or administrative access.
+
 ### headnode_stop
 
 `POST /headnode/<headnode>/stop`
@@ -465,12 +544,22 @@ networks), only deleted --- even if it is stopped.
 Stop (power off) the headnode. This does a force power off; the VM is
 not given the opportunity to shut down cleanly.
 
+Authorization requirements:
+
+* Access to the project which owns `<headnode>` or administrative access.
+
 ### headnode_create_hnic
 
 `PUT /headnode/<headnode>/hnic/<hnic>`
 
 Create an hnic named `<hnic>` belonging to `<headnode>`. The headnode
 must not have previously been started.
+
+Authorization requirements:
+
+* Access to the project which owns `<headnode>` or administrative access.
+
+Possible errors:
 
 * 409, if:
   * The headnode already has an hnic by the given name.
@@ -482,6 +571,12 @@ must not have previously been started.
 
 Delete the hnic named `<hnic>` and belonging to `<headnode>`. The
 headnode must not have previously been started.
+
+Authorization requirements:
+
+* Access to the project which owns `<headnode>` or administrative access.
+
+Possible errors:
 
 * 409, if the headnode has already been started.
 
@@ -516,6 +611,12 @@ separation. Additionally, headnodes may have an arbitrary number of
 nics, and so being able to attach two networks to the same nic is not as
 important.
 
+Authorization requirements:
+
+* Access to the project which owns `<headnode>` or administrative access.
+* Either `<network>` must be public, or its `"access"` field must name
+  the project which owns `<headnode>`.
+
 Possible errors:
 
 * 409, if the headnode has already been started.
@@ -526,6 +627,12 @@ Possible errors:
 
 Detach the network attached to `<hnic>`.  The headnode must not have
 previously been started.
+
+Authorization requirements:
+
+* Access to the project which owns `<headnode>` or administrative access.
+
+Possible errors:
 
 * 409, if the headnode has already been started.
 
@@ -542,6 +649,10 @@ Response body:
         "<headnode2_name>",
         ...
     ]
+
+Authorization requirements:
+
+* Access to `<project>` or administrative access.
 
 ### show_headnode
 
@@ -566,6 +677,10 @@ Response body:
         "vncport": <port number>
     }
 
+Authorization requirements:
+
+* Access to the project which owns `<headnode>` or administrative access.
+
 ## Switches
 
 ### switch_register
@@ -586,6 +701,10 @@ Request body:
         (extra args; depends on <type>)
     }
 
+Authorization requirements:
+
+* Administrative access.
+
 Possible Errors:
 
 * 409, if a switch named `<switch>` already exists.
@@ -599,6 +718,10 @@ Delete the switch named `<switch>`.
 Prior to deleting a switch, all of the switch's ports must first be
 deleted.
 
+Authorization requirements:
+
+* Administrative access.
+
 Possible Errors:
 
 * 409, if not all of the switch's ports have been deleted.
@@ -609,9 +732,13 @@ Possible Errors:
 
 Register a port `<port>` on `<switch>`.
 
-The permissable values of `<port>`, and their meanings, are switch
-specific; see the documentation for the apropriate driver for more
+The permissible values of `<port>`, and their meanings, are switch
+specific; see the documentation for the appropriate driver for more
 information.
+
+Authorization requirements:
+
+* Administrative access.
 
 Possible Errors:
 
@@ -624,6 +751,10 @@ Possible Errors:
 Delete the named `<port>` on `<switch>`.
 
 Prior to deleting a port, any nic attached to it must be removed.
+
+Authorization requirements:
+
+* Administrative access.
 
 Possible Errors:
 
@@ -642,6 +773,10 @@ Request body:
 
 Connect a port a node's nic.
 
+Authorization requirements:
+
+* Administrative access.
+
 Possible errors:
 
 * 409, if the nic or port is already attached to something.
@@ -652,7 +787,82 @@ Possible errors:
 
 Detach the nic attached to `<port>`.
 
+Authorization requirements:
+
+* Administrative access.
+
 Possible errors:
 
 * 404, if the port is not attached to a nic
 * 409, if the port is attached to a node which is not free.
+
+# API Extensions
+
+API calls provided by specific extensions. They may not exist in all
+configurations.
+
+## The `haas.ext.auth.database` auth backend
+
+### user_create
+
+`PUT /auth/basic/user/<username>`
+
+Request body:
+
+    {
+        "password": <plaintext-password>
+        "is-admin": <boolean> (Optional, defaults to False)
+    }
+
+Create a new user whose password is `<plaintext-password>`.
+
+Authorization requirements:
+
+* Administrative access.
+
+Possible errors:
+
+* 409, if the user already exists
+
+### user_delete
+
+`DELETE /auth/basic/user/<username>`
+
+Delete the user whose username is `<username>`
+
+Authorization requirements:
+
+* Administrative access.
+
+### user_add_project
+
+`POST /auth/basic/user/<user>/add_project`
+
+Request Body:
+
+{
+    "project": <project_name>
+}
+
+Add a user to a project.
+
+Authorization requirements:
+
+* Administrative access.
+
+### user_remove_project
+
+`POST /auth/basic/user/<user>/remove_project`
+
+Request Body:
+
+{
+    "project": <project_name>
+}
+
+Remove a user from a project.
+
+Authorization requirements:
+
+* Administrative access.
+
--- a/haas/api.py
+++ b/haas/api.py
@@ -22,6 +22,7 @@ import json
 from schema import Schema, Optional
 
 from haas import model
+from haas.auth import get_auth_backend
 from haas.config import cfg
 from haas.rest import rest_call, local
 from haas.class_resolver import concrete_class_for
@@ -29,31 +30,10 @@ from haas.network_allocator import get_network_allocator
 from haas.errors import *
 
 
-@rest_call('PUT', '/user/<user>')
-def user_create(user, password):
-    """Create user with given password.
-
-    If the user already exists, a DuplicateError will be raised.
-    """
-    _assert_absent(model.User, user)
-    user = model.User(user, password)
-    local.db.add(user)
-    local.db.commit()
-
-
-@rest_call('DELETE', '/user/<user>')
-def user_delete(user):
-    """Delete user.
-
-    If the user does not exist, a NotFoundError will be raised.
-    """
-    user = _must_find(model.User, user)
-    local.db.delete(user)
-    local.db.commit()
-
-
                             # Project Code #
                             ################
+
+
 @rest_call('GET', '/projects')
 def list_projects():
     """List all projects.
@@ -62,6 +42,7 @@ def list_projects():
 
     Example:  '["project1", "project2", "project3"]'
     """
+    get_auth_backend().require_admin()
     projects = local.db.query(model.Project).all()
     projects = [p.label for p in projects]
     return json.dumps(projects)
@@ -72,6 +53,7 @@ def project_create(project):
 
     If the project already exists, a DuplicateError will be raised.
     """
+    get_auth_backend().require_admin()
     _assert_absent(model.Project, project)
     project = model.Project(project)
     local.db.add(project)
@@ -84,6 +66,7 @@ def project_delete(project):
 
     If the project does not exist, a NotFoundError will be raised.
     """
+    get_auth_backend().require_admin()
     project = _must_find(model.Project, project)
     if project.nodes:
         raise BlockedError("Project has nodes still")
@@ -115,6 +98,7 @@ def project_connect_node(project, node):
     If node is already owned by a project, a BlockedError will be raised.
     """
     project = _must_find(model.Project, project)
+    get_auth_backend().require_project_access(project)
     node = _must_find(model.Node, node)
     if node.project is not None:
         raise BlockedError("Node is already owned by a project.")
@@ -132,6 +116,7 @@ def project_detach_node(project, node):
     BlockedError will be raised.
     """
     project = _must_find(model.Project, project)
+    get_auth_backend().require_project_access(project)
     node = _must_find(model.Node, node)
     if node not in project.nodes:
         raise NotFoundError("Node not in project")
@@ -146,36 +131,6 @@ def project_detach_node(project, node):
     node.obm.stop_console()
     node.obm.delete_console()
     project.nodes.remove(node)
-    local.db.commit()
-
-
-@rest_call('POST', '/project/<project>/add_user')
-def project_add_user(project, user):
-    """Add a user to a project.
-
-    If the project or user does not exist, a NotFoundError will be raised.
-    """
-    user = _must_find(model.User, user)
-    project = _must_find(model.Project, project)
-    if project in user.projects:
-        raise DuplicateError('User %s is already in project %s'%
-                             (user.label, project.label))
-    user.projects.append(project)
-    local.db.commit()
-
-
-@rest_call('POST', '/project/<project>/remove_user')
-def project_remove_user(project, user):
-    """Remove a user from a project.
-
-    If the project or user does not exist, a NotFoundError will be raised.
-    """
-    user = _must_find(model.User, user)
-    project = _must_find(model.Project, project)
-    if project not in user.projects:
-        raise NotFoundError("User %s is not in project %s"%
-                            (user.label, project.label))
-    user.projects.remove(project)
     local.db.commit()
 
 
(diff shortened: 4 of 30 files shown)


zenhack:
@henn, I wrote up a blueprint for the auth stuff as we discussed. It looks like I can't edit the wiki, so I posted it [here](https://gist.github.com/zenhack/74b231d76b4364d0be12) instead.

//cc @okrieg, @pjd-nu, @gsilvis

zenhack:
@henn, I think it's going to be a pain to have a conversation *in* the blueprint, so I'll respond to your comments here:

> Does this mean that every API call will need to be parsed by the auth backend, then re-parsed by the API call itself?

I'm not entirely sure what you mean by "parse" in this case. A couple general points, which might cover it:

* This is still after Werkzeug has made a request object out of the raw HTTP request; in the case of the basic auth, it will just mean fishing a particular header to do the auth check (and then comparing against the database). `authenticate()` will typically be looking at different parts of the request than the api calls themselves (e.g. headers vs path/body), so I wouldn't expect much actual redundancy here.
* `authenticate()` will typically cache some information as a request local, which `require_*` will refer to. The authentication only happens once per call.

If neither of those addresses what you were getting at, please clarify.

> It might make sense to group the API calls by functionality area, since presumably a lot of functionality could be provided by extensions in the future (like keystone auth, various switch-specific things) and it'd be messier to expose all of that to the user. In this case, that could look like /auth/basic or in the future /auth/keystone

I don't think we have a clear picture of what the natural categories would be. The reason I want to split things off into `/ext` is so it's clear that the call isn't guaranteed to be there on all installations. What goes under that I have fewer feelings about, doing `/ext/auth` seemed to make sense since you can't have more than one auth backend registered at a time, so this solves the namespacing problem (at least per install).

My guess is that the keystone backend won't need *any* API calls (just pass the token in the request as usual), so `/auth/keystone` won't need to exist. If you're going based on functionality, then using `/auth/basic` for the other extension would be a misnomer --- the calls are all about creating users changing project memberships. The extension uses basic auth to authenticate, but that's entirely orthogonal to the APIs it exposes.

I'm open to arguments to the contrary, but my feeling is that any attempt at functional-category based division is just going to be a shot in the dark, and will probably turn out wrong. My inclination is still to namespace extension APIs somehow and then let the extensions define the rest.

zenhack:
One more thing for folks to weigh in on, I just updated `docs/rest_api.md` with my understanding of what the authorization requirements for the individual calls should be. I'd like it if we could agree on something for the issues mentioned above, and come to a decision on the stuff I just wrote, so i can get hacking.

//cc @henn, @pjd-nu, @gsilvis, @okrieg

docs/rest_api.md
```diff
@@ -192,6 +200,14 @@ operation to be preformed. Each nic may have no more than one pending
 network operation; an attempt to queue a second action will result in an
 error.
 
+Authorization requirements:
+
+* Access to the project to which `<node>` is assigned.
+* If `<network>` is not public, access to the project named by the network's
+  `"access"` field is required.
+
```
henn: If the network isn't public, then it has to be in the same project as the node (we currently check this).

Maybe something like:
* Node must be in same project as network, if network isn't public.

That could also be placed in `Possible errors:`

docs/rest_api.md
```diff
@@ -192,6 +200,14 @@ operation to be preformed. Each nic may have no more than one pending
 network operation; an attempt to queue a second action will result in an
 error.
 
+Authorization requirements:
+
+* Access to the project to which `<node>` is assigned.
+* If `<network>` is not public, access to the project named by the network's
+  `"access"` field is required.
```
henn: This feels redundant, since both need to be in the same project if the network isn't public.

Perhaps instead move the Note to "Possible errors", appending "or `<network>` must be public"?

docs/rest_api.md
```diff
@@ -258,6 +286,10 @@ Register a nic named `<nic>` belonging to `<node>`. `<mac_addr>` should
 be the nic's mac address. This isn't used by HaaS itself, but is useful
 for users trying to configure their nodes.
 
+Authorization requirements:
+
+* Administrative access.
+
```
henn: Copy this to `node_delete_nic`?

docs/rest_api.md
```diff
@@ -492,6 +590,14 @@ separation. Additionally, headnodes may have an arbitrary number of
 nics, and so being able to attach two networks to the same nic is not as
 important.
 
+Authorization requirements:
+
+* Access to the project which owns `<headnode>`.
+* If the network is not public, access to the project named by it's
```
henn: it's=its

Also, similar comment to above re: moving note to `Possible errors`.

henn:
2 other thoughts (the logic-replication concern is all set I think):

1) What do you think about using a framework (like [flask principal](https://pythonhosted.org/Flask-Principal/)) vs. rolling our own?

2) On namespacing, I still have a slight preference for using /auth, though it's certainly not a blocker. The thought is perhaps someday we could support multiple auth extensions simultaneously, like PAM does.

zenhack:
I'll address your inline comments; everything seems reasonable to me.

re (1): I don't think Flask-Principal does quite what we want; reading the docs it looks like you basically have to enumerate all of the things you have access to up front. There are a lot of cases where we have constraints like "must have access to the project controlling node X," and to make that work we'd have to add each project the user is a member of as a role up front, separately. With the current design, implementations can do whatever they need to do to resolve that. It also doesn't look like it would actually save much (any?) time/effort given where we are in the implementation of this stuff. A good chunk of infrastructure is already in place, and it mostly overlaps with what Flask-Principal does.

We also can't use Flask-Principal without using flask, which we're currently not doing.

While on the topic, I've suggested a few times that it would make sense in the long term to invest some time and port things over to flask. We keep having to tweak the low-level plumbing in order to accomidate things, and it basically amounts to developing and mainting our own web framework. Looking at what's needed to do the keystone auth backend, we're going to want to be able to deal with wsgi middleware in extensions soon, and that's going to involve another bit of redesign.

I think we should stick with the plan on this one, but I want to talk about moving over to flask in general (though not necessarily using every extension...). That probably makes more sense to do in a different venue, particularly since we probably need to get @okrieg and/or @pns005 to weight in on whether that's what I/we should focus on.

re (2): I don't feel *that* strongly about the particular path we use; `/auth` is acceptable, so we can do that if you prefer. It will be easy enough to swap in a different path.

zenhack:
@henn, reworded some things, does this look okay?

henn:
I think we all agree on the framework, so let's move ahead! Since there's agreement, it would be good if we could namespace the operations under `/auth/basic`.

2 questions:
* Do you foresee any issues for this framework being used with keystone?
* Can you think of a way that this could be changed now or later to accommodate finer grained permissions than project level? For instance, if someone wants to share a network with two other projects. I like the simplicity of the API you've proposed, and it may just be that we have to redo things later if we want finer granularity (much like OpenStack).

zenhack:
To address your questions in order:

* We _are_ going to have some things to deal with for keystone, but I'm pretty sure they're orthogonal -- basically we don't have a way of integrating wsgi middleware as of yet. Frankly, I think we should do some initial porting to Flask before we bother tackling that. We probably need to have a more synchronous conversation about that though, either by phone, IRC, or in person. I'm fine with whatever, but if it's anything but IRC we need to make sure to write things down. (Are we logging IRC?).
* For the stuff you proposed in #520, we could have something like:

```python

ok = False
for project in access:
    ok = ok or have_project_access(project):
if not ok:
    raise AuthorizationError()
```

Most things will just use the `require_*` wrappers (and I expect we won't need `have_*` at all for the current design), but splitting them out does give us the flexibility we need. I think we can handle anything that's not a _substantial_ redesign.

zenhack:
@henn, I'm finding that making sure we don't expose the existence of a resource by returning an auth error vs 404 is really finicky. I had a conversation with @gsilvis, and the idea came up of just *alway* returning 404 when there's an auth error, and having the message itself be the same. I'd rather do this than choose a design that's prone to exploitable bugs. What do you think? We could maybe log something at debug level so the information is still there somewhere.

zenhack:
Putting this here so I don't lose it; here's the list of calls that still need to have checks added to them:

* [ ] node_power_cycle
* [ ] list_free_nodes
* [ ] list_project_nodes
* [ ] show_node
* [ ] project_delete
* [ ] project_connect_node
* [ ] project_detach_node
* [ ] headnode_create
* [ ] headnode_delete
* [ ] headnode_start
* [ ] headnode_stop
* [ ] headnode_create_hnic
* [ ] headnode_delete_hnic
* [ ] headnode_connect_network
* [ ] headnode_detach_network
* [ ] list_project_headnodes
* [ ] show_headnode
* [ ] switch_register
* [ ] switch_delete
* [ ] switch_register_port
* [ ] switch_delete_port
* [ ] port_connect_nic
* [ ] port_detach_nic

In the database auth backend:

* [ ] user_create
* [ ] user_delete

zenhack:
Hm, re: api call namespacing, I noticed that the path we had the `project_*_user` calls mapped to is rooted from the url to the corresponding project. Doing the thing we suggested with namespacing would require changing that, which is a bit awkward. One thing we could do is just root it at the user object, which is defined in the same extension, but as far as having a general policy re: namespacing, that's just putting off the inevitable. Thoughts?

// cc @henn @gsilvis.

zenhack:
@henn, just a heads up, the latest commit fixes a bug that affects master; see the commit message for 23eefef.

zenhack:
So it looks like #458 never got merged, and this pr involves adding a column to the `User` table (I think that's all it's going to involve), so we're going to need to migrate the DB. Frankly, adding a test harness to that patch to test migrations is going to be a pretty big chunk of work, and that's before accounting for questions like "What about extensions that have their own tables?"

I don't really think it's reasonable to pull it in as a dependency for this task. I suspect if we decide to do something with flask, that will affect the design, and I don't want to invest more time in throwaway work than we have to.

Here's one idea: I can just right an SQL one-liner, which we can run on the current deployment. We can manually test said one-liner on a copy of the database (made with `pg_dump`), and be done with it. Afterwards, we should start talking about the direction this is going in. I'm going to send an email to the haas mailing list to start the ball rolling on a conversation re: Flask, which has implications for a lot of future refactoring/infrastructure questions.

@henn, @gsilvis, thoughts on the migration issue?

zenhack:
@henn, @gsilvis, This is very nearly done at this point. Things that still need to happen:

1. Figure out the DB migration stuff.
2. Decide what to do about the URL paths for those calls.

See the comments I made recently about these above.

Additionally, someone should run the deployment tests, as I don't have access to an appropriate setup. Also, someone other than me should do some manual testing, just to make sure I don't have important cases in my blind spot, though I was pretty thorough with the unit tests.

I'm putting the waiting on reviewer label on this.

zenhack:
@henn, @gsilvis, ping.

gsilvis:
I vote for the one-line fix here, and working on a real database migration scheme later.

zenhack:
@gsilvis, opinions on the URL mapping issue?

henn:
> 1. Figure out the DB migration stuff.

Let's not worry about migration - we aren't using users currently, so
the table shouldn't have anything in it anyway. Do you know if
SQLAlchemy will automatically sync the new schema and drop the current
Users table?

> 2. Decide what to do about the URL paths for those calls.

I'm in favor of using /auth/basic, but if someone feels strongly we
don't have to use that.
>
> Additionally, someone should run the deployment tests, as I don't have
> access to an appropriate setup. Also, someone other than me should do
> some manual testing, just to make sure I don't have important cases in
> my blind spot, though I was pretty thorough with the unit tests.
>
@SahilTikale, could you put Ian's code on our test cluster and see if it
works as expected?

gsilvis:
I don't think I understand the URL mapping issue you're talking about.  I do think that 'project_add_user' and such belongs in an extension, and `/auth/basic/` seems reasonable for that, though.

zenhack:
@gsilvis, project_add_user is currently mapped to `/project/<project>/add_user`. To move that into `/auth/basic` we'd have to take it out of the project subtree. We could switch it to user_add_project, and put it at `/auth/basic/user/<user>/add_project`.

@henn, Sqlalchemy won't change the schema on its own; that's why database migrations exist. We should be able to just issue an ALTER TABLE statement to add the new column.

zenhack:
I went ahead and moved stuff under `/auth/basic` using the scheme I suggested. Let me know what you guys think.

I tested the following command on a local postgres database, and it worked:

`ALTER TABLE "user" ADD COLUMN is_admin BOOLEAN NOT NULL`

The double quotes are needed around `user` because it's a reserved word in postgres. This only works if the table is empty; the `NOT NULL` constraint combined with the lack of a default value causes it to fail otherwise.

This doesn't work with sqlite, but given that we're deferring a proper migration infrastructure for now and just looking to migrate the one database I don't think we care. I also made arguments in the past against support sqlite for migrations, and I stand by those.

haas/cli.py
```diff
@@ -135,10 +156,19 @@ def init_db():
...
-    do_put(url, data={'password': password})
+def user_create(username, password, is_admin):
+    """Create a user <username> with password <password>.
+
+    <is_admin> may be either "admin" or "no-admin", and determines whether
+    the user has administrative priveledges.
+    """
+    url = object_url('/auth/basic/user', username)
+    if is_admin not in ('admin', 'no-admin'):
+        raise TypeError
```
gsilvis: Should have an error string here, describing the constraint

haas/cli.py
```diff
@@ -401,6 +431,24 @@ def stop_console(node):
...
+    """
+    if not config.cfg.has_option('extensions', 'haas.ext.auth.database'):
+        sys.exit("'make_inital_admin' is only valid with the database auth backend.")
+    from haas import model
+    from haas.ext.auth.database import User
+    model.init_db(create=False)
+    db = model.Session()
+    db.add(User(label=username, password=password, is_admin=True))
+    db.commit()
+
```
gsilvis: Perhaps make_initial_admin should be somewhere other than the haas CLI---many openstack services have a split between client library and -admin tool.  Since we're moving towards splitting the CLI out, we should keep that in mind.

tests/unit/api/auth.py
```diff
@@ -0,0 +1,553 @@
...
+    # project_connect_node: Project tries to connect someone else's node
+    # to itself. The basic cases of connecting a free node are covered by
+    # project_calls, below.
+    (api.project_connect_node, BlockedError,
+     False, 'runway',
+     ['runway', 'manhattan_node_0']),
+])
+def test_auth_call(fn, error, admin, project, args):
+    return auth_call_test(fn, error, admin, project, args)
+
```
gsilvis: Oh dear lord!  I see what you're doing here, and I see why you're doing it... but we should figure out if this is really the clearest way to do it.  If at least we could have something in that long list besides tuples (dicts, perhaps), that might be better.  I'll think about it for a bit...

tests/unit/api/auth.py
```diff
@@ -0,0 +1,553 @@
...
+                   False, 'runway',
+                   args)
+
+
+@pytest.mark.parametrize('fn,args', project_calls)
+def test_manhattan_fail(fn, args):
+    auth_call_test(fn, AuthorizationError,
+                   False, 'manhattan',
+                   args)
+
```
gsilvis: I like this section much better than the one above, FYI

haas/cli.py
zenhack: I agree, but we don't have that split yet. Is there anything you want me to do for this pr?

tests/unit/api/auth.py
zenhack: Yeah, I did the other stuff first. I'll take a second look at it to see if I can't improve it. I think I at least have an idea of how to use a dict.

haas/cli.py
zenhack: We also probably should split the commands for starting the server components from the client cli, but right now we have one entry point for *everything*.

gsilvis: Do we have a bug report for making a client library?  If so, a comment there would be enough

tests/unit/api/auth.py
zenhack: It's dictionaries now, which is a bit better. I also lightened up on the list comprehensions a bit.

haas/cli.py
zenhack: Done, though there's some weirdness around how that exception is used; see the commit messages.

haas/cli.py
zenhack: @gsilvis, we have #498, which I haven't taken much of a look at.

zenhack:
@gsilvis, I think I've responded to everything you mentioned.

zenhack:
@SahilTikale, I went to fix the merge conflicts on this pr (caused due to changes of yours that were merged upstream), and was curious as to the purpose for some of your changes.

Specifically, the commit c0e474 introduces two extra fields to the mock switch model. Why? Switch usernames, passwords, and hostnames are implementation specific details of how to talk to a given switch. With other drivers we may use public key auth, they might be directly connected to the haas master via serial (and thus not need a hostname), and they may need other fields. The *whole point* of drivertizing the switches was to factor these out, and the mock switch has no use for them. From what I can tell it's just extra code for no reason.

Since I don't supply those fields in my tests, my tests are failing. I could update them fairly easily, but the original change strikes me as highly dubious, and I want to give it closer examination.

INSTALL.rst
```diff
@@ -257,6 +257,27 @@ location specified in ``haas.cfg``) and initialize its tables::
 
   haas init_db
 
+
+Authentication and Authorization
```
henn: Could you make this a subsection of an "extensions" section?

INSTALL.rst
```diff
@@ -257,6 +257,27 @@ location specified in ``haas.cfg``) and initialize its tables::
...
+--------------------------------
+
+HaaS includes a pluggible architecture for authentication and authorization.
+Right now, the only authentication backend uses HTTP basic auth, with usernames
+and password stored in the database. to enable this, Make sure the
```
henn: Could you make it more clear that not including an extension will cause auth* to be disabled?

Also, please fix the capitalization on this line.

INSTALL.rst
```diff
@@ -257,6 +257,27 @@ location specified in ``haas.cfg``) and initialize its tables::
...
+
+  haas.ext.auth.database =
+
+Then initialize the database as described above. You will need to add an initial
+user with administrative priviledges to the database in order to bootstrap the
+system. You can do this by running the command::
+
+  haas make_initial_admin <username> <password>
+
+From within the directory containing ``haas.cfg``. You can then create
```
henn: maybe "containing the server's ``has.cfg``"

INSTALL.rst
```diff
@@ -257,6 +257,27 @@ location specified in ``haas.cfg``) and initialize its tables::
...
+additional users via the HTTP API. You may want to subsequently delete the
+initial user; this can also be done via the API.
```
henn: Maybe make this more concise by saying:

"Users can then be managed and created via the HTTP API <add link to API docs>"

haas/cli.py
```diff
@@ -401,6 +431,24 @@ def stop_console(node):
     do_delete(url)
 
 @cmd
+def make_initial_admin(username, password):
```
henn: What do you think of naming this `create_admin_user`?

We use "create" elsewhere and there's no restriction (other than our suggestion/good admin practices) that it only be run initially.

zenhack: The inclusion of the word "initial" was quite purposeful there. You're right that it's "just" good practices, but the interface should encourage that. For consistency I can change it to `create_initial_admin`. If you feel strongly I can get rid of the "initial" bit, but I think it's a good idea to have it in there.

INSTALL.rst
zenhack: I wanted to explicitly include the suggestion to delete the initial admin. Again, if you feel strongly I can change it, but I think that's a good thing to include.

INSTALL.rst
zenhack: Fixed the capitalization. I clarified this section, and documented the presence of the null auth backend, but I actually think it's a really horrible idea to have "no auth" be the default. I added a check (which should have been there before, but I missed it) to verify that *some* auth backend was loaded.

Also, due to the fact that this is something that *must* be configured, I think putting it in an extensions subsection would be too obscure.

zenhack:
@henn, I've either incorporated or replied to each of your comments. Per my comment to @SahilTikale, we need to figure out what to do about those merge conflicts; I could just go through and fix them (wouldn't be too much work), but as stated, I'm highly dubious of the change in the first place.

INSTALL.rst
henn: Thoughts on this?

zenhack: I mentioned in a comment on another line, but configuring auth is mandatory, so relegating it to an extensions section seems unwise; it should be somewhere obvious.

haas/cli.py
henn: I think it'd also be good to see "user" in the name, and create_initial_admin_user seems awkwardly long, so maybe `create_admin_user`?

docs/extensions.rst
```diff
@@ -30,6 +30,7 @@ explicitly invite it (e.g. everything by default is *Private*). These currently
 include (see the docstrings in the specific components for details):
```
henn: I know this wasn't changed as part of this PR, but would you mind clarifying the "These" in "These currently include" to specify whether these **should** or **should not** be modified?

Also - if you are indicating that *only* haas.auth is available as an interface and not children, should that be explicit?

zenhack: I'll do some clarification.

zenhack: Done. Per the documentation, the module itself describes what may be used; the list is just a brief outline, so you know where to find stuff.

haas/cli.py
zenhack: Quoting henn (2016-02-23 14:10:00)

>    I think it'd also be good to see "user" in the name, and
>    create_initial_admin_user seems awkwardly long, so maybe
>    create_admin_user?

I think it's probably pretty clear even without `_user`, and I prefer
having the `initial` or something like it in there (again, encouraging
best practices) also given that the command should really only going to
be used once per setup, I don't think length is a big deal.  However, I
don't feel strongly about it; if you want me to change it to any of the
above I'll do so; let me know.

If `initial` is too long, we could also do `temp` or `tmp`.

SahilTikale:
>>> Specifically, the commit c0e4745 introduces two extra fields to the mock switch model. Why? Switch usernames, passwords, and hostnames are implementation specific details of how to talk to a given switch. 

I added the extra fields because I needed that to test the cli call I was writing for the switch register. 
>>> With other drivers we may use public key auth, they might be directly connected to the haas master via serial (and thus not need a hostname), and they may need other fields. The whole point of drivertizing the switches was to factor these out, and the mock switch has no use for them. From what I can tell it's just extra code for no reason.

>>> I felt that the mock driver should look like some 'real driver' because having no username, password field made it difficult for me test it when I was writing the cli call. 
Its API call was too simplistic and I had to spend more time understanding how to pass arguments to a 'real api call'.
  
I agree that specific implementation of how a feature works will differ for every driver, but I thought that it would make it easy for others developers (such as I) to understand how to call using more arguments if needed. Hence added some extra fields to the mock driver.  

>>> Since I don't supply those fields in my tests, my tests are failing. I could update them fairly easily, but the original change strikes me as highly dubious, and I want to give it closer examination.

I had to change my tests too. It is just two lines, I can pass the test files to you if you wish.

zenhack:
I guess I buy the "boring" argument as far as testing goes. I'd encourage you to update the docs to clarify why that's there; it really wasn't obvious. Like I said, the merge won't actually be difficult; I'll take care of it.

SahilTikale:
@zenhack 
was your lastest response in response to my response ( mouthful, ha ha ).
assuming it was, I did not understand the 'boring' part. 
What docs do you want me to update ?

zenhack:
It was; I was referring to the part of your post where you said:

> Its API call was too simplistic and I had to spend more time understanding how to pass arguments to a 'real api call'.

Re: docs, probably in the docstrings for that switch, you should explain why those fields are there (since they aren't actually used for anything).

zenhack:
Okay, I merged master back in and fixted the test failures. @henn, I'm waiting on you again; Let me know about the admin user command and any other comments.

haas/api.py
```diff
@@ -22,6 +22,7 @@
 from schema import Schema, Optional
 
 from haas import model
+from haas.auth import get_auth_backend
```
henn: Could we call this "auth_backend"?

haas/api.py
```diff
@@ -84,6 +66,7 @@ def project_delete(project):
 
     If the project does not exist, a NotFoundError will be raised.
     """
+    get_auth_backend().require_admin()
```
henn: I don't have strong feelings on this, but do you think a user should be able to delete his/her own project?

haas/auth.py
```diff
@@ -0,0 +1,88 @@
...
+        This is a convienence wrapper around ``have_admin``, and should not be
+        overwritten by subclasses.
+        """
+        if not self.have_admin():
+            raise AuthorizationError("This operation is administrator-only.")
+
+    def require_project_access(self, project):
+        """Like ``require_admin()``, but wraps ``have_project_access()``."""
+        if not self.have_project_access(project):
+            raise AuthorizationError("You do not have access to the required project.")
```
henn: Is there a good reason not to implement the admin check here vs where the check is now in the auth plugin itself?

haas/cli.py
```diff
@@ -79,17 +79,38 @@ def object_url(*args):
...
+        `url` - The url to make the request to
+        `data` - the body of the request.
+
+    If the environment variables HAAS_USERNAME and HAAS_PASSWORD are
+    defined, The request will use HTTP basic auth to authenticate, with
+    the given username and password.
+    """
+    kwargs = {}
+    username = os.getenv('HAAS_USERNAME')
+    password = os.getenv('HAAS_PASSWORD')
```
henn: I think it'd be ideal to support CLI parameters for this, but implementing argparse (or something equivalent) would be out of scope for this PR and is part of #431.

haas/cli.py
```diff
@@ -79,17 +79,38 @@ def object_url(*args):
...
+    if username is not None and password is not None:
```
henn: Does this mean we don't support users with null passwords?

haas/cli.py
```diff
@@ -135,10 +156,19 @@ def init_db():
...
 
 @cmd
-def user_create(username, password):
-    """Create a user <username> with password <password>."""
-    url = object_url('user', username)
-    do_put(url, data={'password': password})
+def user_create(username, password, is_admin):
+    """Create a user <username> with password <password>.
+
+    <is_admin> may be either "admin" or "no-admin", and determines whether
```
henn: Could we replace "no-admin" with something like "basic", "regular" or "non-admin"?

haas/cli.py
```diff
@@ -175,16 +205,16 @@ def list_projects():
     do_get(url)
 
 @cmd
-def project_add_user(project, user):
+def user_add_project(user, project):
```
henn: Could we stick with project\_*\_user?

We have many other project_... calls, so keeping this project_ would seem more consistent.

zenhack: The reason I changed it (which I talked about in a previous comment), is that all of the `project_*` calls map to URLs starting with `/project/<projectname>`. If we want to be consistent with that, we can't also have the URLs start with `/auth/basic`. Having them be `user_*` means we can map them to `/auth/basic/user/<username>`.

This kind of question hints at why I keep saying we need to stop thinking of the procedural (python) API as the primary one --- it keeps us from paying attention to what the HTTP API looks like, and that's important, since we don't want to leave developers trying to work in other languages out in the cold. But I digress...

haas/cli.py
zenhack: I'll change it to regular.

haas/cli.py
zenhack: It is possible to set it to an empty string (e.g. `HAAS_PASSWORD= haas create_headnode ...`), but the environment variable must be defined. If you look at `docs/rest_api.md`, you'll see that the password field in the body of the request for `user_create` is not optional, so there's not actually a way to create a passwordless user, even on the server side. This is unchanged from before (remember, these calls *did* exist, we just weren't really using the user objects for anything, so it didn't matter).

haas/cli.py
zenhack: I agree; my thinking was we'd put it in when we did the cli rewrite.

haas/auth.py
zenhack: I considered this, but I'm not exactly sure the right way to go about it. The awkward bit is that the `have_*` variants are also exposed to the user, so if `require_project_access` also checks for admin access, you could end up in a situation where `require_project_access` and `have_project_access` report different results. The `have_*` variants may be necessary for more complex api calls (like the multi-project network access example you asked about before).

It *is* a bit awkward however. I'm open to suggestions, but I think keeping the API consistent is important.

haas/api.py
zenhack: I don't feel super strongly either. A possible disadvantage is that users could lock themselves out of the system, requiring an admin's assitance to get back in. And not just themselves, but other members of the project as well.

I figure just leave it as is, since neither of us seems to feel strongly.

haas/api.py
zenhack: The main reason I did this is it's consistent with `get_network_allocator()`. I also use a construction `auth_backend = get_auth_backend()` in a number of places, so that should probably change if we were going to change the name.

It also seems like it'd be easy to forget that it's a function without the `get_`; I expect a lot of students forgetting to call it and asking us questions. Shorter is better all else being equal, but you should be careful not to omit important information.

I don't feel strongly. I do have an objection to changing this and not the network allocator one, since I feel the consistency is important.

zenhack:
@henn ping

haas/api.py
henn: > It also seems like it'd be easy to forget that it's a function without the get_; I expect a lot of students forgetting to call it and asking us questions. Shorter is better all else being equal, but you should be careful not to omit important information.

That sounds like a good reason to keep it.

henn: After looking at the code more, I'm starting to think we should establish a convention of using `local` for things that happen all the time, as we did for the DB session handle.

What do you think?

haas/api.py
```diff
@@ -537,19 +518,22 @@ def network_create(network, creator, access, net_id):
...
     _assert_absent(model.Network, network)
 
-    # Check legality of arguments, and find correct 'access' and 'creator'
+    # Check authorization and legality of arguments, and find correct 'access' and 'creator'
     if creator != "admin":
+        creator = _must_find(model.Project, creator)
+        auth_backend.require_project_access(creator)
         # Project-owned network
-        if access != creator:
+        if access != creator.label:
```
henn: nice catch

henn:
For methods like api.py:list_free_nodes(), does not having an auth requirement mean that anyone on the Internet can query that?

If not, can we have a mechanism to require basic authorization before executing any other functions? Since apache won't have access to the HaaS DB, we can't do there.

haas/auth.py
henn: What about having abstract methods for `_have_admin` and `_have_project_access` that need to be overridden, and the base class implement the `have_*` variants?

haas/cli.py
henn: Ya - I'd like `create_admin_user`, with documentation emphasizing that the `user_*` commands should be used after initial setup.

haas/api.py
zenhack: I'm disinclined to mess with it until after the flask refactoring; flask has something akin to our `local`, but the db session has it's own request local outside of that, and there may be a better pattern to be had. `get_*()` gives us the ability to tweak exactly where the thing is stored without having to tear through the code changing everything.

These questions *will* come up during the flask integration, and we'll have to answer this for both the network allocator and the auth backend. Making both of these exposed request locals might be sensible, but let's hold off until the dust settles on that.

haas/auth.py
zenhack: Good call, I'll do that.

haas/ext/auth/database.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+class User(model.Model):
+    """A user of the HaaS.
+
+    A user can be a member of any number of projects, which grants them access
+    to that process's resources. A user may also be flagged as an administrator.
+    """
+
+    is_admin = Column(Boolean, nullable=False)
```
henn: It's redundant, but could you add `,default=False` to the schema definition?

haas/ext/auth/null.py
```diff
@@ -0,0 +1,17 @@
+from haas import auth
```
henn: Could you add a quick module docstring?

haas/cli.py
henn: Interesting - changing this would represent a departure between the CLI and API schemas, though in principle that is ok since the CLI is meant to be procedural and user friendly whereas the API has different design goals.

We have #498 to create a proper client library.

How strongly do you feel about this? I prefer `project_*` since it will be easier for users to find the CLI but won't hold up the PR for it.

haas/cli.py
henn: Have you tested this? Is this a case we should add a unit test for?

haas/api.py
henn: Ok. I added this to newly created issue #540 so that we can track it. I also updated the coding guidelines in #511 to prefer removing redundant stuff.

henn:
Once each of the outstanding concerns are addressed, I can +1.

haas/ext/auth/database.py
zenhack: Sure. Keep in mind you'll want to make sure to do the corresponding SQL thing when doing the migration; the one-liner I gave above doesn't do this.

haas/cli.py
zenhack: I did update the names in the cli to match. I feel pretty strongly about the http semantics being somewhat coherent, and if we're going to namespace the user stuff, we *can't* have them be calls against projects. Let's just leave it as is.

haas/cli.py
zenhack: We don't really have any cli tests to speak of, which has been an outstanding issue for a *long* time, as I'm sure you're aware. We have tests for the general schema validation, and the API call's schema doesn't allow an absent or non-string password field. There's not a test for that exact field/call combination, but I'm pretty comfortable with the choke point that the schema library provides; this isn't a class of bugs that we've had problems with, and there's a reason for that. The trouble with testing is there are always missing cases, but I'm not too worried.

zenhack:
Quoting henn (2016-02-26 15:19:30)

>    For methods like api.py:list_free_nodes(), does not having an auth
>    requirement mean that anyone on the Internet can query that?
>
>    If not, can we have a mechanism to require basic authorization before
>    executing any other functions? Since apache won't have access to the
>    HaaS DB, we can't do there.

Currently, yes, if an API call doesn't do any authorization checks, even
unauthenticated users will be able to use it. I can change this if you
like; would it be better to just raise an error if the user is not
authenticated?

I believe I've addressed everything else; let me know if I've missed
anything.

henn:
> Currently, yes, if an API call doesn't do any authorization checks, even unauthenticated users will be able to use it. I can change this if you like; would it be better to just raise an error if the user is not authenticated? 

Could you make requiring auth a tunable, defaulting to required (and disabled for the dev config)?

zenhack:
@henn, per your request, requiring auth is now tunable, and on by default. I'm not thrilled with the code I've added to `RequestContext.__enter__`, but given that most of that module (and certainly that class) is going to be ripped apart soon, I'm disinclined to do major refactoring on it now. It is at least correct.

zenhack:
@henn ping

henn:
+1, with a slight preference for removing the "You" as stated above (not a show-stopper).

zenhack:
Chopping of th "you" makes it not a complete sentence. How about "Authentication is required..."?

Also, want to wrangle a second reviewer for me?

henn:
> Chopping of th "you" makes it not a complete sentence. How about
> "Authentication is required..."?
>
Sounds good.

> Also, want to wrangle a second reviewer for me?
>
Yep - @kylehogan is working on it.

kylehogan:
I like it!
Everything seems to have been addressed and it looks good.

#520 will require changes to calls like node_connect_network (for many:many project:network relationships) but can deal with that during the implementation of the network ACL 

+1

zenhack:
Neat. If y'all are happy, someone should hit merge.

henn:
Would you mind waiting up to 1 day for #537 to go in first, so that we can sync production and avoid DB changes? If not, I can merge this now.

If we don't have approval on that by tomorrow, I'd merge this anyhow (and cherry-pick later).

zenhack:
It's no sweat off my back.

zenhack:
@henn, actually, before merging, can you make sure someone runs the deployment tests? I don't have the gear for it, and while I didn't touch that stuff *much,* this is a big enough change overall that we should make sure to run them. Travis doesn't get those because it *also* doesn't have the gear.

henn:
Yep. We're in the process of setting up a new deployment test server right now.

zenhack:
@henn, ping. Status?

henn:
@SahilTikale - could you let please us know how things are going with
the deployment tests?

SahilTikale:
Hi Ian, 
I have setup a server in PRB to run deployment tests for various PRs.
Currently I have installed your code from #541. 

When I try to query `list_free_nodes`
I am getting the following error:
`
{"msg": "Authentication failed. Authentication is required to use this service.", "type": "AuthorizationError"}
`
Has the command changed in anyway ?

While installing haas, I did create the admin user as described in the INSTALL.rst

SahilTikale:
Also, I was wondering if there should be a separate document describing the auth backend and how to use it. Some examples to get started in the `USING.rst` would be helpful. 

Let me know if you included the documentation that I missed.

zenhack:
@SahilTikale:

https://github.com/zenhack/haas/blob/11332a687f81253ee94c1b3e4c5451d76a7ef8c4/INSTALL.rst#authentication-and-authorization

There are also comments for the relevant options in the example `haas.cfg` files.

zenhack:
There's also a note in `USING.rst` about how to set usernames and passwords for the command line tool.

SahilTikale:
Setup:
pwd: /var/lib/haas
db: haas.db (sqlite)
user: haas

I ran the command, 
`haas create_admin_user admin admin`
It created a user in the database. I could see the entry in the table `user`

then I configured haas.cfg to point to the postgres database.
did the `haas init_db`
run the command successfully: `haas create_admin_user admin admin`
but this time no such user `admin` is visible if I query the database. 

Also, while running the command `list_free_nodes`
I still face the following error. 

```
[sahil@moc-haas-deploy haas]$ haas list_free_nodes
Unexpected status code: 401
Response text:
{"msg": "Authentication failed. Authentication is required to use this service.", "type": "AuthorizationError"}
```

zenhack:
Everything you describe works for me.

Quoting Sahil Tikale (2016-03-09 23:26:37)

>    Setup:
>    pwd: /var/lib/haas
>    db: haas.db (sqlite)
>    user: haas
>
>    I ran the command,
>    haas create_admin_user admin admin
>    It created a user in the database. I could see the entry in the table
>    user
>
>    then I configured haas.cfg to point to the postgres database.
>    did the haas init_db
>    run the command successfully: haas create_admin_user admin admin
>    but this time no such user admin is visible if I query the database.


What's the exact query you're running? I did what you describe, and then
this showed the expected row:

    SELECT * FROM "user";

Note that the quotes are important; while you can usually leave table
names unquoted, user has special meaning to postgres.

Also, what does your haas.cfg look like? (Obviously censor
credentials/IP addresses if you're going to post).

Also, can you post the output of running:

    \d

At the postgres prompt?

>    Also, while running the command list_free_nodes
>    I still face the following error.
> [sahil@moc-haas-deploy haas]$ haas list_free_nodes
> Unexpected status code: 401
> Response text:
> {"msg": "Authentication failed. Authentication is required to use this service."
> , "type": "AuthorizationError"}


Given that the user isn't being created, that's to be expected. Just to
be sure, you're setting HAAS_USERNAME and HAAS_PASSWORD?

SahilTikale:
@zenhack, 
I wrote you couple of emails. 
Was wondering if you were getting those ?

I'd like to meet you and get clarification on  few things regarding this PR.

I am having difficulties setting up deployment tests for this PR.

SahilTikale:
@zenhack 
While running deployment tests I keep getting this time out error from pexpect module tries to connect to the switch. I had to run tests multiple time until I get lucky and it finishes.

Any ideas on how to fix this behavior ?

SahilTikale:
This PR passes all deployment tests, after I did some minor changes to the tests to reflect change in the code. 

Just make the following changes and I guess we are ready for merge.

in File `haas/tests/deployment/native_networks.py` line 103, remove the extra argument from `_must_find`
do the same thing in  `haas/tests/deployment/native_networks.py` line 127.

zenhack:
Quoting Sahil Tikale (2016-03-15 01:54:06)

>    do the same thing in haas/tests/deployment/native_networks.py line
>    127.

I assume you meant `vlan_networks.py`? In any case, I've fixed both
instances. Thanks for catching these.

Should probably be more pedantic about running these; this is actually
not breakage from the auth stuff, but one of my erlier prs.

zenhack:
@SahilTikale, @henn, I noticed you guys merged #526. I'd like to point out that I specifically asked you guys to wait on that, since it was going to cause merge conflicts with this pr (and it has), and I was annoyed with the prospect of having to deal with those given how much earlier this would have gone in if review had been timely.

What's done is done. The conflicts seem to be fairly straightforward, mostly just tedious (lots of changing calls to node_register in the new tests I added, mostly. A few things in the CLI). @SahilTikale, would you mind fixing those conflicts, and then sending a pr with the fixes to my branch?

SahilTikale:
I added a patch fixing those conflicts. Have a look.

zenhack:
@SahilTikale has fixed the merge conflicts, so this is good to go whenever.

SahilTikale:
Passes all deployment tests. Good to merge in my opinion. 

```
(.venv) [sahil@moc-haas-deploy haas]$ py.test tests/deployment/
=============================================== test session starts ===============================================
platform linux2 -- Python 2.7.5 -- py-1.4.31 -- pytest-2.6.2
plugins: cov, xdist, xdist, xdist
collected 6 items 

tests/deployment/headnodes.py ..
tests/deployment/ipmi.py ..

============================================ 6 passed in 95.04 seconds ============================================
(.venv) [sahil@moc-haas-deploy haas]$ git branch
* auth
  devel
(.venv) [sahil@moc-haas-deploy haas]$ git remote -v
origin	https://github.com/zenhack/haas.git (fetch)
origin	https://github.com/zenhack/haas.git (push)

```

SahilTikale:
@henn @kylehogan 
Please check this and if nothing else, then lets merge this.

henn:
Sounds like deployment tests are passing, per @SahilTikale. Merged!

PR #492, opened 2015-08-31
Haas as an obm

SahilTikale:
This branch has code  that does the following
1. Implement "node_register" call in recursive HaaS
2. Sets up Base HaaS as an Obm driver. 
3. "power_cycle" method implemented for nodes in Recursive HaaS
4. node_discover.py script added to auto-populate nodes from Base haas into recursive HaaS
5. Documentation describing how to configure and use recursive HaaS


examples/haas.cfg.recursive
```diff
@@ -0,0 +1,47 @@
+# This is an example haas.cfg file suitable for use for development without a
+# full set of hardware (switches, nodes, libvirtd...). It can be used as-is,
+# but using the vlan_pool allocator may also be desirable (see the  comments
+# below).
```
gsilvis: This comment needs updating, I think (as well as some of the comments below)

haas/ext/obm/haas_obm.py
```diff
@@ -0,0 +1,78 @@
...
+import logging
+
+
+from haas.model import Obm
+from haas.errors import OBMError
+from haas.dev_support import no_dry_run
+from haas.config import cfg
+
+
+class Haas_obm(Obm):
```
gsilvis: I think this should be HaasObm, by PEP 8

haas/ext/obm/ipmi.py
```diff
@@ -0,0 +1,128 @@
...
+
+from sqlalchemy import Column, String, Integer, ForeignKey
+import schema
+
+
+from haas.model import Obm
+from haas.errors import OBMError
+from haas.dev_support import no_dry_run
+
+class Ipmi(Obm):
```
gsilvis: I'd slightly prefer IpmiObm, given that you have HaasObm above

haas/ext/obm/haas_obm.py
SahilTikale: Currently the 'config parser' converts all class names to lowercase. I have made another pull request #497 that enables camel case. Once this request is accepted I will implement the suggestion made here.

haas/ext/obm/ipmi.py
SahilTikale: Putting it here to avoid repetition of what was discussed between I, George and Jonathan.

We have decided not to put the name IpmiObm and continue with Ipmi. Only in case of HaaS acting as OBM or Switch or some other role on behalf of recursive HaaS we will be using the name HaaS<driverTypeName> and maintain the camel case for the classes.

haas/ext/obm/haas_obm.py
henn: Should this just be `Haas` ?

henn:
This requires some additional updates and merges. Please re-open when it's ready for review.

PR #493, opened 2015-09-02
Autostart headnodes when they are started. Disable autostart when they are stopped.

henn:


gsilvis:
This sounds pretty good to me.  The only thought I have is:  If I, as a user, run 'shutdown -h now' to turn my headnode off, and then the headnode host crashes, my headnode will boot up.  This might not be totally intuitive.

But, I don't have a good solution to this besides having explicit 'autostart' and 'noautostart' commands.

henn:
@gsilvis I agree, but like you don't see a good way around this. To get a clean shutdown, the user needs to run `halt -p`, then `haas headnode_stop` to keep it from coming back.

What do you think about the PR at hand? Is there something better it should be doing? I'd rather not add user-visible API commands to a subsystem that will be going away soon.

henn:
@gsilvis What do you think of my response?

henn:
@gsilvis Thoughts?

gsilvis:
I have no strong opinion on the right behavior here.  I definitely agree that we shouldn't add more API calls for the headnode system.  This code is obviously correct (assuming we want this behavior).  So, I'm going to +1 this, but remove the 'bug' tag.

lokI8:
This is a good solution in case of a headnode machine crash. +1, going to merge

PR #498, opened 2015-10-04
Introduce python client library for HaaS

SahilTikale:
Hi,
This is my first version of the client library for HaaS. Attempt is to make a unified library that can become the basis client programs both cli and gui based. 

I have kept the exceptions and interfaces to the minimum will add and improve based on feedback.
Added meaningful messages for cli calls and formated the output sent by HaaS to the client. 

Once this base code is approved I will add functionality for the api calls that were re-factored during the summer.


henn:
@SahilTikale - could you incorporate some examples of using the client library, perhaps converting @ryabouza's pull requests #319 and #320?

henn:
@SahilTikale - could you take a look at my comments above when you get a chance?

Thanks for your pull request!

ron-unrau:
Sorry, I'm new to github (and python and HaaS and ...), so this may be a little rough. I notice lots of instances of:

except DuplicateName as e:
+        print e.message
+    except IncompleteDependency as e:
+        print e.message
+    except UnknownError as e:
+        print e.message
+

Is it a haas or python standard that each exception has it's own clause? Does the following work?

 except (DuplicateName, IncompleteDependency, UnknownError) as e:
         print e.message

ron-unrau:
In the library API I see names like port_register and stop_console. Fine names, to be sure, but notice that has a noun and a verb (or object and action). Is it important to be consistent, e.g. to always have the object/noun first? If so, stop_console would become console_stop. 

It looks to me like the noun is usually first (head_node_create, user_delete, etc.) but I have to admit I like the sound of show_node better than node_show. Thankfully, it's not up to me to decide :-)

ron-unrau:
Another question about exceptions and exception handling. I notice we never use:

except:

i.e. the wild card. If the library doesn't catch the exception it will go to the caller, correct? Is that what we want? Do we need to document what we handle? Or pass back an error code? Or reraise the exception?

I was looking up exception handling in docs.python.org, and saw something odd (to me):

for arg in sys.argv[1:]:
    try:
        f = open(arg, 'r')
    except IOError:
        print 'cannot open', arg
    else:
        print arg, 'has', len(f.readlines()), 'lines'
        f.close()

I thought the else was odd because execution can only get to the body if there was no exception, so it seems superfluous. My question: is it haas convention to use the else? I just noticed a few functions (e.g. project_delete) that had unprotected code after the exception clauses.

haas/cli.py
```diff
@@ -138,7 +144,7 @@ def serve_networks():
     """Start the HaaS networking server"""
     from haas import model, deferred
     from time import sleep
-    server.init()
+#    server.init()
```
henn: Why comment this out?

haas/cli.py
```diff
@@ -129,6 +134,7 @@ def serve(port):
     # We need to import api here so that the functions within it get registered
     # (via `rest_call`), though we don't use it directly:
     from haas import model, api, rest
+#    server.api_server_init() # Copied from b4cli.py
```
henn: Why comment this out?

If you intend to remove it permanently, just remove it. If folks want to see the way it was, they can use version control :)

haas/cli.py
```diff
@@ -166,25 +172,58 @@ def user_create(username, password, is_admin):
...
 @cmd
 def network_create(network, creator, access, net_id):
-    """Create a link-layer <network>.  See docs/networks.md for details"""
-    url = object_url('network', network)
-    do_put(url, data={'creator': creator,
-                      'access': access,
-                      'net_id': net_id})
+    """Create a link-layer <network>.  See docs/networks.md for details
+            --          --              --              --
+        network: Any string to name the network
```
henn: We don't need to duplicate the documentation from docs/networks.md or docs/rest_api.md here, in accordance with [DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself).

henn:
As part of the CLI changes, could you return a non-zero exit status when an exception is thrown, and 0 when there's success?

haas/cli.py
```diff
@@ -166,25 +172,58 @@ def user_create(username, password, is_admin):
...
 def network_create_simple(network, project):
     """Create <network> owned by project.  Specific case of network_create"""
-    url = object_url('network', network)
-    do_put(url, data={'creator': project,
-                      'access': project,
-                      'net_id': ""})
+
+    try:
+        ret = h.network_create_simple(network, project) 
+        print "    SUCCESS: Network %s created " % network
```
henn: I would suggest changing these to omit the Success/Fail, and just return the meat (like `Network %s created`).

henn: Also, there's no need to leading whitespace

haas/cli.py
```diff
@@ -195,8 +234,18 @@ def user_delete(username):
 @cmd
 def list_projects():
     """List all projects"""
-    url = object_url('projects')
-    do_get(url)
+    try:
+        ret = h.list_projects()
+        print "Success: Total %d projects found: " % len(ret())
+#        print type(ret())
+        for i in ret():
```
henn: Should `ret()` be a function call?

haas/cli.py
```diff
@@ -213,49 +262,110 @@ def user_remove_project(user, project):
...
+        print e.message
 
 @cmd
 def project_delete(project):
     """Delete <project>"""
-    url = object_url('project', project)
-    do_delete(url)
+    try:
+        ret = h.project_delete(project)
+        print "Success: project %s deleted. " % project
```
henn: Please trim trailing whitespace

haas/cli.py
```diff
@@ -213,49 +262,110 @@ def user_remove_project(user, project):
...
 @cmd
 def headnode_create(headnode, project, base_img):
     """Create a <headnode> in a <project> with <base_img>"""
-    url = object_url('headnode', headnode)
-    do_put(url, data={'project': project,
-                      'base_img': base_img})
+
+    try:
+        ret = h.headnode_create(headnode, project, base_img)
+        print "Success: Headnode Created for project  %s " % project
```
henn: trailing whitespace (please apply this widely)

haas/cli.py
```diff
@@ -213,49 +262,110 @@ def user_remove_project(user, project):
...
 @cmd
 def project_connect_node(project, node):
     """Connect <node> to <project>"""
-    url = object_url('project', project, 'connect_node')
-    do_post(url, data={'node': node})
+
+    try:
+        ret = h.project_connect_node(project, node)
+        print "Success: Node %s connected to project %s " % (node, project)
+    except DuplicateName as e:
```
henn: Instead of duplicating these exceptions in every function, why not surround the overall CLI dispatch function with them?

haas/cli.py
henn: In thinking about this some more, maybe we should not print anything on success, and print the error on failure?

henn: Excepting the info functions (`list_*`, `show_*`) of course!

haas/cli.py
```diff
@@ -441,68 +628,169 @@ def port_detach_nic(switch, port):
...
 @cmd
 def list_project_networks(project):
     """List all networks attached to a <project>"""
-    url = object_url('project', project, 'networks')
-    do_get(url)
+
+    try:
+        ret = h.list_project_networks(project)
+        print "SUCCESS: Total networks with Project %s: %s " % (project,len(ret()))
+        for i in ret():
```
henn: I think you mean `ret` not `ret()`.

Also - for these types of functions, I'd be fine with dumping the json. What do others think?

haas/client.py
```diff
@@ -0,0 +1,801 @@
...
+from haas.rest import APIError, ServerError
+    # Do we need ServerError ? It is not supposed to know 
+    # what happens on the server side
+import requests
+import os
+import urllib
+import json
+
+
+#Return Codes:
```
henn: Return codes for what?

haas/client.py
```diff
@@ -0,0 +1,801 @@
...
+
+        If endpoint is None, use the endpoint specification from:
+            1) The HAAS_ENDPOINT env variable or
+            2) Take the endpoint from [client] endpoint
+
+        Exceptions:
+            LookupError - no endpoint could be found
+        """
+        config.load(filename='haas.cfg')
+        if endpoint != None:
```
henn: We're not preferring HAAS_ENDPOINT over the config file like the comment above specifies.

haas/client.py
```diff
@@ -0,0 +1,801 @@
...
+        "vlan/native")
+
+    Errors are thrown when receiving HTTP status_codes from the HaaS server
+    that are not in the [200,300) range
+    """
+
+    def __init__(self, endpoint=None):
+        """Initiatlize an instance.
+
+        If endpoint is None, use the endpoint specification from:
```
henn: Maybe we should place this documentation in the "haas help" output?

haas/client.py
```diff
@@ -0,0 +1,801 @@
...
+        r = requests.put(url, data=json.dumps(data))
+
+        if (200 <= r.status_code < 300):
+            return                    #Operation successful
+        elif (r.status_code == 409):
+            raise DuplicateName("FAILURE: Name already exists" )    #Username already exists
+        elif(500 <= r.status_code):
+            raise ServersideError("FAILURE: Server failed to process the request ")
+        else:
+            raise UnknownError("FAILURE: Contact the HaaS Admin. Unknown error occured !!! ")      #Operation failed
```
henn: Please remove unnecessary whitespace

haas/client.py
```diff
@@ -0,0 +1,801 @@
...
+
+    def network_create(self, network, creator, access, net_id):
+        """Create a link-layer <network>. See docs/networks.md for details."""
+        url = self.object_url('network', network)
+        data = {'creator': creator, 'access': access, 'net_id': net_id}
+        r = requests.put(url, data=json.dumps(data))
+
+        if (200 <= r.status_code < 300):
```
henn: Same comment here as in the CLI: it might make sense to check the status code in the dispatcher function once for all of the client calls, then throw exceptions as necessary from there.

haas/client.py
```diff
@@ -0,0 +1,801 @@
...
+        else:
+            raise UnknownError("FAILURE: Contact the HaaS Admin. Unknown error occured !!! " )     #Operation Failed 
+
+
+
+
+
+
+
+#    def node_connect_network(self, node, nic, network, channel="vlan/native"):
```
henn: Why comment this out?

haas/client.py
```diff
@@ -0,0 +1,801 @@
...
+            raise DuplicateName("FAILURE: Name already exists" )      #Network-name already exists
+        elif(r.status_code == 404):
+            raise IncompleteDependency("FAILURE: does not meet dependency ")
+        elif(500 <= r.status_code):
+            raise ServersideError("FAILURE: Server failed to process the request ")
+        else:
+            raise UnknownError("FAILURE: Contact the HaaS Admin. Unknown error occured !!! ")      #Operation failed
+
+
+    def node_register(self, node, ipmi_host, ipmi_user, ipmi_pass):
```
henn: Does this need a type?

haas/error_clientlib.py
```diff
@@ -0,0 +1,27 @@
...
+class DuplicateName(Exception):
+    """ An exception indicating that the name of the entity already exists. """
+
+class IncompleteDependency(Exception):
+    """ Some calls require the input argument to be objects that are already.
+
+    present in the database. If the objects used as input are missing in the database, 
+    this exception will be raised.
+    """
+
```
henn: Extra whitespace

SahilTikale:
I am working a different 'smaller branch'. Will submit that soon.

matsuuran:
@SahilTikale anything going on with this?

SahilTikale:
This was made long time ago, then we revisited the design. Also now I am making smaller pull requests. 
I still need this PR for reference. Let it be here. I will close it later.

SahilTikale:
Work moved to 
https://github.com/CCI-MOC/hil/pull/617

SahilTikale:
closing this.

PR #501, opened 2015-10-21
list_network() and Show_node for network

xuhang57:
Changed the api.py and cli.py for creating a new call: list_network() and have a unit test under tests/unit/api.py 

Also testing Ian's pull request for #469. Show_node() passes the unit test cases. 

One concern here.  "deferred.apply_networking()" is used in the script in api.py unit test. I know it is essential, but I don't know why. When I used the cli to test the show_node(), I got the error saying:

sqlalchemy.exc.UnboundExecutionError: Could not locate a bind configured on mapper Mapper|User|User or this Session

And I cannot figure it out so far.

PR #504, opened 2015-10-30
List_Network and Show_node

xuhang57:
A new PR trying to avoid the merge error.  Hopefully, this is a merge-able version 
It passed all tests in the tests/unit 

Will add vlan_id in the list Network as Jay mentioned in the PR #501 
But not sure if it should be within the list_network function since, as Jay said, it is already in the show_network()

show_node now is showing the network info. Thinking about adding a port info in there. 
And for now, it doesn't support for adding port to a network via using the UI.


tests/unit/api.py
```diff
@@ -1224,23 +1245,22 @@ def test_show_node(self, db):
...
             ],
         }
         self._compare_node_dumps(actual, expected)
 
-
-    def test_show_node_unavailable(self, db):
-        api.node_register('robocop', 'ipmihost', 'root', 'tapeworm')
-        api.node_register_nic('robocop', 'eth0', 'DE:AD:BE:EF:20:14')
-        api.node_register_nic('robocop', 'wlan0', 'DE:AD:BE:EF:20:15')
-
```
xuhang57: deleting it because the call is simply doesn't exist in the api.py.

xuhang57:
**TO Do:**
1. For list_networks, add channels info to the output

haas/api.py
```diff
@@ -534,6 +534,16 @@ def headnode_detach_network(headnode, hnic):
...
+@rest_call('GET', '/networks')
+def list_networks():
+    """List all networks.
+    Returns a JSON array of strings representing a list of projects.
+    Example:  '["network1", "network2", "network3"]'
+    """
+    db = model.Session()
+    networks = db.query(model.Network).all()
+    networks = [n.label for n in networks]
+    return json.dumps(networks, sort_keys=True)
```
xuhang57: Along with the networks, I will add channels info in the output

xuhang57:
Please ignore  **add uuid/base_image to show_headnod**, it should be in this PR. And I've already deleted that.

tests/unit/api.py
```diff
@@ -1196,6 +1197,19 @@ def test_list_projects(self, db):
...
+        assert json.loads(api.list_networks()) == []
+        api.project_create('anvil-nextgen')
+        network_create_simple('eth0', 'anvil-nextgen')
+        assert json.loads(api.list_networks()) == ['eth0']
+        network_create_simple('wlan0', 'anvil-nextgen')
+        network_create_simple('wlan1', 'anvil-nextgen')
+        assert sorted(json.loads(api.list_networks())) == [
+            'eth0',
+            'wlan0',
+            'wlan1',
```
henn: These names sound like NICs. Could you change them to be something more networky, like "netA"?

haas/api.py
```diff
@@ -534,6 +534,17 @@ def headnode_detach_network(headnode, hnic):
...
+def list_networks():
+    """List all networks.
+    Returns a JSON array of strings representing a list of projects.
+    Example:  '["network1", "network2", "network3"]'
+    """
+    db = model.Session()
+    networks = db.query(model.Network).all()
+    network_labels = [n.label for n in networks]
+    network_ids = [n.network_id for n in networks]
+    return json.dumps([network_labels, network_ids], sort_keys=True)
```
henn: This will output two parallel lists. I think it would be better to return a list of the JSON blobs returned in show_network().

If implementing the show_network() logic, it would be nice if we could rename "access" to "project", and if there is a project to output it but if not (ie- it's NULL) to output something like "PUBLIC".

tests/unit/api.py
```diff
@@ -1475,4 +1499,4 @@ def test_node_power_cycle(self, db):
         api.project_create('anvil-nextgen')
         api.node_register('node-99', 'ipmihost', 'root', 'tapeworm')
         api.project_connect_node('anvil-nextgen', 'node-99')
-        api.node_power_cycle('node-99')
+        api.node_power_cycle('node-99')
```
henn: I think that little arrow in the github diff is indicating that there isn't a final newline. Some tools break when that happens.

tests/unit/api.py
henn: Which call doesn't exist?

henn:
Thanks for your code, Lucas! I think it looks good other than the things I pointed out.

tests/unit/api.py
xuhang57: Will do!

tests/unit/api.py
xuhang57: i will try to get rid of it asap

SahilTikale:
@henn Is this pull request and #469 trying to do the same thing.

tests/unit/api.py
SahilTikale: @henn That little arrow is simply complaining that there is no newline at the end of file.
@xuhang57 A simple enter should do the trick.

SahilTikale:
Hi @xuhang57  I tested this pull requests. 
Looks OK to me. But the suggestions mentioned by @henn do seem to be implemented.
If you can update this pull request with said changes we can merge this pull request soon enough.

SahilTikale:
@henn I just one question. The call "show node" provides information about which networks a nic is attached to. It says "networks" is there any case in which a single nic of node will be connected to multiple networks ? Otherwise I suggest @xuhang57 that it should be changed to just "network" and not "networks"

tests/unit/api.py
xuhang57: show_node_unavailable

henn:
@SahilTikale RE: network vs networks, one can indeed have multiple networks connected to a NIC via [channels](https://github.com/CCI-MOC/haas/blob/45dae53798786be08a2e57092edc0e51da31fffd/docs/rest_api.md#show_network).

xuhang57:
@henn @SahilTikale  Please see PR #519 for the latest update on list_networks function. 

Sorry about the delay.

henn:
@xuhang57 - should we close this pull request? If so, could you please do so. Thanks!

henn:
@xuhang57 Are you planning to implement the show_node() functionality somewhere?

Thanks

xuhang57:
@henn 
I believe show_node is already in our code base, isn't it? And I will close it asap, just want to make sure nothing valuable left here.

PR #505, opened 2015-11-04
Make database sessions request-local (address #454)

### LINKED ISSUE #454: Database session should be request-local
This would simplify a large portion of the control flow. Even in spite of the refactoring changes I'm doing, it's still a little awkward to get at the session. The session could be created in `haas.rest`.

@gsilvis?

zenhack:
This fixes #454. There are still a couple TODOs though. I think it's worth doing a lot of that cleanup; if we  don't do it now it will never happen. See the bullet points in the last commit message. Also, I'm unable to run the deployment tests, so while I *think* I've made the necessary adjustments, someone else should check.

### DIFF
--- a/haas/api.py
+++ b/haas/api.py
@@ -23,7 +23,7 @@ from schema import Schema, Optional
 
 from haas import model
 from haas.config import cfg
-from haas.rest import rest_call
+from haas.rest import rest_call, local
 from haas.class_resolver import concrete_class_for
 from haas.network_allocator import get_network_allocator
 from haas.errors import *
@@ -35,7 +35,7 @@ def user_create(user, password):
 
     If the user already exists, a DuplicateError will be raised.
     """
-    db = model.Session()
+    db = local.db
     _assert_absent(db, model.User, user)
     user = model.User(user, password)
     db.add(user)
@@ -48,7 +48,7 @@ def user_delete(user):
 
     If the user does not exist, a NotFoundError will be raised.
     """
-    db = model.Session()
+    db = local.db
     user = _must_find(db, model.User, user)
     db.delete(user)
     db.commit()
@@ -64,7 +64,7 @@ def list_projects():
 
     Example:  '["project1", "project2", "project3"]'
     """
-    db = model.Session()
+    db = local.db
     projects = db.query(model.Project).all()
     projects = [p.label for p in projects]
     return json.dumps(projects)
@@ -75,7 +75,7 @@ def project_create(project):
 
     If the project already exists, a DuplicateError will be raised.
     """
-    db = model.Session()
+    db = local.db
     _assert_absent(db, model.Project, project)
     project = model.Project(project)
     db.add(project)
@@ -88,7 +88,7 @@ def project_delete(project):
 
     If the project does not exist, a NotFoundError will be raised.
     """
-    db = model.Session()
+    db = local.db
     project = _must_find(db, model.Project, project)
     if project.nodes:
         raise BlockedError("Project has nodes still")
@@ -119,7 +119,7 @@ def project_connect_node(project, node):
 
     If node is already owned by a project, a BlockedError will be raised.
     """
-    db = model.Session()
+    db = local.db
     project = _must_find(db, model.Project, project)
     node = _must_find(db, model.Node, node)
     if node.project is not None:
@@ -137,7 +137,7 @@ def project_detach_node(project, node):
     If the node has network attachments or pending network actions, a
     BlockedError will be raised.
     """
-    db = model.Session()
+    db = local.db
     project = _must_find(db, model.Project, project)
     node = _must_find(db, model.Node, node)
     if node not in project.nodes:
@@ -162,7 +162,7 @@ def project_add_user(project, user):
 
     If the project or user does not exist, a NotFoundError will be raised.
     """
-    db = model.Session()
+    db = local.db
     user = _must_find(db, model.User, user)
     project = _must_find(db, model.Project, project)
     if project in user.projects:
@@ -178,7 +178,7 @@ def project_remove_user(project, user):
 
     If the project or user does not exist, a NotFoundError will be raised.
     """
-    db = model.Session()
+    db = local.db
     user = _must_find(db, model.User, user)
     project = _must_find(db, model.Project, project)
     if project not in user.projects:
@@ -198,7 +198,7 @@ def node_register(node, ipmi_host, ipmi_user, ipmi_pass):
 
     If the node already exists, a DuplicateError will be raised.
     """
-    db = model.Session()
+    db = local.db
     _assert_absent(db, model.Node, node)
     node = model.Node(node, ipmi_host, ipmi_user, ipmi_pass)
     db.add(node)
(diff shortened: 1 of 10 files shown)


zenhack:
Note: I've got another branch `use-local.db` built on top of this one where I've removed the session parameters from the helpers in `haas.api`. My inclination is to basically remove the session parameters from *everything.*

// cc @henn, @gsilvis

haas/rest.py
```diff
@@ -180,6 +160,20 @@ def _format_arglist(*args, **kwargs):
...
+    """Context manager that sets up `local` for use in the request handler.
+
+    This is used internally by the request handler, but is exposed to make
+    testing easier.
+    """
+
+    def __enter__(self):
+        local.db = Session()
+
+    def __exit__(self, exc_type, exc_value, trackeback):
```
henn: think you meant "traceback"

zenhack: Quoting henn (2015-11-05 20:46:24)

>    think you meant "traceback"

Good call, fixed (rebased).

zenhack:
Decided to just add the bits to use `local.db` to this pr. I'm waiting on you guys at this point.

zenhack:
@henn, @gsilvis, ping.

henn:
+1

zenhack:
One thing I specifically wanted commentary on: I put the test in `tests/stress.py`, which is outside the usual `tests/unit`. This makes some organizational sense to me, but we've been in the habit of running `py.tests tests/unit`, to avoid hitting the deployment tests as well. There could be a danger of people forgetting to run stuff if it's outside units.

One Idea is to tweak setup.cfg such that `py.test` (which no arguments) does the right thing.

henn:
What do you think of including it as part of the deployment tests?

I think of unit tests as something to run that's quick and local, and deployment tests as the more involved ones we should run to be sure the whole thing works. Stress seems to fit that profile, and having just 2 targets would ensure that the stress tests are run occasionally.

zenhack:
So I've typically tested against postgres on my laptop, and used SQLite when on one of the test boxes, since it makes it easier to ensure a clean database. So having in deployment would make that marginally more annoying. I'd be fine either way in the short term though.

Longer term I think we ought to be governing this sort of thing with pytest's skip/xfail stuff:

http://pytest.org/latest/skipping.html

I'm actually of the understanding that,  having drivertized the switches, most of the deployment tests should work without a full setup either, so long as you use one of the mock switch drivers. I think only the headnode tests are still an issue. @gsilvis, can you confirm (and also weigh in)?

zenhack:
Can we make a decision re: the location of the test and get this merged? @henn, @gsilvis.

gsilvis:
Test location looks fine to me

haas/api.py
```diff
@@ -333,10 +319,10 @@ def node_detach_network(node, nic, network):
 
     Raises BadArgumentError if the network is not attached to the nic.
     """
-    db = model.Session()
-    node = _must_find(db, model.Node, node)
-    network = _must_find(db, model.Network, network)
-    nic = _must_find_n(db, node, model.Nic, nic)
+    db = local.db
```
gsilvis: any reason this line is here?

haas/api.py
```diff
@@ -555,8 +531,8 @@ def network_create(network, creator, access, net_id):
     Details of the various combinations of network attributes are in
     docs/networks.md
     """
-    db = model.Session()
-    _assert_absent(db, model.Network, network)
+    db = local.db
```
gsilvis: ditto

haas/api.py
```diff
@@ -927,23 +883,22 @@ def stop_console(nodename):
     ####################
 
 
-def _assert_absent(session, cls, name):
+def _assert_absent(cls, name):
```
gsilvis: Maybe document that this function, and the following ones, require you to be in request context

tests/deployment/headnodes.py
```diff
@@ -40,10 +40,17 @@ def server_init():
     server.validate_state()
 
 
+@pytest.yield_fixture
```
gsilvis: This is a little opaque... what's going on here?

tests/unit/api.py
```diff
@@ -664,8 +672,8 @@ def test_headnode_delete_hnic_success(self, db):
         api.headnode_create('hn-0', 'anvil-nextgen', 'base-headnode')
         api.headnode_create_hnic('hn-0', 'hn-0-eth0')
         api.headnode_delete_hnic('hn-0', 'hn-0-eth0')
-        api._assert_absent(db, model.Hnic, 'hn-0-eth0')
-        hn = api._must_find(db, model.Headnode, 'hn-0')
+        api._assert_absent(model.Hnic, 'hn-0-eth0')
+        hn = api._must_find(model.Headnode, 'hn-0')
```
gsilvis: Hm.  I wonder if this test is missing a line at the end... it's weird that it ends by setting a variable

gsilvis:
Looks mostly pretty good.  The only real request I have is to document that those functions now only work in a request context

henn:
Having a separate stress test module is fine. I reiterate my +1.

Jason

gsilvis:
For the few remaining functions that take a 'db' parameter---I'm not sure whether it's right to make them use local.db instead.  It means that they too can only be used in request context, but maybe that's fine.

haas/api.py
zenhack: Probably left over from an intermediate step (where I just replaced all of the `db = model.Session()` with this). I'll remove it.

haas/api.py
zenhack: Good call, will do.

tests/unit/api.py
zenhack: Yeah, that's odd. My guess is it's just sloppy coding, and the point of that statement it just to make sure the headnode is still in there. Wouldn't it be nice if our test suite was more readable?

In any case, it's orthogonal to this patch.

haas/api.py
zenhack: (After refreshing my memory): I left it in here in a few places as a shorthand, where it makes sense and where it doesn't is a bit subjective.

tests/deployment/headnodes.py
zenhack: pytest's yield fixtures let you abuse generators to do cleanup (rather than having to explicitly register finalizers). The test gets run where the yield statement is. RequestContext is a context manager, which seemed the safest way to make sure we actually were cleaning up after ourselves, but it makes explicit cleanup callbacks... awkward. The yield fixture solves the control flow issue presented there.

haas/api.py
gsilvis: Oops, misread the following lines.  Yes, that makes sense.  (I thought the variable was /never/ used.)

tests/deployment/headnodes.py
gsilvis: Hm, interesting.  If this is pretty standard, then okay.  Otherwise maybe comment it

tests/unit/api.py
gsilvis: Agree that it's orthogonal

zenhack:
@gsilvis, I added the docs you requested. I'm comfortable deferring any possible future removal of that parameter.

gsilvis:
Agree on deferring.  +2 and merging

PR #506, opened 2015-11-08
Hn uuid

xuhang57:
Add UUID and Base_img info to the show_headnode function. However, cannot find the column of the name of the virtual machine in the data base. 

For the unit test case, I am having trouble of retrieving the 'uuid'. Since there is no any function such as get_headnode_uuid(), I cannot check whether the show_headnode() outputs a correct uuid. Seeking for help to solve this issue.

### DIFF
--- a/haas/config.py
+++ b/haas/config.py
@@ -25,7 +25,7 @@ import importlib
 import sys
 
 cfg = ConfigParser.RawConfigParser()
-
+cfg.optionxform = str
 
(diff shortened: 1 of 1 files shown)


xuhang57:
Also, I am having a hard time for switching the branch. Trying to create a new branch using git checkout -b hn-uuid, and make some changes over there. However, after I commit the changes, I cannot push the code upstream to my own repo.

PR #507, opened 2015-11-09
add uuid/base_img to show_headnode

xuhang57:
Added uuid and base_mg to show_headnode here.

I cannot have it output the virtual machine name for now, since, according to Sahil, I need to setup libvirt first and then I can do something with the name of virtual machine in show_headnode(). 

**For the unit test**, I am having trouble for write a unit test for the uuid in show_headnode() since uuid is a unique id and I cannot hard-code it in the unit test script for testing. 

Any suggestion would be much appreciated.

### DIFF
--- a/haas/api.py
+++ b/haas/api.py
@@ -877,7 +877,9 @@ def show_headnode(nodename):
         'project': headnode.project.label,
         'hnics': [n.label for n in headnode.hnics],
         'vncport': headnode.get_vncport(),
-    })
+        'uuid' : headnode.uuid,
+        'base_imgs': headnode.base_img,
+    }, sort_keys = True)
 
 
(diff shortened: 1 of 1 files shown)


henn:
+1

SahilTikale:
I tested the code. Just by copying the lines from this api.py to mine, it works as expected.
But when I pulled the branch and tried to rerun the test it is giving errors.

@xuhang57 have you changed anything else in the api.py. other than adding the two lines. 
api.py for this branch is broken.

henn:
@SahilTikale Any update on this based on our conversations yesterday?

SahilTikale:
Hi Tested the pull request. It works as intended. 
+1 from me.

PR #517, opened 2016-01-22
Fix unit test failure. Closes #512

### LINKED ISSUE #512: show_headnode test suite failure
```
___________________________________________________________________________ TestQuery.test_show_headnode ___________________________________________________________________________
self = <api.TestQuery instance at 0x7f727cf78560>, db = <sqlalchemy.orm.session.Session object at 0x7f727cf51c10>

    def test_show_headnode(self, db):
        api.project_create('anvil-nextgen')
        network_create_simple('spiderwebs', 'anvil-nextgen')
        api.headnode_create('BGH', 'anvil-nextgen', 'base-headnode')
        api.headnode_create_hnic('BGH', 'eth0')
        api.headnode_create_hnic('BGH', 'wlan0')
        api.headnode_connect_network('BGH', 'eth0', 'spiderwebs')
    
    
        result = json.loads(api.show_headnode('BGH'))
        # For the lists to be equal, the ordering must be the same:
        result['hnics'].sort()
>       assert result == {
            'name': 'BGH',
            'project': 'anvil-nextgen',
            'hnics': [
                'eth0',
                'wlan0',
            ],
            'vncport': None
        }
E       assert {'base_imgs':...nextgen', ...} == {'hnics': ['et...ncport': None}
E         Omitting 4 identical items, use -v to show
E         Left contains more items:
E         {u'base_imgs': u'base-headnode',
E          u'uuid': u'bce1ed76-b948-11e5-b2b0-001e65327848'}

tests/unit/api.py:1335: AssertionError
```

This looks like it was due to a recent change. About that CI...

Also, what was the reasoning behind putting the uuid in the output? It was originally just there as an implementation detail, to make sure that the libvirt names were unique; exposing it as part of the API doesn't make a lot of sense to me.

henn:
Address #512

### DIFF
--- a/tests/unit/api.py
+++ b/tests/unit/api.py
@@ -1346,6 +1346,7 @@ class TestQuery:
         api.project_create('anvil-nextgen')
         assert json.loads(api.list_project_nodes('anvil-nextgen')) == []
 
+    import uuid
 
     def test_show_headnode(self, db):
         api.project_create('anvil-nextgen')
@@ -1357,11 +1358,18 @@ class TestQuery:
 
 
         result = json.loads(api.show_headnode('BGH'))
+
+        # Verify UUID is well formed, then delete it, since we can't match it
+        # exactly
+        temp = uuid.UUID(result['uuid']) # Verify UUID is well-formed
+        del result['uuid']
+
         # For the lists to be equal, the ordering must be the same:
         result['hnics'].sort()
         assert result == {
-            'name': 'BGH',
-            'project': 'anvil-nextgen',
+            'name':     'BGH',
+            'project':  'anvil-nextgen',
+            'base_img': 'base-headnode',
             'hnics': [
                 'eth0',
                 'wlan0',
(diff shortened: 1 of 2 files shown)


henn:
@zenhack from #512, You're not wrong about exposing something for the admins to all users. My thought right now is that we'll hopefully be splitting out the headnode service and so this will all become a moot point.

Perhaps when auth makes it in, we can make the uuid a field only shown to admins.

knikolla:
Looks Ok to me. Testing the validity of the UUID before deleting it makes sense as a malformed one (even though it shouldn't happen) will throw an exception, making the test fail.

gsilvis:
+2, merging

tests/unit/api.py
```diff
@@ -1346,6 +1346,7 @@ def test_no_project_networks(self, db):
         api.project_create('anvil-nextgen')
         assert json.loads(api.list_project_nodes('anvil-nextgen')) == []
 
+    import uuid
```
zenhack: Why is this imported way down here, instead of at the top? Noticed this while merging back into my auth branch for #491.

PR #519, opened 2016-01-22
List_Networks function.

xuhang57:
Similarly to PR #504 
This is for the update of list_networks function.

The unit test part still needs time to be optimized. But, the function call should works well for now.


haas/api.py
```diff
@@ -510,6 +510,19 @@ def headnode_detach_network(headnode, hnic):
...
+def list_networks():
+    """List all networks.
+    Returns a JSON array of strings representing a list of projects.
+    Example:  {"channels": ["101", "107"], "name": ["net-05", "qnet-00"]}
+    """
+    db = model.Session()
+    networks = db.query(model.Network).all()
+    result = {
+        'channels': [n.network_id for n in networks],
+        'names': [n.label for n in networks],
```
henn: Similar comment as to what I posted in #504, I think it would be cleaner for clients to parse to return a single list of either tuples or a dict, where channels are paired with their name, than the current set of two lists.

Also, we should return the project of each network, since this could only presumably be run by admins. It would be nice if we could rename "access" to "project", and if there is a project to output it but if not (ie- it's NULL) to output something like "PUBLIC".

As a dict, this would look something like:
```python
results = {}
for n in networks:
  net = {'channel':n.network_id, 'project':n.project}
  results[n.label] = net

return json.dumps(results)
```

henn:
@xuhang57 - could you please update the unit tests to match the new output and address continuous integration failure?

haas/api.py
xuhang57: Thanks for the useful note. I am bit confused on how to show project within list_network. We currently don't have a n.project to query projects that connected to this network. I think we do have access_id which accentually just project_id, right? 

For changing "access" to "project", it needs me to change the unit test as well, and it causes lots of failure. I will try to fix them.

xuhang57:
I've updated the unit test. But I don't understand this continuous-integration error. Does this relates to unit test at all?

knikolla:
@xuhang57 Seems like there are 2 error during unit test. 

```
assert json.loads(api.list_networks()) == []
E       assert {} == []
```
json_loads returns a dictionary, but the unit test expects it to return a list. Maybe change the square braces in the unit test to curly braces should solve that. 

```
>       assert result == {
            'name': 'public-network',
            'creator': 'admin',
            'channels': ['null'],
        }
E       assert {'access': 'P...blic-network'} == {'channels': [...blic-network'}
E         Omitting 3 identical items, use -v to show
E         Left contains more items:
E         {u'access': u'Public'}
```
result contains one more entry in the dictionary in the dictionary. Maybe doing this in the unit test will solve it. 
```
assert result == {
            'name': 'public-network',
            'creator': 'admin',
            'channels': ['null'],
            'access': u'Public'
        }
```

xuhang57:
@knikolla  Thanks for the excellent tips. I've fixed the two errors and please let me know if there is any other problem.

haas/api.py
```diff
@@ -510,6 +510,23 @@ def headnode_detach_network(headnode, hnic):
                             # Network Code #
                             ################
 
+@rest_call('GET', '/networks')
+def list_networks():
+    """List all networks.
+    Returns a JSON array of strings representing a list of projects.
+    Example:  {"channels": ["101", "107"], "name": ["net-05", "qnet-00"]}
+    """
```
gsilvis: This docstring isn't right.  Currently, you return a JSON dictionary, indexed by the network names, whose values are dictionaries.  The example is also incorrect.

haas/api.py
```diff
@@ -510,6 +510,23 @@ def headnode_detach_network(headnode, hnic):
...
+    db = model.Session()
+    networks = db.query(model.Network).all()
+    result = {}
+    for n in networks:
+        if n.access is not None:
+            net = {'channel': n.network_id, 'project': n.access.label}
+        else:
+            net = {'channel': n.network_id, 'project': 'Public'}
```
gsilvis: Nitpick, but if there's a project named "Public" this will be ambiguous.  Maybe do ``'public': True`` instead?

haas/api.py
```diff
@@ -611,6 +628,8 @@ def show_network(network):
 
     if network.access is not None:
         result['access'] = network.access.label
+    else:
+        result['access'] = 'Public'
```
gsilvis: Same comment as at line 526

haas/api.py
xuhang57: @henn What do you think Jay?

tests/unit/api.py
```diff
@@ -1204,6 +1204,15 @@ def test_list_projects(self, db):
             'runway',
         ]
 
+    def test_list_networks(self, db):
+        assert json.loads(api.list_networks()) == {}
+        api.project_create('anvil-nextgen')
+        network_create_simple('netA', 'anvil-nextgen')
+        result = json.loads(api.list_networks())
+        assert result == {
+            'netA': {'project': anvil-nextgen, 'channel':["null"]}
```
gsilvis: typo: anvil-nextgen should be inside a string

xuhang57:
Differing items:
E         {'netA': {'channel': '43c0481a-cab5-11e5-9e43-3aee41243083', 'project': 'anvil-nextgen'}} != {'netA': {'channel': ['null'], 'project': 'anvil-nextgen'}}

pns005:
@henn Do you see anything else that need to be done here ?
Can we close on this pull request ?

haas/api.py
```diff
@@ -510,6 +510,25 @@ def headnode_detach_network(headnode, hnic):
                             # Network Code #
                             ################
 
+@rest_call('GET', '/networks')
+def list_networks():
+    """List all networks.
+    Returns a JSON dictionary, indexed by the name work names whose values
+    are ditionaries with index as channel and project name.
```
lokI8: Better doc string for the first two lines: 

Returns a JSON dictionary of dictionaries, indexed by the name of the network.
The interior dictionaries are indexed by project and channel.

xuhang57:
Thanks @lokI8  and please let me know if you find other errors.

lokI8:
@xuhang57 It looks like you just deleted channel from the response and are no longer checking it in the assert... can you try using network_create instead of network_create_simple? The test is to verify the full return value, not just 1/2...

haas/api.py
```diff
@@ -510,6 +510,25 @@ def headnode_detach_network(headnode, hnic):
...
+    The interior dictionaries are indexed by project and channel.
+    
+    Example:  {"netA": {"channel": "101", "project": "qproj-01"}}
+    """
+    db = model.Session()
+    networks = db.query(model.Network).all()
+    result = {}
+    for n in networks:
+        if n.access is not None:
+            net = {'channel': n.network_id, 'project': n.access.label}
```
knikolla: @lokI8 Should channel point to network_id (which is a uuid) or to the vlan number?

xuhang57: @lokI8 I forgot to mention that, Ravi and I've discussed on what channel should point to this afternoon and he would like it point to the network_id. Should I add the vlan number for it? 

Thanks

haas/api.py
```diff
@@ -510,6 +510,25 @@ def headnode_detach_network(headnode, hnic):
...
+@rest_call('GET', '/networks')
+def list_networks():
+    """List all networks.
+    Returns a JSON dictionary of dictionaries, indexed by the name of the network.
+    The interior dictionaries are indexed by project and channel.
+    
+    Example:  {"netA": {"channel": "101", "project": "qproj-01"}}
+    """
+    db = model.Session()
+    networks = db.query(model.Network).all()
```
henn: The db should be accessed through local.db (see the above function) to prevent session leakage.

Thus, the `db = ...` line should be deleted and this line should look like `networks = local.db.query(...`

haas/api.py
henn: I believe `channel` should be set to `n.channel`.
Documentation for channel formats can be found in docs/rest_api.md (search for `channel`).

haas/api.py
henn: Sorry for the delay - we just had some discussion on this in #537. We decided to set project to the null entry (`None` in python) for objects without projects, so I think you should be able to use very similar logic here:
```python
'project': None if n.creator_id is None else n.creator.label
```

haas/api.py
```diff
@@ -510,6 +510,25 @@ def headnode_detach_network(headnode, hnic):
...
+    result = {}
+    for n in networks:
+        if n.access is not None:
+            net = {'channel': n.network_id, 'project': n.access.label}
+        else:
+            net = {'channel': n.network_id, 'project': 'Public'}
+        result[n.label] = net
+
+    return json.dumps(result)
```
henn: Please add `sort_keys=True`

haas/api.py
henn: ditto :-)

haas/api.py
```diff
@@ -611,6 +630,8 @@ def show_network(network):
 
     if network.access is not None:
         result['access'] = network.access.label
+    else:
+        result['access'] = 'Public'
 
     return json.dumps(result)
```
henn: Please add `sort_keys=True`. Not sure if it's needed, but since we're in the code anyway, might as well be consistent.

henn:
Sorry again for the delay, @xuhang57. Thanks for your work on this!

Other than the comments I made, the only other thing is that we should document list_networks() in `docs/rest_api.md`, similar to what we're doing in #537.

xuhang57:
@henn  No worries!

I just add some words on the `docs/rest_api.md` for list networks function. And please let me know if there is any problem. Thanks

tests/unit/api.py
```diff
@@ -1204,6 +1204,21 @@ def test_list_projects(self, db):
             'runway',
         ]
 
+    import uuid
+
+    def test_list_networks(self, db):
+        assert json.loads(api.list_networks()) == {}
+        api.project_create('anvil-nextgen')
+        network_create_simple('netA', 'anvil-nextgen')
```
SahilTikale: Please add a test to output a network not assigned to any project. 
See how that shows up in the output. 
Just add another network using `network_create` instead `network_create_simple`.

henn: Reading docs/networks.md and docs/rest-api.md should give you the info you need on what the different network creation arguments mean.

xuhang57:
@SahilTikale  Thanks for your review. I just pushed the file with adding a new test under the `test_list_networks`. Please let me know if there is any problem on it.

Thanks

tests/unit/api.py
xuhang57: Thanks Jay. @henn  I just pushed it. Please take a look. Thanks

haas/api.py
henn: Woops - the above statement was from looking at the `networkattachment` table, which is different.

Maybe let's rename `channel` to `driver_id`, since the ID is driver-specific and is *not* a channel. Currently, they are VLANs, but they could be something else in the future. @gsilvis @lokI8 @SahilTikale - what do you think?

docs/rest_api.md
```diff
@@ -124,6 +124,19 @@ Possible Errors:
     * The network is connected to a node or headnode.
     * There are pending actions involving the network.
 
+### list_networks
+
+`GET /networks`
+
+List all networks
+
+The result contains a JSON dictionary of dictionaries, indexd by the name of the network. Then interior dictionaries are indexed by project and channel.
```
henn: typo: indexd = indexed.

Also - could you describe this using the terms "key" and "value", like we do in #537? Something like:

> The result contains a JSON dictionary of dictionaries. The first-level dictionary has keys equal to the names of the networks, with values being another dictionary containing the keys \`channel\` and \`project\`.

henn:
Thanks @xuhang57.

Could you please check out my other code comments? I think we're very close!

haas/api.py
xuhang57: I think I will leave it for now.

xuhang57:
@henn @SahilTikale @lokI8 @knikolla  Just push the fix suggested from the comments. Please let me know if there is anything I should fix. Thanks for your time.

gsilvis:
Superseded by #554

Issue #364, opened 2014-11-19
Need upstart/systemd script for haas serve_networks


henn:
The install doc currently says to run:
    haas serve_networks &

This means that we're depending on the user's terminal to host a long-lived process. We should provide a startup script to run the daemon in the background. CentOS 6 supports upstart, CentOS 7 is systemd.

henn:
The output should also get recorded somewhere. Currently, it's not so if the daemon dies you have to guess why (unless it reproducibly keeps dying :)

zenhack:
Quoting henn (2015-07-25 02:00:29)

>    The output should also get recorded somewhere. Currently, it's not so
>    if the daemon dies you have to guess why (unless it reproducibly keeps
>    dying :)

Systemd captures stdout/stderr by default, so writing a service script
for it would do the trick on systemd based systems.

henn:
FYI - `serve_networks` seems to die sometimes on some haas installs. This feature would make it much easier to debug!

henn:
Closed by #525

PR #525, opened 2016-02-01
Systemd haas

SahilTikale:
Addresses #364.

Scripts located in haas/scripts folder

Consists of two separate commits.
1. scripts for dev environment
2. Network script for production environment

Updated INSTALL.rst and HACKING.rst respectively.
@henn @gsilvis 
Please review and comment.


INSTALL.rst
```diff
@@ -304,13 +304,31 @@ You should also set apache to start on boot::
...
 
-  haas serve_networks &
 
-as the HaaS user. To make this happen on boot, add the following to ``/etc/rc.local``::
+A systemd script for running the network server is available in the 'scripts' directory.
+Name of the script is: haas_network.service
+
+Centos uses systemd to controll all its processes. 
+Ubuntu does not use systemd by default. To use the following script
+on ubuntu server, you will have to install systemd.
```
henn: Newer versions of ubuntu [come with systemd by default](https://wiki.ubuntu.com/SystemdForUpstartUsers).

scripts/dev_haas.service
```diff
@@ -0,0 +1,30 @@
+[Unit]
+# Change the port and the username to match your enviorment
+Description=HaaS Dev Server listening on port 5000, for user sahil
+After=network.target
+Wants=
+
+[Service]
+#Change the username your username. same for group
+User=sahil 
+Group=sahil
```
henn: Maybe make the default User/Group "user"?

We could also make them "ubuntu", which matches the cloud image

scripts/dev_haas.service
```diff
@@ -0,0 +1,30 @@
...
+
+#Provide absolute path to where Haas is installed.
+WorkingDirectory=/home/sahil/git/haas/
```
henn: Similar to above, let's make this more generic

scripts/haas_network.service
```diff
@@ -0,0 +1,19 @@
+[Unit]
+Description=HaaS Network Server 
+After=network.target
+Requires=httpd.service
```
henn: Please remove the dependency on the httpd service, as there's no requirement for httpd to be running before the network action service runs.

Is it possible to conditionally add a dependency on the database service if it's there? The database could be running elsewhere as well or it could be using sqlite.

HACKING.rst
```diff
@@ -54,8 +54,41 @@ operations, set ``dry_run = True`` in the ``[devel]`` section. For supressing
...
+On Centos systemd is the default way of managing processes.
+On Ubuntu you will have to install systemd. 
+
+Sample scripts for managing haas server and its network server is provided under 'scripts' folder.
+dev_haas.service manages the haas_server for your development enviornment
+dev_haas_network.service manages the network_server for haas for your development enviornment
+
+Instructions on how to modify the file to suit your enviornment are included in the file.
+After adjusting the parameters place both the file in
+/usr/lib/systemd/system (using sudo)
```
henn: Could we document ubuntu as well (and differentiate them)? I think it uses either /lib/systemd/system or /etc/systemd/

SahilTikale:
Made changes as per your comments @henn.
Let me know if anything else is needed.

scripts/haas_network.service
zenhack: > Is it possible to conditionally add a dependency on the database service if it's there? The database could be running elsewhere as well or it could be using sqlite.

I'm pretty sure the `After=` option does what you want. You'll have to explicitly enable the database server, but things should come up in the right order.

https://www.freedesktop.org/software/systemd/man/systemd.unit.html

INSTALL.rst
```diff
@@ -304,13 +304,52 @@ You should also set apache to start on boot::
...
+A systemd script for running the network server is available in the 'scripts' directory.
+Name of the script is: haas_network.service
+
+Centos:
+------
+
+Centos uses systemd to controll all its processes. 
+
+Place the file haas_network.service under:
+``/usr/lib/systemd/system/``
```
zenhack: This should probably be `/etc/systemd/system`; `/usr/lib` is for stuff installed via proper distro packages (rpms in the case of CentOS).

scripts/haas_network.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=HaaS Network Server 
+After=network.target
+
+[Service]
+User=haas_user
+Group=haas_user
+WorkingDirectory=/var/lib/haas/
+ExecStart=/usr/bin/haas serve_networks
```
zenhack: So everything after this line has sane defaults afaik, can you explain why you've explicitly filled in things like `KillMode` and `ExecReload`? `Type=simple` probably does make sense even though it is the default, just for clarity.

scripts/haas_network.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=HaaS Network Server 
+After=network.target
```
zenhack: Maybe add `After=posgresql` as well. I *think* as long as you leave of the requires, it won't actually cause postgres to be started, but if postgres is set to start, it will make things come up in the right order. Probably deserves a comment explaining why it's there, and we'll want to add any additional supported DBMSes in the future.

INSTALL.rst
```diff
@@ -304,13 +304,52 @@ You should also set apache to start on boot::
...
+
+Ubuntu:
+-------
+LTS version of Ubuntu, Ubuntu 14.04 does not come with systemd pre-installed.
+It uses "Upstart" an equivalent of systemd to manage its daemons/processes.
+
+Systemd is available from Ubuntu 15.04 onwards and LTS version 16.04 will ship with systemd by default.
+
+If you are using Ubuntu with any version prior to 15.04, you will have to install systemd before your can use the scripts provided here.
```
zenhack: We should *not* be advising people to install systemd on older versions of Ubuntu. The main utility of these scripts is integration into the distro's existing init system --- they just should not be used on non-systemd distros.

zenhack:
I'm seriously dubious of the idea that folks should be using systemd as part of a development environment. What's the dev scenario under which it's worth all that setup work as opposed to just having a pair of terminals?

INSTALL.rst
SahilTikale: As far as I understand, you put your .service files in ``/usr/lib/systemd/system`` and when you run ``sytemctl enable`` it creates a symlink under ``/etc/systemd/system`` 

For Ubuntu the place is ``/lib/systemd/system``

scripts/haas_network.service
SahilTikale: ``KillMode=process`` will kill only the main process. 
Since we are running only one process I thought to state it explicitly on how to kill it.

In general I want to be as much explicit as possible and not depend on defaults. 
That gives more control and consistency. 
Same reason for using ``Type=simple`` 

ExecReload was set to replicate the same behavior as what we would do manually to restart a process. But in retrospect (and did some reading) I guess it is a not a good idea to use  ``kill -HUP`` command as it is asynchronous.

scripts/haas_network.service
SahilTikale: Originally I had also added the dependency for Http server such that if someone tries to start this service it makes sure that http server is running before this service is invoked.

But @henn argued that we may want to run the web server and the network server on different machines in future hence such dependency is not required. 

On the same line of reasoning, do you think we will ever run the haas_network and the database on separate machines. If the answer is no, I will add database dependency to the service. 

Since we support only postgres for production, right now will put only that. 
With a comment in that it can support more databases in future. 

The dev service can support both postgres and sqlite.

INSTALL.rst
SahilTikale: But I thought it would be fair if I provide that option with a fair warning. 
Also, from the next LTS version systemd is going to be the default for managing daemons/processes. 

To address your concern I can append the old instruction for them. Thus giving the user the choice. Let me know what you think.

SahilTikale:
Elaboration about why we need systemd scripts for dev enviornment. 

From my personal experience in contributing to HaaS, I find it waste of screen estate when two terminals are occupied for showing debug messages. 

If I run the processes in background it defeats the purpose of having the debug messages on.
I can redirect the output to a file but it has so happened that couple of time the background process died and it took me sometime to realize that. 

Having systemd scripts eliminates those issues. 

Also, I prefer the development environment to be as close to the production environment. Idea is to reduce the time to adjust the code for production. 

For example, the mock switch driver does not have username, password fields and hence the corresponding switch register is fairly simple api call.
But when I tried testing an api call for the dell switch, I had to understand how it was different from the mockswitch api call. I believe this small differences add up and make it difficult for a newbie to contribute to haas. 

Another reason for having systemd scripts is that it reduces the startup/cleanup time. 
When you start your computer the processes are ready to receive instructions and they get shutdown in clean fashion when the computer is shutdown. 

In a scenario where two users want to setup a dev environment on a same system (server) this systemd scripts are already available to them. The only thing they have to mutually resolve is the port number to be used. 

Lastly I think it was just cool, to have a systemd script that can run as non-root user, in the working directory you wanted using the customized dev, enviornment you needed. :)

scripts/haas_network.service
zenhack: So again, `After=` doesn't actually introduce a hard dependency on postgresql; Unless you do `systemctl enable postgresql` at some point, it won't have any effect. It only has an effect if postgres is otherwise set to start on that machine, in which case it makes sure the haas network daemon starts *after* postgres. If we're running the two on separate machines, then this won't do anything, since postgresql will be disabled on the machine running the haas network daemon.

I'm actually not 100% sure this matters; it may be the case that you can start things out of order and be fine. I would test a few things. First:

1. Start the haas network daemon, configured to use postgres, with postgres running.
2. Do an api call that puts a network action in the queue. This will *probably* result in some kind of error from the network daemon.
3. Start postgres.

You should also try the above without step (2).

It may be that things will just start working once postgres comes up, in which case it may be fine for them to start out of order.

scripts/haas_network.service
zenhack: So, I don't have super strong feelings about this, but I'd say if you're explicitly setting something you should at least put a comment as to why. `Type=simple` is probably obvious enough to be an exception to this.

INSTALL.rst
zenhack: On my system:

```
# systemctl enable guix-daemon
Created symlink from /etc/systemd/system/mutli-user.target.wants/guix-daemon.service to /etc/systemd/system/guix-daemon.service.
# systemctl disable guix-daemon
Removed symlink /etc/systemd/system/mutli-user.target.wants/guix-daemon.service.
```

Where guix-daemon is a custom `.service` file I've added to `/etc/systemd/system`. Note that the symlink is in the subdirectory `mutli-user.target.wants`. If the service file is provided via an RPM (or deb, in the case of ubuntu), it will go in `/usr/lib/systemd/system` (or perhaps `/lib` on ubuntu), but usually the convention is that scripts not supplied by a package go in `/etc`. I don't know what changes ubuntu has made to the standard convention.

zenhack:
My objection to using systemd to solve the window management problem is that it's complete and total overkill. Alternatives:

* Use tabs in your terminal app
* GNU screen
* Minimize the windows

If you really need something to monitor the two processes, I wouldn't object to adding a sub-command to the haas cli called `dev_servers` or something, that did roughly this:

```python
import os
import signal

api_pid = os.spawnlp(os.P_NOWAIT, 'haas', 'haas', 'serve', '5000')
net_pid = os.spawnlp(os.P_NOWAIT, 'haas', 'haas', 'serve_networks')

try:
    (pid, status) = os.wait()
    if pid == api_pid:
        print("API daemon exited with status %d\n", status)
        print("Terminating network daemon.\n")
        os.kill(net_pid, signal.SIGTERM)
    else:
        print("Network daemon exited with status %d\n", status)
        print("Terminating API daemon.\n")
        os.kill(api_pid, signal.SIGTERM)
except KeyboardInterrupt:
    print("Keyboard interrupt recieved; terminating daemons.\n")
    os.kill(net_pid, signal.SIGTERM)
    os.kill(api_pid, signal.SIGTERM)
```

To be clear, I actually think having a command like this to address the problem you described is probably a good idea,  but introducing something as heavyweight as systemd for such a simple problem strikes a *very* poor complexity to benefit trade off.

INSTALL.rst
zenhack: For 16.04 folks should definitely be using the systemd service files. I think it might be good to have a manual setup section for reference for folks needing to set up haas on a non-systemd system, but we never actually tested the old thing on anything but CentOS/RHEL, and we should make that clear if we're going to keep them. Ultimately though, we can't document ubuntu setup without actually *doing* it. We could also add a bit suggesting writing your own upstart script if you've got the resources (and considering sending it to us if so).

SahilTikale:
While working on creating a systemd script I wanted a way to test my work in the dev environment. As a result I refined my scripts and committed them here thinking that people would benefit from it.
Also I am still in favor of having the dev environment mimic the production environment as closely as possible. To that end having systemd dev-scripts helps. 

For the overkill part, I have a simpler solution which I have used on regular basis while developing for HaaS. I have added that script in the latest commit to this PR. 

That way we don't have to decide on the best approach rather let developers choose how they wish to stand up their dev-environment.

INSTALL.rst
SahilTikale: Generally systemd files can be placed at any of the two locations
1 . ``/usr/lib/systemd/system`` (Centos) or ``/lib/systemd/system/`` (Ubuntu)
2. ``/etc/systemd/system`` 

When ``systemctl enable <servicename>`` command is issued no matter where the file is placed among the above mentioned locations, it will create a symlink under 
``/etc/systemd/system/<some folder>``

Exact value of ``<some folder>`` is decided by what is the value of the directive
``WantedBy=`` in the service file. 

After reading up on various sources and 
https://wiki.archlinux.org/index.php/systemd#Writing_unit_files

I gathered that the general opinion is that location 1. is used by pacakges and location 2. is used by users to put their custom .service files.

I decided upon location 1. because HaaS is a package and will be made available as an
RPM or pip install in future.

scripts/haas_network.service
SahilTikale: So I will remove ``ExecReload`` directive. 
What should I do for ``KillMode=process`` ? 
Let me know if you have a better suggestion. 
Or else I will just add an comment explaining what it is doing.

scripts/haas_network.service
SahilTikale: I tried testing dependencies using httpd.
rather than ``after=`` I used ``requires=`` 
what it does is that if you start haas_network.service and it sees that it contains a directive called
``requries=httpd.service`` 
it will check whether the service is running or not. If not it starts the service before it starts this service.

My thinking was that we should have dependencies on both httpd, and database for now since all three components are running on the same server and change it when we actually come to the point of running them on different machines. 

The reason being simply that dependencies makes for a self-check style of running services which in my opinion is more robust and less maintenance prone. 

But I am open to suggestions. 
@henn @gsilvis any thoughts, besides just two of us discussing here. :)

INSTALL.rst
SahilTikale: I liked the suggestion of 'writing your own upstart script' and sending it to us.

Please see my latest commit for the dev-script. If you like that we can incorporate something like that for the non-systemd environment.
I did test the systemd setup on the ubuntu machine, but that was 15.04. I did not install systemd on a LTS version.

zenhack:
> That way we don't have to decide on the best approach rather let developers choose how they wish to stand up their dev-environment.

This aludes to a broader issue that I want to bring up: If you have a handy-dandy one-off script that might be useful to some folks, you shouldn't need to get it into the offical, blessed, endorsed, someday-this-will-be-something-you-can-bet-your-company-on repository just to share it with folks. We should have a place where people can put things maybe not fit for general consumption without having to run the code review gamut; this kind of thing is a drain on resources, both for the reviewers and submitters.

Besides needing to hold stuff that goes into the mainline to a high standard in the general case, I worry about cluttering the repo with lots and lots of little scripts and examples. If we're not careful it could get hard to find the really important stuff.

@henn, @gsilvis, what do you think of adding a top level directory `contrib/`, where developers can put this sort of handy-dandy thing that's maybe not endorsed by the whole team, and just mentioning its existence in HACKING. We can have per-developer subdirs, and e.g. @SahilTikale could have pretty much free-reign over `contrib/SahilTikale`. I think we'd still do the pr process, but maybe only ask for one signoff if it only touches a developer's contrib folder, and have a generally lower standard for this sort of thing. The comment I made about the tarball @SahilTikale commited should be about the most extensive changes we should ask for.

I think this would do a lot to help us share these things without adding too much extra work; we already aren't great at keeping up with review.

As for this specific PR: This should have been a small simple change: add the *production* service file only, and corresponding doc changes. That could have gone in quickly and easily, and it might be best to just do that.

The rest of it can go under `contrib/SahilTikale`, if others like the idea of adding that directory. I think it's interesting to have there, I just feel nervous about having it as a blessed, supported thing.

What do others think?

INSTALL.rst
henn: I think I agree with Sahil on this - on a random ubuntu 15.10 system, /etc/systemd just has a bunch of symlinks to /lib/systemd/...:

```
$ find /etc/systemd -type f
/etc/systemd/logind.conf
/etc/systemd/timesyncd.conf
/etc/systemd/user.conf
/etc/systemd/bootchart.conf
/etc/systemd/system.conf
/etc/systemd/journald.conf
/etc/systemd/resolved.conf
```

scripts/haas_network.service
henn: I'd prefer to get rid of the `KillMode` line, and let it [default to control-group](https://www.freedesktop.org/software/systemd/man/systemd.kill.html). That way, if there's an inadvertent background process, it'll get killed as well.

Also - is there a way to make haas dependent on the postgres service if it's present? We could alternatively mark this as after

scripts/haas_network.service
henn: I like the `After=` option, since it doesn't introduce a hard requirement that the DB is on the same machine

henn:
I agree we shouldn't include a .tar in the source tree.

RE: `contrib/`, I'd be fine with that (whether it's flat or in subdirectories) as long as there was a README having at least a one-liner describing what each file did. We could treat it like documentation changes (one reviewer).

I think the user-level systemd scripts should probably end up there.

zenhack:
I like having it subdivided by user just so it's clear who's responsible for what --- I can imagine this getting very cluttered, and I think having somewhat clear ownership would be good.

@SahilTikale, how about you move the extra bits to `contrib/SahilTikale`, and adjust such that only the basics (prod .service file and corresponding doc changes) are outside of that? Put a file at `contrib/README` describing the whole `contrib` thing, and another readme in your directory describing whats in there.

SahilTikale:
I agree with @zenhack 
> Besides needing to hold stuff that goes into the mainline to a high standard in the general case, I worry about cluttering the repo with lots and lots of little scripts and examples. If we're not careful it could get hard to find the really important stuff.

But, I do not like the idea of having a `contrib/<developer>` folder. HaaS being an open source project it wont take longtime for  `contrib` to become larger than HaaS. That is just bloatware that not everyone is going to use. 

I don't have a solution right this moment, but like the idea suggested by @knikolla that
we should have our own repositories from where interested people can download scripts to customize HaaS as they wish at their own risk. No more code reviews or stuck PRs. 

These should address what @zenhack expressed:
>We should have a place where people can put things maybe not fit for general consumption without having to run the code review gamut; this kind of thing is a drain on resources, both for the reviewers and submitters.

So , for now I will remove all the extra work I submitted.
It will be only one script for running network daemon using `systemd` and related info in Install.rst.
Will remove everything else and reset the Hacking.rst. 

This PR has taken far more time then required.
I hope that this will get it through.

zenhack:
Quoting Sahil Tikale (2016-02-28 16:31:32)
>    I don't have a solution right this moment, but like the idea
>    suggested
>    by @knikolla that
>    we should have our own repositories from where interested people can
>    download scripts to customize HaaS as they wish at their own risk. No
>    more code reviews or stuck PRs.

One thing we could do is just have developers put stuff wherever they
like, and have a page on the wiki that indexes it; people can just put
links.

henn:
> One thing we could do is just have developers put stuff wherever they
> like, and have a page on the wiki that indexes it; people can just put
> links.


That sounds good to me. We can point to the developer wiki page from the README.

henn:
@SahilTikale - FYI, still waiting on the two changes I mentioned above regarding `KillMode` and `After=` before I can +1.

Thanks

SahilTikale:
@henn  
Done.

zenhack:
Things I've already mentioned that haven't been addressed:

* The diff still includes a tarball
* The instructions still suggest the idea of installing systemd on non-systemd platforms. I still fell pretty strongly that we shouldn't encourage that. We should just refer them to the "without systemd" instructions.

Also, the `dev_*.service` files are still included; I really don't think they belong here (especially since we've removed the corresponding docs). Could you remove those? Feel free to post them elsewhere and link to them from the wiki of course.

SahilTikale:
@zenhack 
I had deleted the files from my local branch, and then committed the changes. 
Apparently git just ignores the deleted files and locations. To delete the files from git I did not know that `git rm` is required. 
Done that. Check on my github branch the files do not show up. 

Also I rephrased the instructions in INSTALL.rst.
Removed the Warning and the recommendation to install systemd on UBUNTU systems LTS 14.04 and prior.

Just made two sections. One for systems running systemd by default and one for systems without it.

scripts/haas_network.service
henn: Assuming you **can** have multiple `after=` lines (does anyone know this?), could you add `After=postgresql` in addition to `after=network.target` so that we are depending on both the network and postgres being started?

PR #526, opened 2016-02-07
Ipmi driver

SahilTikale:
Following pull request follows the drivertization approach adopted for HaaS design and implements IPMI support in the form of a pluggable driver.

An older pull request #484 was a similar patch but in context of older haas code. A lot has changed since then, hence I made this new pull request. Once this is merged upstream #484 should be closed.

This patch consists of

1. Changes to model.py to accomodate a superclass Obm 
2. A new folder obm and the driver code is under this folder named ipmi.py
3. Changes to api.py as this patch changes how call node_register works. All other calls are not affected
4. Changes to cli,py for node_register. 
5. Fixed relevant changes for unit tests for api,py. 
        Previously api.py reported 59 failed tests. 
        58 of which gets fixed with this patch.


SahilTikale:
@henn @gsilvis 
check the logs this patch https://travis-ci.org/CCI-MOC/haas/builds/107641271
It will show 7 tests failed and rest passed.
Here is the breakdown of these tests.
1 test failed in api.py which is related to the switch driver. Unrelated to this patch.
6 tests failed in model.py which is not related to this patch. It is getting tested by travis since I tried to modify some code which I undid but still committed the file.

SahilTikale:
fixed the tests, I hope this should help in expediting the merge process. 
@henn  @gsilvis

haas/api.py
```diff
@@ -143,8 +143,10 @@ def project_detach_node(project, node):
     for nic in node.nics:
         if nic.current_action is not None:
             raise BlockedError("Node has pending network actions")
-    node.stop_console()
-    node.delete_console()
+#    node.stop_console()
+#    node.delete_console()
```
zenhack: Don't leave commented out code there when you commit, just delete it.

haas/api.py
```diff
@@ -183,29 +185,42 @@ def project_remove_user(project, user):
...
     _assert_absent(model.Node, node)
-    node = model.Node(node, ipmi_host, ipmi_user, ipmi_pass)
-    local.db.add(node)
+    obm_type = kwargs['obm']['type']
+    cls = concrete_class_for(model.Obm, obm_type)
+    if cls is None:
+        raise BadArgumentError('%r is not a valid OBM type.' % obm_type)
+    cls.validate(kwargs['obm'])
+    node_obj = model.Node(label=node, obm=cls(**kwargs['obm']))
+#    node = model.Node(node, ipmi_host, ipmi_user, ipmi_pass)
```
zenhack: Same thing; delete the commented out code.

haas/api.py
```diff
@@ -183,29 +185,42 @@ def project_remove_user(project, user):
...
+    local.db.add(node_obj)
     local.db.commit()
 
 
 @rest_call('POST', '/node/<node>/power_cycle')
 def node_power_cycle(node):
     node = _must_find(model.Node, node)
-    node.power_cycle()
+#    node.power_cycle()
```
zenhack: ...and again

haas/api.py
```diff
@@ -183,29 +185,42 @@ def project_remove_user(project, user):
...
+    node.obm.power_cycle()
 
 
 @rest_call('POST', '/node/<node>/power_off')
 def node_power_off(node):
-    db = model.Session()
-    node = _must_find(db, model.Node, node)
-    node.power_off()
+#    db = model.Session()
+#    node = _must_find(db, model.Node, node)
```
zenhack: Okay, not going to point out all of these; please go through and clean these up.

haas/cli.py
```diff
@@ -236,12 +236,31 @@ def headnode_stop(headnode):
...
+    obm_api = "http://schema.massopencloud.org/haas/v0/obm/"
+
+    if subtype=="ipmi":
+	if len(args) == 3:
+	    obminfo = {"type": obm_api+subtype, "host": args[0],
+	    		"user": args[1], "password": args[2]
+	    	      }
+	else:
+	    print "Error: subtype <ipmi> requires exactly 3 arguments"
+	    print "<hostname> <ipmi-username> <ipmi-password>"
```
zenhack: If you're going to write out custom error messages, there are a couple things to keep in mind:

1. The other error messages get printed to stderr, so these should do the same.
2. It might make more sense to raise a `TypeError`, which will trigger the usual error printing mechanism (and print the docstring).

The only disadvantage of (2) is that it won't currently actually print out your custom error, but if your docstring is clear enough this might be okay. It also wouldn't be too hard to change, but given that we want to swap the whole thing out for something that uses the `argparse` library, it may not be worth bothering.

I'm fine not raising an exception and printing the error ourselves if that's what you want to do, but if you do that you should use `os.stderr.write` instead of `print`.

haas/ext/obm/ipmi.py
```diff
@@ -0,0 +1,136 @@
+# Copyright 2013-2014 Massachusetts Open Cloud Contributors
```
zenhack: Probably want to bump the copyright date.

haas/ext/obm/ipmi.py
```diff
@@ -0,0 +1,136 @@
...
+            'type': Ipmi.api_name,
+            'host': basestring,
+            'user': basestring,
+            'password': basestring,
+            }).validate(kwargs)
+
+    def _ipmitool(self, args):
+        """Invoke ipmitool with the right host/pass etc. for this code.
+
+        `args`- A list of any additional arguments to pass the ipmitool.
```
zenhack: grammar: you don't need the word `the` here.

haas/ext/obm/ipmi.py
```diff
@@ -0,0 +1,136 @@
...
+    def validate(kwargs):
+        schema.Schema({
+            'type': Ipmi.api_name,
+            'host': basestring,
+            'user': basestring,
+            'password': basestring,
+            }).validate(kwargs)
+
+    def _ipmitool(self, args):
+        """Invoke ipmitool with the right host/pass etc. for this code.
```
zenhack: s/code/node/

haas/ext/obm/ipmi.py
```diff
@@ -0,0 +1,136 @@
...
+            '-P', self.password,
+            '-I', 'lanplus',
+            'sol', 'deactivate'],
+            stdin=PIPE,
+            stdout=PIPE,
+            stderr=PIPE)
+        proc.wait()
+
+    def delete_console(self):
+        return
```
zenhack: um?

haas/model.py
```diff
@@ -122,115 +122,12 @@ class Node(Model):
     project_id    = Column(ForeignKey('project.id'))
     project       = relationship("Project",backref=backref('nodes'))
 
-    # ipmi connection information:
-    ipmi_host = Column(String, nullable=False)
-    ipmi_user = Column(String, nullable=False)
-    ipmi_pass = Column(String, nullable=False)
+    # The Obm info is fetched from the obm class and its respective subclass
+    # pertaining to the node 
+    obm_id = Column(Integer, ForeignKey('obm.id'))
```
zenhack: This should probably have `nullable=False`

haas/model.py
```diff
@@ -122,115 +122,12 @@ class Node(Model):
...
-
-    def get_console(self):
-        if not os.path.isfile(self.get_console_log_filename()):
-            return None
-        with open(self.get_console_log_filename(), 'r') as log:
-            return "".join(i for i in log.read() if ord(i)<128)
-
-    def get_console_log_filename(self):
-        return '/var/run/haas_console_logs/%s.log' % self.ipmi_host
+    # The node is initially registered with no nics; see the Nic class.
```
zenhack: This seems like an odd place to put this -- a random comment at the end of a class definition? Two points:

1. I'm not sure what this has to do with this change.
2. If you think this needs clarification, there's probably a more sensible place to put it, maybe the api docs for `node_register`?

Despite (1) I don't really mind you slipping more docs in if you want, but put them in a more reasonable place.

haas/model.py
```diff
@@ -371,6 +268,49 @@ def get_port_networks(self, ports):
...
+        assert False, "Subclasses MUST override the stop_console method"
+
+    def delete_console(self):
+        assert False, "Subclasses MUST override the delete_console method"
+
+    def get_console(self):
+        assert False, "Subclasses MUST override the get_console method"
+
+    def get_console_log_filename(self):
+        assert False, "Subclasses MUST override this method"
```
zenhack: Probably want to put the method name here, just for consistency.

tests/unit/model.py
```diff
@@ -76,15 +77,17 @@ def sample_obj(self):
 class TestNic(ModelTest):
 
     def sample_obj(self):
-        return Nic(Node('node-99', 'ipmihost', 'root', 'tapeworm'),
-                   'ipmi', '00:11:22:33:44:55')
+	return Nic(Node(label='node-99', obm=Ipmi(type="http://schema.massopencloud.org/haas/v0/obm/ipmi",
```
zenhack: You seem to have bumped the indentation out a space; keep it in multiples of 4.

zenhack:
This looks mostly reasonable. I made some inline comments, Beyond that, I would add some tests to make sure the driver selection is doing the right thing. Look at the switch related tests for inspiration. You'll probably have to write null and or test obm drivers for that purpose.

Also, this is going to cause some merge conflicts with #491, though not horribly severe. Given how long that's been sitting for review I'd really appreciate it if you guys could merge that first and deal with the conflicts yourselves; it's somewhat unfair to make me deal with bitrot that occurs because you're doing other things instead of merging that.

//cc @henn.

haas/ext/obm/ipmi.py
SahilTikale: done.

haas/ext/obm/ipmi.py
SahilTikale: I did not understand. What does that mean ?

haas/ext/obm/ipmi.py
SahilTikale: What was the question ?

haas/model.py
SahilTikale: Done

haas/model.py
SahilTikale: It removed it and put it under api.py docstring.

haas/model.py
SahilTikale: Done.

SahilTikale:
Added a null obm driver by the name of mock.py

Added tests under class_resolver.py to confirm that the correct drivers are getting loaded.

zenhack:
Re the tests: what you've written is more just exercising class_resolver itself, which I think is pretty well covered and you didn't really make any significant changes to it anyway. What I had in mind was something more like:

1. Make a call to `node_register`, to create a node with a given obm type (maybe more than one, with different types).
2. Look in the database, and verify that the created nodes' Obms have the right type.

I realize now that we actually don't have such a thing for the switches, which makes my prior advice not very useful; whoops.

Also, I don't see the actualy file containing the mock obm driver; did you forget to `git add`? My guess is this is why the tests are failing.

haas/cli.py
SahilTikale: Will sort this out soon.

SahilTikale:
removed the unwanted imports from mock.py.
Added a test to make sure that node_registers inputs information into correct obm tables.

haas/model.py
zenhack: Not according to the diff. I checked out your branch locally to be sure; this is unchanged.

zenhack:
@SahilTikale, just noticed, you haven't updated `docs/rest_api.md`. Please do so, this is really important. See also #536

zenhack:
Most of this looks okay now. Update the api docs for this and I'll approve.

SahilTikale:
@zenhack 
I had originally used `None` because of which the output used to be `"project": "null"`
I and @lokI8 agreed that since it is going to be read by users it should be in a more readable format.
We felt `"project": "None"` was a better output than `"project": "null"`

SahilTikale:
@zenhack  
Did you have a look at the tests I included. I tried different things (dictionary, json) but comparing the expected and actual output was difficult. Hence took the simpler approach of testing one field from each table. If you are fine with it I can include a similar test (new PR) for the switch register call. 

Let me know.

haas/model.py
SahilTikale: which method name you want me to include.
I assume it was the power_off method and I included it.
It is there (lines 295 - 299 in model.py). 
Is there anything else I should be including ?

SahilTikale:
updated rest_api.md

haas/model.py
zenhack: Quoting Sahil Tikale (2016-02-25 13:00:19)

> > +    def get_console(self):
> > +        assert False, "Subclasses MUST override the get_console method"
> > +
> > +    def get_console_log_filename(self):
> > +        assert False, "Subclasses MUST override this method"
>
>    which method name you want me to include.
>    I assume it was the power_off method and I included it.
>    It is there (lines 295 - 299 in model.py).
>    Is there anything else I should be including ?

Compare the two methods I quoted: the first (and most all of the others)
say "Subclasses must override the <foo> method".
`get_console_log_filename` however says "Subclasses MUST override this
method." These should be made consistent.

zenhack:
I did, the tests looked okay.

Quoting Sahil Tikale (2016-02-25 12:55:28)

haas/model.py
zenhack: Quoting Ian Denhardt (2016-02-25 15:32:47)

> Quoting Sahil Tikale (2016-02-25 13:00:19)
>
> > > +    def get_console(self):
> > > +        assert False, "Subclasses MUST override the get_console method"
> > > +
> > > +    def get_console_log_filename(self):
> > > +        assert False, "Subclasses MUST override this method"
> >
> >    which method name you want me to include.
> >    I assume it was the power_off method and I included it.
> >    It is there (lines 295 - 299 in model.py).
> >    Is there anything else I should be including ?
>
> Compare the two methods I quoted: the first (and most all of the others)
> say "Subclasses must override the <foo> method".
> `get_console_log_filename` however says "Subclasses MUST override this
> method." These should be made consistent.

I see you figured this out and fixed it before I replied (b12772a).
Thanks. Annoyingly, I get email notifications about comments but not new
commits.

zenhack:
Quoting Sahil Tikale (2016-02-25 12:52:59)

>    @zenhack
>    I had originally used None because of which the output used to be
>    "project": "null"
>    I and @lokI8 agreed that since it is going to be read by users it
>    should be in a more readable format.
>    We felt "project": "None" was a better output than "project":
>    "null"

I disagree -- the point of having it in JSON is that it should be
*machine* readable. If we just wanted it to look nice for people, we
would do something like:

    Node cisco-35 is not assigned to a project. It has two Nics:
        * eth0 (with a mac address of "de:ad:be:ef:20:16")
        * eth1 (with a mac address of "de:ad:be:ef:20:17")

Obviously there's a balance to be struck, but null is the canonical
"absence of value" marker. Note that it *is* legal to create a project
called "None".

I think `null` will make sense tp most people who are going to have the
sophistication to be looking at the raw JSON output, and it's nicer for the
machine. If we want to have a nicer UI for less sophisticated users, it can handle
the null specially (and maybe present something like the above).

zenhack:
@SahilTikale, I made a couple minor inline comments. In addition, I'd like to see documentation on the individual obm drivers. Have a look at the `switch_register` docs to see what kind of thing I'm after.

SahilTikale:
I agree with your point that 
> the point of having it in JSON is that it should be machine readable. 

Once this important pull requests are done, I will make a miscellaneous pull request and change the `"None"` to keyword `None`, write a test for switch register api call and any other small task that may be too small for having its own pull request.

SahilTikale:
Removed the trailing comma from an example of a json object in rest_api.md
Instructions in node_register made consistent in style with that of switch_register in rest_api.md
Added obm_drivers.md to docs.

zenhack:
I'm pretty happy with this, +1. I'll reiterate my request that we wait on #491 before merging that, because I feel it's somewhat unfair to ask me to fix merge conflicts given how long that was sitting there. I'll put the do not merge yet label on this one for now, but if folks disagree with that we can discuss/remove the label.

zenhack:
Okay, there's actually one big thing we have to figure out before we merge this: how are you guys going to migrate the DB? I'm going to remove my +1 for the time being; we need a plan there. We don't have a proper framework in place, and if we merge it means you guys can't upgrade your existing system.

knikolla:
+1. I agree on the DB migration issue, but this pull request is blocking a new HaaS deployment.

henn:
@zenhack @SahilTikale - as long as Kyle's show_nodes is merged before this, I can sync production to that and leave this one out till we have time to migrate the db.

zenhack:
To be totally explicit on my position: I think the pr is fine as is, but we need to hold off on merging until we know how we're going to merge the DB, as merging will block deploying other changes otherwise.

SahilTikale:
I realized that node_delete affected only the table `node` and left the corresponding OBM info in orphan state. My latest commit fixes that.

henn:
2 +1's. Merging

PR #528, opened 2016-02-11
Cli switch register

SahilTikale:
Adding cli call for the api call Switch Register. 
I modified the mockswitch driver to include the hostname, username and password.

Most switches take at least these parameters. 

It helped me test the cli.

PR #530, opened 2016-02-17
Adding CLI call for switch_register

SahilTikale:
Updates same as #528.
#528 was mistakenly branched from a non-master branch. Thus it contained lot of stuff that should not be there.

Will delete that pull request. Replacing #528 with this pull request.

The following commit:
	1) modifies the mock switch driver to resemble a real switch driver
	2) Adjusted the unit tests to accomodate extra arguments from switch driver
 	3) Added cli call corresponding to api call for switch register
	4) Updated docs/USING.rst with an example on how to use it via API and cli calls.

### DIFF
--- a/docs/USING.rst
+++ b/docs/USING.rst
@@ -55,7 +55,7 @@ been run succesfully.
 Usage Examples
 =================
 
-Included herewith are some examples about using some newly added cli and api calls. 
+Included herewith are some examples about using cli and api calls. 
 
 
 
--- a/docs/USING.rst
+++ b/docs/USING.rst
@@ -63,17 +63,17 @@ Included herewith are some examples about using cli and api calls.
 -------------------------------
 
 Eg> Switch name: mockswitch01
-     hostname: switchhost01
-     username: switchuser01
-     Password: password1234
+     Host name:  switchhost01
+     User name:  switchuser01
+     Password:   password1234
 
 api call
 
 ::
 
-    curl -X put http://127.0.0.1:5000/switch/mockswitch_fromapi -d '
+    curl -X put http://127.0.0.1:5000/switch/mockswitch01 -d '
         {"type": "http://schema.massopencloud.org/haas/v0/switches/mock",
-        "hostname": "switchHost01",
+        "hostname": "switchhost01",
         "username": "switchuser01",
         "password": "password1234"}'
 
@@ -81,7 +81,7 @@ cli call
 
 ::
 
-       haas switch_register mock03 mock switchHost01 switchuser01 password1234
+       haas switch_register mockswitch02 mock switchhost01 switchuser01 password1234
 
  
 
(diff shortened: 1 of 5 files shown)


docs/USING.rst
```diff
@@ -50,3 +50,38 @@ to hard-disk booting the installed system.
 This is, as the filepath states, merely an example of how you might deploy to
 physical nodes.  Existing deployment systems such as Canonical's MAAS have also
 been run succesfully.
+
+
+Usage Examples
+=================
+
+Included herewith are some examples about using some newly added cli and api calls.
```
knikolla: Not sure this line is needed. Newli added api calls will soon be old api calls.

You can just say api and cli calls for registering a switch with HaaS.

docs/USING.rst
```diff
@@ -50,3 +50,38 @@ to hard-disk booting the installed system.
...
+    curl -X put http://127.0.0.1:5000/switch/mockswitch_fromapi -d '
+        {"type": "http://schema.massopencloud.org/haas/v0/switches/mock",
+        "hostname": "switchHost01",
+        "username": "switchuser01",
+        "password": "password1234"}'
+
+cli call
+
+::
+       haas switch_register mock03 mock switchHost01 switchuser01 password1234
```
knikolla: This didn't convert into a code block if you view the whole file.

haas/ext/switches/mock.py
```diff
@@ -32,17 +33,30 @@ class MockSwitch(Switch):
...
-        Schema({}).validate(kwargs)
+        schema.Schema({
+            'username': basestring,
+            'hostname': basestring,
+            'password': basestring,
+        }).validate(kwargs)
+
+
+#    @staticmethod
+#    def validate(kwargs):
```
knikolla: I think we don't need to keep the old code commented out. It's in version control if we want to go back in time.

knikolla:
+1

haas/api.py
```diff
@@ -899,6 +899,26 @@ def _assert_absent(cls, name):
...
+
+def _assert_absent(cls, name):
+    """Raises a DuplicateError if the given object is already in the database.
+
+    This is useful for most of the *_create functions.
+
+    Arguments:
+
+    cls - the class of the object to query.
+    name - the name of the object in question.
```
lokI8: Looks like you duplicated about 1/2 this function somehow, if python calls the second instance of the function it should work? I'm not sure if it does or not. Either way its gross =) Please change

docs/USING.rst
```diff
@@ -50,3 +50,39 @@ to hard-disk booting the installed system.
...
+
+Included herewith are some examples about using cli and api calls. 
+
+
+
+1) Register a switch with HaaS:
+-------------------------------
+
+Eg> Switch name: mockswitch01
+     hostname: switchhost01
```
lokI8: Host is lowercase in your hostname here, but uppercase in the arguments for the call. Please be consistent one way or the other. I know its not big, but no reason to potentially confuse people that a hostname needs to be camel case.

lokI8:
Looks good to me now. +1, merging

PR #545, opened 2016-03-10
Network_ACL

kylehogan:
Networks and projects now have a many to many relationship allowing a network to be shared between multiple projects and a project to own multiple networks.

Still needs:

- [ ] testing for auth
- [ ] auth
- [x] API call for a project to grant other projects access to networks it owns

### DIFF
--- a/examples/haas.cfg.dev-no-hardware
+++ b/examples/haas.cfg.dev-no-hardware
@@ -23,7 +23,7 @@ dry_run=True
 [extensions]
 haas.ext.switches.mock =
 
-# haas.ext.network_allocators.null =
+haas.ext.network_allocators.null =
 #
 # Depending on what you're trying to do, you may want to use the vlan_pool
 # network allocator instead of the null allocator. To do this, comment out the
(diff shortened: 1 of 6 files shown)


gsilvis:
Going to put some comments from our conversation yesterday:

Prefer ``network_grant_access`` and ``network_revoke_access`` to the current ``project_add_network`` and ``project_remove_network``.

There are several auth subtleties that we need to A) decide B) document.  I'm going to put the questions, along with the answers I recall us generally agreeing on.

--Can the creator of a network remove other projects' nodes from it? [Yes]
--What information on other projects' attached nodes can the network creator see? [nic name, node name, channel; what's required for the remove call]
--What information on other projects' attached nodes can someone with access to a network see [none]
--If an administrator creates a network for a particular project, can that project delete it? [Uncertain---perhaps allow both methods?  Requires reworking 'create_network' a little bit.]
--When you revoke a project's access to a network, is it okay if they still have nodes on it? [I lean towards yes, but we definitely did not decide for sure.]

gsilvis:
Superseded by #554

Issue #549, opened 2016-03-15
Modify CLI calls related to 'port' operations


SahilTikale:
New API calls for port operations (register, detach, attach etc) requires name of the switch as argument. 
Corresponding CLI need to modify to accomodate those changes.

Also, need to write unit tests to test these changes.

kylehogan:
@SahilTikale closed now?

SahilTikale:
I merged it. We can close it.

PR #551, opened 2016-03-15
Fixes CLI calls for 'port related' operations.

SahilTikale:
Addresses issue #549 
Amended to incorporate re-factored API call that requires to
provide a <switchname> as an argument for every port related operation.


SahilTikale:
@knikolla @kylehogan 
This is a small PR but import for engage 1.

Please review it when you get time, so that we can merge it with the master to the earliest.

tests/unit/obmTest.py
```diff
@@ -0,0 +1,87 @@
...
+	api.node_register('compute-01', obm={
+                  "type": "http://schema.massopencloud.org/haas/v0/obm/ipmi",
+                  "host": "ipmihost",
+                  "user": "root",
+                  "password": "tapeworm"})
+
+	node_obj = db.query(model.Node).filter_by(label="compute-01")\
+			.join(model.Obm).join(haas.ext.obm.ipmi.Ipmi).first()
+
+
```
knikolla: Remove extra blank line.

SahilTikale: done.

kylehogan:
can the names of the functions defined in cli.py ('port_register' and 'port_delete') also be changed to match api.py ('switch_register_port' and 'switch_delete_port')?

other than that it looks good

SahilTikale:
> can the names of the functions defined in cli.py ('port_register' and 'port_delete') also be changed to match api.py ('switch_register_port' and 'switch_delete_port')?

I don't think so. Ports are for switches, therefore no point in stating the obvious. 

But I while we are at it, I would like to change the name of `port_delete`
to  `port_unregister`, for consistency sake.

@kylehogan @henn @zenhack 
What do you guys think ?

zenhack:
Re: consistency, nothing else is named `*_unregister`, so I don't see how that helps.

I think @kylehogan is right; we should rename to `switch_*_port`. This is consistent with the `node_*_nic` calls. The naming is a bit odd, and it basically comes down to trying to map a somewhat object-oriented notion onto a purely procedural interface; It's called `node_register_nic` because it's conceptually a method on the node. In the past I've toyed with the idea of doing some meta-programming goop on the server side to make this actually how it works (e.g. `/node/<node>/<stuff>` fetches the node object and then dispatches from there based on `<stuff>`), and while I came to the conclusion that it was a bad idea, it helps illustrate what's going on there conceptually.

I'll point out that the reason for the renaming is *not* that the server and client side should match; the server side is strictly an implementation detail (the public api is the HTTP request) so from the client's perspective it doesn't really matter what they're called. The functions don't even need to map one to one (and they actually don't; see `network_create_simple`). But the above explains the reasons the server-side name was changed, and the same reasoning applies to the client in this case.

tests/unit/obmTest.py
```diff
@@ -0,0 +1,87 @@
+"""Unit tests for api.py"""
```
zenhack: This test file has absolutely nothing to do with the content of this patch; it doesn't exercise the CLI at all. Explain?

SahilTikale: Happened by mistake ? Will remove that.

SahilTikale:
So what is the general consensus, change the API names are not ?

SahilTikale:
If we all do not have a consensus, we can just merge this the way it is.
Engage 1 deployment depends on this. 

We will worry about the change of API names in future.

zenhack:
My vote is for making the name change that @kylehogan suggested.

SahilTikale:
This PR is focused on adding cli calls for port operations. So let this do just that.
I will create another PR for changing API names at a later point.

knikolla:
+1

zenhack:
Sounds good to me. Sanity check, since we don't actually have unit tests for the cli: you did try this, correct? If so, you have my +1 as well (and can thus merge).

SahilTikale:
I tried these cli operations before commiting

PR #541, opened 2016-03-03
Flask (routing)

zenhack:
@henn, worth a first pass I think.

This builds on top of #491, so most of the commits are from that.

PR #556, opened 2016-03-18
Flask-SQLAlchemy and Database Migrations

zenhack:
This isn't quite ready yet, but it's worth having folks start looking at it. This is both moving over to the flask sqlalchemy extension and adding a database migration infrastructure. It includes a migration script that upgrades the content of my auth pr (#491) to the contents of the latest commit on this repository. Note that it does *not* handle upgrading the database for the OBM work that @SahilTikale did, since that was merged first.

It also includes some changes to the travis stuff; we run `tests/stress.py`, and we run the tests against both sqlite and postgres.

It does not *yet* include automated testing for the database migration infrastructure; I wanted to run my ideas about how to do that by you guys first:

1. Write a function like `initial_db` in `tests/unit/api/auth.py` (or just use that one), which populates a db with some useful objects.
2. Before making any changes to the schema, use said script to populate a postgres database, then dump the database as a sql script (via `pg_dump`). Commit the result.
3. Make your changes, and add a migration script.
4. Testing will consist of:
  1. run the sql script to populate a (previously empty) database.
  2. run the migration script, to upgrade to the latest schema.
  3. run a series of tests that expect said initial database.

I'm also still trying to pin down how the schema changes due to adding and removing extensions should work. Alembic has some support for branching and merging schema, but I need to do a bit more reading before I can decide what I think the right thing to do is there.

I also need to write some developer docs describing how to do database migrations.

@SahilTikale, @henn, let me know what you think.


zenhack:
(The travis failure is something I'm aware of and still trying to pin down).

zenhack:
I'm pretty sure I understand the nature of the test failure. basically, pytest is loading env.py at an inappropriate time (when collecting tests). env.py should never be executed outside of a specific context set up by alembic, and so it's freaking out. I need to figure out how to get pytest to not do that, but I don't think there's anything else wrong.

henn:
For some strange reason, this PR is including the merged changes from #541.

Closing this out and opening a new PR so we can discuss without the extra stuff.

PR #560, opened 2016-03-25
initial ipmi v1.5 support

kylehogan:
Just added an ipmi-obsolete to support v0.5 removed lanplus option in status and modified api_name


haas/ext/obm/ipmi-obsolete.py
```diff
@@ -0,0 +1,131 @@
+# Copyright 2015-2016 Massachusetts Open Cloud Contributors
```
SahilTikale: Please replace the hyphen with an underscore. Sometimes python misconstrues the hyphen to be minus sign.
It will also create problems while importing.

haas/ext/obm/ipmi-obsolete.py
```diff
@@ -0,0 +1,131 @@
...
+from subprocess import call, check_call, Popen, PIPE
+import os
+
+class Ipmi(Obm):
+    id = Column(Integer, ForeignKey('obm.id'), primary_key=True)
+    host = Column(String, nullable=False)
+    user = Column(String, nullable=False)
+    password = Column(String, nullable=False)
+
+    api_name = 'http://schema.massopencloud.org/haas/v0/obm/ipmi-obsolete'
```
SahilTikale: here also, replace hyphen with underscore, just for consistency.

haas/ext/obm/ipmi-obsolete.py
zenhack: It's a comment...

kylehogan: For the filename?

zenhack: Quoting kylehogan (2016-03-25 23:58:34)
>    For the filename?

Ah, I misunderstood.

kylehogan:
needs to be tested

haas/ext/obm/ipmi_obsolete.py
```diff
@@ -0,0 +1,131 @@
...
+import schema
+import subprocess
+
+from haas.model import Obm
+from haas.errors import OBMError
+from haas.dev_support import no_dry_run
+from subprocess import call, check_call, Popen, PIPE
+import os
+
+class Ipmi_obsolete(Obm):
```
henn: Would recommend Ipmi_1_5 to be more descriptive

haas/ext/obm/ipmi_obsolete.py
```diff
@@ -0,0 +1,131 @@
...
+    id = Column(Integer, ForeignKey('obm.id'), primary_key=True)
+    host = Column(String, nullable=False)
+    user = Column(String, nullable=False)
+    password = Column(String, nullable=False)
+
+    api_name = 'http://schema.massopencloud.org/haas/v0/obm/ipmi_obsolete'
```
henn: Ditto with changing the name

haas/ext/obm/ipmi_1_5.py
```diff
@@ -0,0 +1,129 @@
...
+    api_name = 'http://schema.massopencloud.org/haas/v0/obm/ipmi_1_5'
+
+    __mapper_args__ = {
+        'polymorphic_identity': api_name,
+        }
+
+    @staticmethod
+    def validate(kwargs):
+        schema.Schema({
+            'type': Ipmi.api_name,
```
SahilTikale: Change the class name for the value of `type`

SahilTikale:
I tried adding a node using your driver and then tried deleting it. the command hung forever and did not delete the node from the database. I had to manually exit the session (Control C) 
After repeating the step ( haas node_delete <node_name> then aborting manually using Control C)
multiple times, it deleted the node from the database.

I suggest you try adding dummy nodes using this driver and see if it deletes them gracefully.

haas/ext/obm/ipmi_1_5.py
```diff
@@ -0,0 +1,129 @@
...
+            }).validate(kwargs)
+
+    def _ipmitool(self, args):
+        """Invoke ipmitool with the right host/pass etc. for this node.
+
+        `args`- A list of any additional arguments to pass to ipmitool.
+        Returns the exit status of ipmitool.
+        """
+        status = call(['ipmitool',
+            '-U', self.user,
```
SahilTikale: Comment out the user part "-U ..." 
Sun servers at least does not need the user.

haas/ext/obm/ipmi_1_5.py
```diff
@@ -0,0 +1,129 @@
...
+# IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language
+# governing permissions and limitations under the License.
+
+"""IPMI driver for implementing out of band management. """
+
+from sqlalchemy import Column, String, Integer, ForeignKey
+import schema
+import subprocess
+
```
SahilTikale: "import logging" is missing here 
and also in ipmi file. 
Can you also add this in the ipmi.py also ?

SahilTikale:
I have made some suggestions after which the driver is good to go.

Also I was thinking if there is a better way to do this then writing a new driver that is marginally different than the first one. 

I was thinking having a flag for nodes with version or any special consideration and invoking those methods based on it. 

Like if node provides its ipmi version to be 1.5 we can invoke methods with "-I lan" flag
and if the server do not require the user, we can change the command accordingly by updating or deleting the requisite arguments.

Does it sound better or more complex ?
what do you guys think @kylehogan @zenhack @henn @gsilvis @knikolla 

I just wanted to make sure if we can do this better before this PR gets merged.

SahilTikale:
@kylehogan 
I tested your driver and after the in-line changes I suggested, it is good to go.

henn:
@kylehogan , @SahilTikale  - have either of you tested the sun blades with the current Ipmi driver after I changed the IPMI settings?

ipmitool seemed to be able to do some things when using `-I lanplus`

SahilTikale:
yes I have tested the feature in the Beta HaaS server. `-I lanplus`.
It works with both `-I lan` and `-I lanplus` only difference being that sun-blades do need `-U <username>`
hence my earlier comment as to whether we need this PR should we put some function in the current ipmi driver to not user `-U <username>` flag and may be the no-access to the console methods too.

kylehogan:
I would say that if the only difference is the `-U username` then it doesn't make sense to have an entire new driver. especially because if the lanplus flag is working it means that they actually are running 2.0 not 1.5

henn:
> yes I have tested the feature in the Beta HaaS server. -I lanplus.
> It works with both -I lan and -I lanplus only difference being that sun-blades do need -U <username>
> hence my earlier comment as to whether we need this PR should we put some function in the current ipmi driver to not user -U <username> flag and may be the no-access to the console methods too.
> 

Have you tested using haas to run the commands? I wonder if specifying the username as "" in node_register would work, as it does when running ipmitool directly.
If so, then we don't need to make any haas changes other than perhaps adding something to the deployment guide.

SahilTikale:
Yes I commented out the `-U <username>` option from the driver and ran the HaaS commands it booted the sun-blade nodes. We can still pass some arbitary strings while registering nodes and not use it when passing the ipmi commands. 

Because of this I feel we can have the same driver and just have some `if clause` to supress the `-U <username>` option in case of the sun-blades. As we find need to include more feature in the ipmi driver, we should include more functions that do selective things. Therefore I think, we should just have two functions, one that uses the `-U` flag and one that does not.

henn:
Did you try 'node_register sun-123 ipmi "" password' on an unmodified haas? (Apologies if the args are out of order- doing this from memory)

We may not need any haas modifications if we can do that.

zenhack:
@henn, that worked when we were last playing with the sun blades; the only thing that (we thought) broke support for those blades was the addition of `-I lanplus`; if that doesn't *actually* break them we don't need to do anything, just pass the empty string as the username.

henn:
@kylehogan - I think we can close this once we verify that everything works with the current driver.

I have a feeling sun serial consoles might require a little extra work to setup. I stuck a pointer to the manuals on the cluster page if you want to give it a shot (you can do everything remotely).

SahilTikale:
I tried registering a sun-blade on the beta server with the user string to be empty. 
It registering correctly with a vacant username field. 
All the power_cycle commands are working as intended. For that purpose I think ipmi driver will suffice.
Don't know what we would want to do for the console though.

zenhack:
Do those nodes support anything we can use for the console at all?

Last I looked the console support is pretty broken anyway, and I think we need to pin down how we're going to fix it even for ipmi 2.

SahilTikale:
Console support in HaaS is a dedicated work on its own. I think we should close this for now.
If all agree I will go ahead and close this PR.
@henn @zenhack @kylehogan

zenhack:
Sounds good to me

Quoting Sahil Tikale (2016-05-09 17:44:50)

>    Console support in HaaS is a dedicated work on its own. I think we
>    should close this for now.
>    If all agree I will go ahead and close this PR.
>    @henn @zenhack @kylehogan

kylehogan:
I concur

SahilTikale:
While working on a related project recently (MAAS) I realized that we might need a separate driver for ipmi version 1.5. so leaving this open for now.

matsuuran:
hey @SahilTikale just wanted to see at what point we should implement the new driver

shwsun:
I remember we agree to close this one for now. 
//cc @henn

matsuuran:
@henn sounds like this one should be closed when you get a chance

shwsun:
We should close it. @zenhack and @SahilTikale  agree with it along time ago.

zenhack:
I'm going to go ahead and click the close button.

PR #561, opened 2016-03-25
 Flask-SQLAlchemy and Database Migrations (try 2)

henn:
This is replacing #556, as for some reason github wasn't removing the merged commits and thus the diffs were bigger than they had to be.

Ian's (@zenhack) comments from #556:
> This isn't quite ready yet, but it's worth having folks start looking at it. This is both moving over to the flask sqlalchemy extension and adding a database migration infrastructure. It includes a migration script that upgrades the content of my auth pr (#491) to the contents of the latest commit on this repository. Note that it does not handle upgrading the database for the OBM work that @SahilTikale did, since that was merged first.
> 
> It also includes some changes to the travis stuff; we run tests/stress.py, and we run the tests against both sqlite and postgres.
> 
> It does not yet include automated testing for the database migration infrastructure; I wanted to run my ideas about how to do that by you guys first:
> 
>     Write a function like initial_db in tests/unit/api/auth.py (or just use that one), which populates a db with some useful objects.
>     Before making any changes to the schema, use said script to populate a postgres database, then dump the database as a sql script (via pg_dump). Commit the result.
>     Make your changes, and add a migration script.
>     Testing will consist of:
>         run the sql script to populate a (previously empty) database.
>         run the migration script, to upgrade to the latest schema.
>         run a series of tests that expect said initial database.
> 
> I'm also still trying to pin down how the schema changes due to adding and removing extensions should work. Alembic has some support for branching and merging schema, but I need to do a bit more reading before I can decide what I think the right thing to do is there.
> 
> I also need to write some developer docs describing how to do database migrations.
> 
> @SahilTikale, @henn, let me know what you think.
>


HACKING.rst
```diff
@@ -31,7 +31,7 @@ you may need to run::
 
   pip install importlib
 
-You may get an error 'psycopg2 package not found' when you do 'haas init_db' 
+You may get an error 'psycopg2 package not found' when you do 'haas-admin db create'
```
SahilTikale: What the motivation behind this change. Does it really create a database or populates an already created database ? (in case of postgresql)

.travis.yml
```diff
@@ -1,9 +1,19 @@
...
+env:
+  - DB=sqlite
+  - DB=postgres
+install:
+  - sh -c "if [ $DB = postgres ]; then pip install psycopg2; fi"
+  - python setup.py install
+before_script:
+  - sh -c "if [ $DB = postgres ]; then psql -c 'CREATE DATABASE haas_travis;' -U postgres; fi"
+  - sh -c "if [ $DB = postgres ]; then cp ci/testsuite.cfg.postgres testsuite.cfg; fi"
+script: "py.test tests/unit tests/stress.py"
```
henn: Could you split this into two commands, with `tests/unit` being run with with `-n auto`?

With the addition of `tests/stress.py` and the multiple DB types (postgres + sqlite), our CI builds are going from 1.5 mins to around 5. While I think running these are good changes, if we can take advantage of the [multiple cores](https://docs.travis-ci.com/user/migrating-from-legacy/) than folks can get responses a little faster.

docs/UPGRADING.rst
```diff
@@ -0,0 +1,17 @@
...
+==============
+
+This file describes the procedure for upgrading HaaS to a new version.
+
+1. Read the release notes, as they may contain version-specific information.
+2. Stop the HaaS services (webserver and HaaS network daemon).
+3. Download and install the new version of HaaS::
+
+     $ git clone https://github.com/cci-moc/haas
+     $ python.setup.py install
```
SahilTikale: there is a typo in the command.

Also does running `python setup.py install` overwrites the executable with newer version ?
I am guessing that there should be some command for uninstalling old code before running this command.

examples/haas.cfg
```diff
@@ -62,10 +62,20 @@ endpoint = http://127.0.0.1:5000
...
-# connect to. At present, only SQLite and PostgreSQL are supported.
+# connect to. At present, only SQLite and PostgreSQL are supported, and only
+# PostgreSQL is supported by our migration scripts.
+#
+# If a relative path is used with SQLite, it will be relative to the HaaS source
+# tree. Unless you're doing development, this is almost certainly not what you
+# want, so use an absolute path.
+#
+# Use the sqite database file ``/absolute/path/to/haas.db``:
+uri = sqlite:////absolute/path/to/haas.db
```
SahilTikale: In an example, the activated option should be the one that works. 
I think this should be commented out and replaced with an option that would work without requiring to make any changes by the user.

haas/ext/switches/migrations/dell/099b939261c1_rename_dell_switch_table_for_flask_.py
```diff
@@ -0,0 +1,27 @@
+"""Rename dell switch table for Flask-SQLAlchemy
```
SahilTikale: why is this file named as such. What does it do ?

haas/ext/switches/migrations/mock/b5b31d19257d_rename_mockswitch_table_for_flask_.py
```diff
@@ -0,0 +1,27 @@
+"""Rename mockswitch table for Flask-SQLAlchemy
```
SahilTikale: same question as before ?

SahilTikale:
I went through the PR. There are many new files introduced and it is hard to see how they relate to each other. @zenhack  Can you provide a simple work flow on how this migration would work so that I can understand the inter-relationships better. Will help me to provide quality feedback.

haas/ext/switches/migrations/dell/099b939261c1_rename_dell_switch_table_for_flask_.py
zenhack: @SahilTikale, it's the same thing as ` haas/migrations/versions/6a8c19565060_move_to_flask.py ` but for the extensions, as opposed to the tables in haas core. That file contains a better explanation. I'll add a cross reference at least.

examples/haas.cfg
zenhack: So I don't feel super strongly about that specific option, but some points:

1. That file's purpose is to serve as a thorough reference; we have another file in that directory that works as an out-of-the-box config for developers (mentioned at the top of this file), and in production there's really no way to provide something like that (besides, we encourage using postgres anyway).
2. My reasoning was that per the comments above, storing the database in the source tree isn't something you want unless you're developing, and most of the options in this file are more production-oriented. Besides, as I mentioned, we already have an off-the-shelf config.

If we are going to change it I'd recommend going for the postgres as the one that's exposed. That doesn't actually address the reason you're asking for it, but that's not really achieveable in a production setting anyway.

I don't feel super strongly; read over the above points and let me know what you think.

docs/UPGRADING.rst
zenhack: Good catch on the typo; I will fix.

There is an uninstall command of some variety, but from what I know it isn't really to be relied upon. This should at least overwrite things, so should be okay, but idally we *would* want to do a proper uninstall. For that though we should be doing rpm/deb packages, as pip/setup.py won't do a reasonable job of it. We should do that at some point.

.travis.yml
zenhack: To actually run these in parallel without interfering with eachother, they'd have to have their own postgres DB; I don't know that there's a way to do that without at least a little bit of hacking. With sqlite that isn't an issue since we're using an in-memory db (and it *should* work, though I haven't tested it), but sqlite is also a lot faster anyway.

HACKING.rst
zenhack: The `db` subcommand has a bunch of stuff in it provided by the migration framework (including upgrades); it seemed the place to add it. The `create` subcommand does what `haas init_db` did before. I'm actually not sure I think this is the right thing; most (but not all) of the other stuff under `haas-admin db` is developer-oriented, so it might make sense to have another entry point to the create/upgrade commands. I called it `create` because `init` already does something else, but that's part of what makes me think I should just move it to another command; `init` is really only interesting to devs.

The framework provides some of it's own CLI tooling, so I put this in a separate command since (1) we've been talking about splitting out the client/server stuff into two or more commands anyway, and (2) it meant not having to write as much glue code.

zenhack:
@SahilTikale, I've tried to answer your specific questions. I'll put together a more extensive developer doc, but as far as a guide to what's in the PR, there are two big things:

1. Moving over to the [Flask-SQLAlchemy][1] plugin. This doesn't really provide any single big piece of functionality, but it integrates the library we're already using more tightly with Flask, and lets us remove a bit of the glue/support code. For example, it makes a request-local session itself, so my hand-rolled session management stuff is gone. Per the comments I added to `tests/stress.py`, this makes the specific bug that was written for no longer something our own code is responsible for. It also auto-generates table names, which I'd written a small amount of custom stuff to do for us before (and which was part of our root level `Model` class). Per the comment in one of the migration scripts, the names it choose are not quite the same in all cases, so the schema needs to be tweaked to support the new names. Because of the fact that this is now covered by the library, I removed our own `Model` and `AnonModel` classes, and explicitly put the `id` and `label` fields in each table, since the two lines of code seemed not worth the abstraction without the other stuff that was in those superclasses. A *lot* of the line-by-line changes in this patch are just prepending `db.` to things, a fairly mechanical but pervasive change. Because of this, basically every file that declares a model has a lot of superficial changes in it. Similarly, basically every instance of `local.db` became `db.session`.
2. Integrating database migration support. We're using [Flask-Migrate][2], which integrates [Alembic][3] with Flask in the same way that `Flask-SQLAlchemy` integrates sqlalchemy itself. The developer docs I'm going to write will tell you to go read those docs for a general overview, and then just explain HaaS specific stuff, so worth looking at those up front anyway.

[1]: http://flask-sqlalchemy.pocoo.org/2.1/
[2]: https://flask-migrate.readthedocs.org/en/latest/
[3]: http://alembic.readthedocs.org/en/latest/

Those two big things happen in that order in the history; most of the migration related commits are in the latter half. There are a handful of tiny cleanups in their own commits too; I tried to make sure those were particularly well explained and separated since they were otherwise out of place, but stuff like adding `.cache` to `.gitignore` didn't seem worth its own pr. There aren't too many of those though.

zenhack:
797a057 adds some high-level docs, hopefully they're useful for understanding what's going on.

//cc @henn, @SahilTikale.

zenhack:
@gsilvis, would you have a look at this?

.travis.yml
henn: How difficult would it be to create/destroy a random postgres DB for every test (maybe placing a hook into fresh_database())? Perhaps we could add an option to testsuite.cfg.postgres to set `[database] random = True` or something.

On my workstation, `-n auto` cut the tests in half for sqlite, which would be valuable to a developer anxiously awaiting a green checkmark :-)

conftest.py
```diff
@@ -0,0 +1,10 @@
+"""pytest hooks"""
```
henn: Could we put this in a subdirectory and not in the root?

docs/UPGRADING.rst
```diff
@@ -0,0 +1,20 @@
...
+==============
+
+This file describes the procedure for upgrading HaaS to a new version.
+
+1. Read the release notes, as they may contain version-specific information.
+2. Stop the HaaS services (webserver and HaaS network daemon).
+3. Download and install the new version of HaaS::
+
+     $ git clone https://github.com/cci-moc/haas
+     $ python setup.py install
```
henn: Should this be `python setup.py install --upgrade` ?

docs/UPGRADING.rst
```diff
@@ -0,0 +1,20 @@
...
+
+3. Upgrade the database::
+
+     $ haas-admin db upgrade heads
```
henn: Could you explain what `heads` means here?

docs/UPGRADING.rst
```diff
@@ -0,0 +1,20 @@
+Upgrading HaaS
+==============
+
+This file describes the procedure for upgrading HaaS to a new version.
+
+1. Read the release notes, as they may contain version-specific information.
+2. Stop the HaaS services (webserver and HaaS network daemon).
```
henn: Maybe include an example:
```markdown
2. Stop the HaaS services (webserver and HaaS network daemon). For example: `systemctl stop httpd; systemctl stop haas_network`
```

docs/UPGRADING.rst
```diff
@@ -0,0 +1,20 @@
...
+
+     $ git clone https://github.com/cci-moc/haas
+     $ python setup.py install
+
+3. Upgrade the database::
+
+     $ haas-admin db upgrade heads
+
+4. If additional extensions have been added to ``haas.cfg``, re-run ``haas db
+   create``, which will create any tables needed by those extensions.
```
henn: 1. Do sysadmins need to worry about removed extensions?
2. Should this be `haas-admin db create`?
3. Does `upgrade` do this as well?

docs/migrations.md
```diff
@@ -0,0 +1,108 @@
...
+
+Finally, run `haas-admin db upgrade heads` to upgrade the database according
+to your new migration.
+
+### If the extension already has migrations, or the script is for HaaS core
+
+To generate a new migration script, execute:
+
+    haas-admin db migrate --head ${branch_name}@head \
+        -m 'Summary of change'
```
henn: Could you make this (and the one above)
```markdown
-m '<Summary of change>'
```
So that folks know that it's not the literal string?

examples/haas.cfg
henn: I like including an option that works with minimal setup (which would be sqlite), but if you feel strongly, @SahilTikale I won't object.

henn: Maybe we could add an example of a postgres line commented out if we keep it the way it is?

haas/commands/admin.py
```diff
@@ -0,0 +1,16 @@
...
+
+manager = Manager(app)
+manager.add_command('db', db.command)
+
+
+def main():
+    """Entrypoint for the haas-admin command."""
+    config.load()
+    config.configure_logging()
+    config.load_extensions()
```
henn: I feel like we do this mantra in several places now (`config.load(); config.configure_logging(); config.load_extensions()`), including cli.py and haas.wsgi, in violation of [DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself).

What do you think about placing these into a single call in config.py?

examples/haas.cfg
zenhack: @henn, there's already a postgres example commented out a few lines below.

Again, no strong feelings on this part; come to a consensus and I'll make the change.

conftest.py
zenhack: Unfortunately no; pytest won't recognize it otherwise.

docs/UPGRADING.rst
zenhack: `setup.py` doesn't have the `--upgrade` option (only pip does), so if we wanted to change it we'd have to change it to pip. The `--upgrade` option won't have any different behavior when used to install a package from a file path; it's only useful when specifying packages to be fetched.

As far as reasons for preference, the only think I can think of is the case where pip isn't present on the system, which is not unheard of.

The reason I wrote it this way is because it's consistent with what's already in `INSTALL.rst`.

docs/UPGRADING.rst
zenhack: The alembic docs go into the full detail (please do read them), but basically it says to upgrade *all* branches (one per extension with scripts, plus haas core) to the latest version. The default is head, which will spit out an error if there's more than one.

I don't know that we want to go into the full detail in this document. `haas-admin db` has a lot of stuff that's mostly of interest to developers. It might make sense to move that to a `haas-dev` command and write a few custom sub-commands to `haas-admin` doing just the basics.

docs/UPGRADING.rst
zenhack: re (1), we don't really support that use case. I'll add a note to that effect.

re (3), no, it doesn't.

Good catch on (2); I'll fix.

.travis.yml
zenhack: I added `-n auto` to travis for sqlite only. I think writing the logic to distribute the tests amongst multiple postgres databases is going to be nontrivial, enough so that it probably should be its own issue.

haas/commands/admin.py
zenhack: done.

zenhack:
@henn, I've either implemented or replied to each of your comments. Please have another look.

.travis.yml
```diff
@@ -16,4 +16,11 @@ install:
 before_script:
   - sh -c "if [ $DB = postgres ]; then psql -c 'CREATE DATABASE haas_travis;' -U postgres; fi"
   - sh -c "if [ $DB = postgres ]; then cp ci/testsuite.cfg.postgres testsuite.cfg; fi"
-script: "py.test tests/unit tests/stress.py"
+script: |
+    # On SQLite we run the tests in parallel. This speeds things up
+    # substantially, but i  currently only safe for sqlite, since it
```
henn: typo: i=is

zenhack: Quoting henn (2016-03-31 15:51:49)

>    typo: i=is

Good catch, fixed.

docs/UPGRADING.rst
henn: How about a link to the appropriate alembic documentation here?

.travis.yml
henn: Great! Thanks!

I filed #565 for the postgres issue.

haas/server.py
```diff
@@ -43,14 +45,13 @@ def stop_orphan_consoles():
     These may exist if HaaS was shut down uncleanly.
     """
     # Stop all orphan console logging processes on startup
-    db = model.Session()
-    nodes = db.query(model.Node).all()
+    nodes = db.session.query(model.Node).all()
```
henn: Would the pattern of `model.Node.query.all()` be a little more readable (here and other places), similar to flask/sqlalchemy's [quickstart](http://flask-sqlalchemy.pocoo.org/2.1/quickstart/)?

henn:
Thanks for all this great work, Ian.

Besides the comments I made, one other request is that we add a DB migration test in order to ensure that applying all of the alembic updates works. This would ideally detect times when folks change the DB without writing migrations as well as bad migration scripts themselves.

Perhaps this could work when running the postgres tests by creating a DB from the output of a `pg_dumpall` run on an old version of the DB, applying all of the migration scripts, then running some set of tests.

What do you think?

zenhack:
@henn, re: testing, I think I suggested something similar way up top, asking you to weigh in. Since we're of similar mind I'll get started on that.

haas/migrations/env.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    url = config.get_main_option("sqlalchemy.url")
+    context.configure(url=url)
+
+    with context.begin_transaction():
+        context.run_migrations()
+
+
+def run_migrations_online():
+    """Run migrations in 'online' mode.
```
zenhack: This file is almost entirely boilerplate generated by alembic, so I didn't actually write that. Maybe the thing to do is to put a note at the top saying that, and referring to the docs? Iirc, offline mode spits out SQL that you can run later, while online mode goes in and makes the changes itself.

docs/UPGRADING.rst
zenhack: Can do.

haas/server.py
zenhack: Quoting henn (2016-04-01 00:40:21)

>    Would the pattern of model.Node.query.all() be a little more
>    readable (here and other places), similar to flask/sqlalchemy's
>    quickstart?

I think that's a good idea; I mostly did it this way because it made the
change more mechanical. I'll go through and update it.

zenhack:
@henn, I've addressed everything but the testing. I'm going to go back to bed, I'll get to that soonish.

zenhack:
@henn, the testing stuff is in place; please have a look.

Additionally, I've pushed a4db137, which obviates (and removes) `conftest.py`; I'd thought it was weird that it was scanning that file in the first place.

docs/migrations.md
```diff
@@ -0,0 +1,144 @@
...
+  These will cause permission errors if your database user is not root.
+- Delete all statements of the form ``ALTER TABLE ... SET OWNER TO
+  ...``; these may cause failures if the connection details on the
+  machine which runs the tests are different from yours, since the
+  refer to specific database users/roles.
+- At the top of the file, document the following:
+  - In roughly what stage of HaaS's development this dump was taken,
+    e.g. "Just after re-integrating flask" or "first integration of
+    openvpn"
+  - The commit hash of the revision of HaaS which created the database.
```
henn: Could you make this "git commit hash". It might seem obvious, but alembic has its own hashes so it'd be good to be explicit.

haas/test_common.py
```diff
@@ -59,6 +60,166 @@ def config_testsuite():
         })
 
 
+def initial_db():
```
henn: Maybe move this function into a separate file, or at the very least to the bottom of this file?

It's so big that it could cause the small helper functions to get lost.

tests/deployment/ipmi.py
```diff
@@ -55,19 +53,17 @@ def server_init():
 class TestIpmi():
     """ Test IPMI driver calls using functions included in the IPMI driver. """
 
-    def collect_nodes(self, db):
+    def collect_nodes(self):
         """Collects nodes in the free list."""
-        free_nodes = db.query(Node).filter_by(project_id=None).all()
+        free_nodes = Node.query.filter_by(project_id=None).all()
```
henn: Where are we importing Node from?

tests/unit/migrations.py
```diff
@@ -0,0 +1,185 @@
...
+            drop_tables()
+            f()
+
+        return get_db_state()
+
+    upgraded = run_fn(lambda: load_dump(filename))
+    fresh = run_fn(fresh_create)
+    drop_tables()
+
+    def censor_nondeterminism(string):
```
henn: What about adding an argument to `Headnode.__init__()` that takes a uuid with a default of generating a new uuid?

Also, if you do like that approach, please change the default to use `uuid.uuid4()` instead of `uuid.uuid1()` so that we get some randomness.

henn:
Thanks. All set for comments on this pass.

I'm really excited that we have test that verify the schema can migrate!

tests/unit/migrations.py
zenhack: Quoting henn (2016-04-05 01:20:23)

>    What about adding an argument to Headnode.__init__() that takes a
>    uuid with a default of generating a new uuid?

The current solution is more extensible to other potential sources of
non-determinism, and doesn't require special support in the non-test
code. Especially the latter part of that I like, since it avoids
complicating other interfaces. However, if you have a strong opinion
here I'll defer; let me know.

>    Also, if you do like that approach, please change the default to use
>    uuid.uuid4() instead of uuid.uuid1() so that we get some
>    randomness.

I don't know that the randomness actually does much for us here --- it
would be perilous to assume uuid4 is *cryptographically* random (I
imagine it's not), and otherwise any mechanism for avoiding accidental
collisions is just as good, no? Not that I'm *opposed* to the random
one, but I'm dubious of the motivation.

Right now we're only using it for having unique libvirt names; we could
just use the primary key for that. That would require renaming all the
VMs though, which is probably not worth the trouble.

tests/deployment/ipmi.py
zenhack: Quoting henn (2016-04-05 01:03:36)

>    Where are we importing Node from?

Good question! I was just mechanically making the change, and assumed
since it was being referred to as `Node` already, the change was safe. I
bet this is broken. @SahilTikale, you wrote this, can you weigh in?

haas/test_common.py
zenhack: Quoting henn (2016-04-05 01:00:12)

> > +def initial_db():
>
>    Maybe move this function into a separate file, or at the very least to
>    the bottom of this file?

Moved it to the bottom of the file. `test_common` has gradually been
getting more cluttered, and we probably ought to do some general
refactoring. I'll open a separate issue for that.

docs/migrations.md
zenhack: Quoting henn (2016-04-05 00:46:48)

>    Could you make this "git commit hash". It might seem obvious, but
>    alembic has its own hashes so it'd be good to be explicit.

Done. Fyi, alembic actually *doesn't* have hashes, those are just random
strings. I agree the extra clarity is good though.

haas/test_common.py
zenhack: Opened a bug #566

zenhack:
@henn, I think that's everything.

> I'm really excited we have a test to verify that the schema can migrate!

Me too!

tests/unit/migrations.py
henn: That's an interesting point. I like it.

Could you use `re.sub()` instead of the join/split?

zenhack: Done. Not sure why I didn't do it that way to begin with. Anything else?

zenhack:
Fixed some trivial merge conflicts.

henn:
+1

zenhack:
Quoting henn (2016-04-06 23:54:24)

>    +1

Great! Let's get a second reviewer on this. @SahilTikale, care to have
another look?

zenhack:
@pjd-nu, @gsilvis, @kylehogan, @SahilTikale, a second review from any of you would be welcome.

SahilTikale:
I am a little occupied right now, will do it at the earliest

pjd-nu:
I looked at the latest 4 commits and they look good.

zenhack:
@pjd-nu, don't know if that counts as a sign-off or not. If so, please go ahead and merge.

PR #564, opened 2016-03-31
Basic API Logging

knikolla:
Added basic API logging. New configuration option is read from
haas.cfg with the 'log_dir' of the logs.

```
2016-03-31 14:55:07,961 - haas.rest - INFO - (guest) - API call: list_projects()
```

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -87,6 +87,9 @@ important: ``trunk_nic`` must match your choice of trunk NIC in the "Networking
 For a detailed description of the configuration needed for various switch
 setups, see ``docs/network-drivers.md``.
 
+Logging level and directory can be set in the ``[general]`` section. For more
+information view ``docs/logging.md``.
+
 The file should be placed at ``/etc/haas.cfg``; The ``haas.wsgi``
 script, described below, requires this. Awkwardly, the ``haas``
 command line tool expects the file to be present in its current
--- a/docs/logging.md
+++ b/docs/logging.md
@@ -1,0 +1,29 @@
+# Overview
+HaaS supports basic logging of the API calls received in the below format,
+recording the user who issued the API call (or guest if the user was
+unauthenticated) and the time. Log files are rotated daily.
+
+```
+2016-03-31 14:55:07,961 - haas.rest - INFO - (guest) - API call: list_projects()
+```
+
+# Setup
+Logging can be configured with the following options in `haas.cfg`.
+
+```
+[general]
+log_level = DEBUG
+log_dir = /var/log/haas
+```
+
+`log_dir` specifies the directory where the log files will be stored.
+The HaaS user must have write permissions to this directory.
+If the option is omitted logging to files is disabled.
+
+`log_level` specifies the logging level to record. Valid options are:
+`CRITICAL`, `DEBUG`, `ERROR`, `FATAL`, `INFO`, `WARN`, `WARNING`.
+The default value is `WARNING`, but an option of `INFO` is recommended
+for an API log (A log level of `INFO` is set for API calls).
+
+For more information on logging visit the
+[python 2 documentation](https://docs.python.org/2/howto/logging.html#when-to-use-logging).
--- a/haas/auth.py
+++ b/haas/auth.py
@@ -42,10 +42,18 @@ class AuthBackend(object):
         """
 
     @abstractmethod
+    def get_user(self):
+        """Return a string with the currently authenticated user.
+
+        Return String with username of the currently authenticated user,
+        or None if no user is authenticated.
+        """
+
+    @abstractmethod
     def _have_admin(self):
         """Check if the request is authorized to act as an administrator.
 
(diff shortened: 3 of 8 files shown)


SahilTikale:
Can you introduce some unit tests for capturing the logging functionality.

knikolla:
@gsilvis 
@SahilTikale

haas/rest.py
```diff
@@ -209,9 +209,13 @@ def _rest_wrapper(f, schema):
 
     def wrapper(**kwargs):
         kwargs = _do_validation(schema, kwargs)
-        logger.debug('Got api call: %s(%s)' %
-                     (f.__name__, _format_arglist(**kwargs)))
+
         init_auth()
+        username = '(guest)' if not hasattr(local, 'auth') \
+            else flask.request.authorization.username
```
gsilvis: Note that this line puts an additional constraint on auth methods----they must set flask.request.authorization.username .  We should document this.  We could also potentially have the null auth backend set username to (no user), and get rid of the special case

gsilvis:
@SahilTikale We talked about this in person, and decided we don't need unit tests for this

haas/auth.py
```diff
@@ -42,10 +42,17 @@ def authenticate(self):
         """
 
     @abstractmethod
+    def get_user(self):
+        """Return a string with the currently authenticated user.
+
+        Return String with username of the currently authenticated user,
+        or None if no user is authenticated."""
```
gsilvis: There should be a newline before the ending ``"""```, technically

[We want to enable PEP-8 autochecking eventually, so it's worth it to fix]

haas/ext/auth/null.py
```diff
@@ -10,6 +10,9 @@ class NullAuthBackend(auth.AuthBackend):
     def authenticate(self):
         return True
 
+    def get_user(self):
+        return False
```
gsilvis: By the docs, this should be ``None``, not ``False``

gsilvis:
Besides the two little comments, looks good to me.  +1

knikolla:
@gsilvis addressed your comments.

gsilvis:
Thanks

haas/ext/auth/database.py
```diff
@@ -137,6 +137,12 @@ def authenticate(self):
         else:
             return False
 
+    def get_user(self):
+        if hasattr(local, 'auth') and local.auth:
```
knikolla: @gsilvis after your +1 I changed this line to also check if `local.auth` is not None (in the case of a failed authentication local.auth would be set to None therefore the first condition isn't enough and the server returned 500 since it then tried to access local.auth.label. Caught this bug during testing.

gsilvis: Hm, failed auth doesn't just reject the request in haas?

gsilvis:
+1 again---I still don't see any problems, and this time, I hope I'm right :)

haas/ext/auth/database.py
knikolla: @gsilvis If you see the authenticate method. It sets `local.auth = None` before doing anything. So regardless of if the authentication was successful or not, `local` will have an attribute `auth`. 

Therefore when I call `get_user` on the backend the flow is:
* Is there an attribute called `auth` in `local`?
* Yes, access `local.auth.label` - This failed because `local.auth` is None. 

-

gsilvis: Okay, I understand now.  Looks good, yeah

SahilTikale:
We need some documentation about logging feature. 

What is the default level ?
What other levels can be configured ? 
Look at this for example:
https://github.com/CCI-MOC/haas/blob/master/docs/obm-drivers.md

Although not directly related, it will give you some idea about documenting a new feature introduced under HaaS.

INSTALL.rst
```diff
@@ -87,6 +87,14 @@ important: ``trunk_nic`` must match your choice of trunk NIC in the "Networking
 For a detailed description of the configuration needed for various switch
 setups, see ``docs/network-drivers.md``.
 
+Logging level and directory can be set in the ``[general]`` section.
+``log_dir`` shouldn't end with a trailing slash and the HaaS user must have
+ownership of the specified directory.::
```
SahilTikale: this line "user must have ownership ..." should go in the documentation about in may be docs/logging.md

docs/logging.md
```diff
@@ -0,0 +1,29 @@
...
+HaaS supports basic logging of the API calls received in the below format,
+recording the user who issued the API call (or guest if the user was
+unauthenticated) and the time. Log files are rotated daily.
+
+```
+2016-03-31 14:55:07,961 - haas.rest - INFO - (guest) - API call: list_projects()
+```
+
+# Setup
+Logging can be configures with the following options in `haas.cfg`.
```
SahilTikale: typo `configures` should be `configured`

SahilTikale:
+1 merging.

Issue #508, opened 2015-11-12
reset_port() and/or reset_nic()


henn:
We need a function to reset a switch port to be on the VLANs that HaaS thinks they should be on.

This would be useful when:
- someone manually modifies the switch port to be on certain VLANs 
- troubleshooting network connectivity problems (admin can use reset_nic() flavor)
- when in environments where machines come in and out of haas on a regular basis (we'll want to reset the port when bringing the machine in so we can ensure it's not connected to anything)
- when a switch is replaced, whereby NICs might be connected to different ports and a reset_nic() could be called to set the NIC properly.

henn:
This has two components:

1. Use switch-specific commands to reset a port to an empty state
2. Apply haas's views of which VLANs a port should be connected to

kylehogan:
After reading some of the docs I get the impression that reset is the wrong word here. Reset implies reverting to some default state. 

I think what we want is to remove every VLAN from a port. 

e.g. the command 'no switchport allowed vlan' is described as 'reset a trunking characteristic to
the default' which would be "its Native VLAN and the port belongs to either all
VLANs or only to the Default VLAN depending on a value of parameter
Trunk Port Default Configuration."

while 'switchport allowed vlan none —Specifies an empty VLAN list The port does not belong to any VLAN'

Is this correct?

SahilTikale:
I understand what you are saying @kylehogan . I agree that reset_<device> sounds like resetting to its default state. We are not doing that. What we are doing here is a restore operation.
We should call this calls `port_restore` or `nic_restore`

kylehogan:
I like `port_restore` and `nic_restore` better as well. Was I correct in assuming that I would want to remove access for all VLANs and that this would be considered 'restoring' it?

zenhack:
So in a comment on #621, @henn suggested sending the the actual networking operations to the backend via the usual mechanism, rather than handling them in the api call proper. There's a problem with this, that requires at least a bit of rearchitecting: right now the only thing we can send to the network daemon is "attach this port to this network on this channel" or "detach this port from any network on this channel." This interface is unable to express the operation we want to preform here.

So the question becomes, how do we extend this?

One possibility: have different types of networking actions. We could use a similar inheritance mechanism to have subclasses of NetworkingAction, and differentiate on the other end.

One awkwardness with this is right now we're passing the networking action to the session, so all of the knowledge about how to interpret a networking action needs to be per-switch, which means each driver needs to implement it. Ick.

[...]

PR #567, opened 2016-04-05
revert_port

kylehogan:
addresses issue #508 

- [ ] call to remove all vlan access
- [x] show port call
- [x] revert_port can reinstate network attachments that were present in haas database
- [ ] test the case where there are multiple network attachments


kylehogan:
@ron-unrau 
my show_switch can be seen in the api here. I also have a few unit tests for it. Can you let me know if you decide to use any of it and I can then review it in your pull request and remove it from here? Thanks!

ron-unrau:
The basic framework of my implementation is exactly the same as yours, the
only difference is that I was going to return the information provided in
switch_register, while yours is returning the info from
switch_register_port.

I don't speak github yet, but in my world we would have said: since your
review is pretty much done, you may as well push your changes to the
mainline, and then I can pull them into my branch and add the extra fields.
(Did that make any sense to you?)

I don't think anyone is really waiting for the stuff in my show_switch
implementation, and besides, I have no idea how to handle the polymorphism
yet anyway. In summary, I think it's fine if my changes wait for a week.

< ron

kylehogan:
This pull request is blocked until it can get tested in a development environment because of the revert_port functionality so nothing in here will get merged soon. Because of that it may be better to keep show_port as your separate pull request.

matsuuran:
@kylehogan @ron-unrau 
just wanted to see if this would be able to be tested on the server @SahilTikale is setting up

shwsun:
I have tested the functions included in devhaas, but it looks like some of them are broken.

I fixed the schema before that. If there are no PR regarding any functions related then I guess I have to fix broken ones by myself.

matsuuran:
@shwsun just wanted to see how this was going

shwsun:
@matsuuran I guess we should close this one and I will reference it in my PR.

matsuuran:
@henn @SahilTikale @gsilvis can you close this? there is another PR open for this #621

henn:
Done!

Issue #500, opened 2015-10-14
Query interface should sort entries


henn:
It's an optional argument to [json.dump()](https://docs.python.org/2/library/json.html#basic-usage): sort_keys

kylehogan:
This is currently being used. Are there any places it still needs to be added or can this be closed?

henn:
Looking through [haas/api.py](https://github.com/CCI-MOC/haas/blob/master/haas/api.py), I see:

- [ ] list_projects
- [ ] list_free_nodes
- [ ] list_project_nodes
- [ ] list_project_networks
- [ ] list_project_headnodes
- [ ] list_headnode_images

SahilTikale:
#568 addresses it.

ron-unrau:
Sorry to jump in after the horses have already left the barn, but what motivated this issue? Do the clients require the lists to be sorted for some reason? If some do and some don't, it's a lot more scalable to let the clients do their own sorting. If the server has to do the sorting, it's more efficient to keep the lists in sorted order so the server doesn't have to continuously sort the same list again and again for each reply.

PR #568, opened 2016-04-05
Sorts output that provides lists

SahilTikale:
Addresses issue #500 
Short, simple PR. 
@henn @knikolla 
If you have time for reviewing this.

### DIFF
--- a/haas/api.py
+++ b/haas/api.py
@@ -44,8 +44,7 @@ def list_projects():
     """
     get_auth_backend().require_admin()
     projects = local.db.query(model.Project).all()
-    projects = [p.label for p in projects]
-    projects.sort()
+    projects = sorted([p.label for p in projects])
     return json.dumps(projects)
 
 @rest_call('PUT', '/project/<project>')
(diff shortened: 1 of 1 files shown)


knikolla:
+1, Looks good.

pjd-nu:
+1

zenhack:
So, just noticed this because it caused some (utterly trivial) merge conflicts with #561. This is fine, but I would point out the existence of SQL's `ORDER BY`; you can have the DB do the sorting.

ron-unrau:
Do you know if SQLAlchemy exports/supports 'ORDER BY'? I would like to use it, but I have no idea how it would look in python...

zenhack:
Quoting Ron Unrau (2016-04-09 01:46:10)

>    Do you know if SQLAlchemy exports/supports 'ORDER BY'? I would like to
>    use it, but I have no idea how it would look in python...

It's just an `.order_by` method on the query object. We might have some
examples of this in the HaaS source tree already, but if not check out
the SQLAlchemy docs.

PR #570, opened 2016-04-06
Adds logic to switch_register cli to accomodate disparate inputs

SahilTikale:
Currently supporting mock, dell and nexus

### DIFF
--- a/haas/cli.py
+++ b/haas/cli.py
@@ -371,10 +371,48 @@ def switch_register(switch, subtype, *args):
     eg. haas switch_register mock03 mock mockhost01 mockuser01 mockpass01
     """
     switch_api = "http://schema.massopencloud.org/haas/v0/switches/"
-    switchinfo = { "type": switch_api+subtype, "hostname": args[0],
+@cmd
+def switch_register(switch, subtype, *args):
+    """Register a switch with name <switch> and
+    <subtype>, <hostname>, <username>,  <password>
+    eg. haas switch_register mock03 mock mockhost01 mockuser01 mockpass01
+    """
+    switch_api = "http://schema.massopencloud.org/haas/v0/switches/"
+    if subtype == "nexus":
+        if len(args) == 4:
+            switchinfo = { "type": switch_api+subtype, "hostname": args[0],
+                        "username": args[1], "password": args[2], "dummy_vlan": args[3] }
+        else:
+            sys.stderr.write('ERROR: subtype '+subtype+' requires exactly 4 arguments\n')
+            sys.stderr.write('<hostname> <username> <password> <dummy_vlan_no>\n')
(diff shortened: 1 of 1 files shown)


SahilTikale:
@knikolla @henn 
simple PR, have a look when you can.

SahilTikale:
@kylehogan please have a look if you have time.

knikolla:
-1. Instead of using *args, you should use **kwargs.

SahilTikale:
cli should be taking command line arguments hence args is the appropriate thing. 
I do not wish to use kwargs as it will compels the command line user to know what keys are there for each switch driver. 

About code duplication, I thought about it but I do not think that I can simply it.
I can make the error as a single function that can be invoked if desired input is not received, but that would make it more complex as it will require defining new variables to suit each case.

Also, this is to keep current HaaS code operational. 

Once I am done with the client library, the CLI is going to be completely revamped.

knikolla:
@SahilTikale yeah, I backtracked on my kwargs argument after I saw the `cmd` decorator.

haas/cli.py
```diff
@@ -371,11 +371,49 @@ def switch_register(switch, subtype, *args):
...
+            return
+
+    else:
+        sys.stderr.write('ERROR: Wrong subtype supplied\n')
+        return
+
     url = object_url('switch', switch)
     do_put(url, data=switchinfo)
 
+
```
knikolla: extra blank line

SahilTikale:
@knikolla  fixed it.

knikolla:
+1

kylehogan:
sorry for the late reply, I missed the initial email.
It looks good to me, your decision on how to handle the errors makes sense.

My only comment is that it might be helpful if the error for 'wrong subtype supplied' was instead either 'invalid subtype' or said that subtype must be 'nexus, mock, or powerconnect55xx' 
Listing all the subtypes might not make sense if there end up being many of them, but with only three it seems reasonable. This way it is clear that the subtype supplied is not one that is supported rather than 'wrong' which could imply that the subtype was valid, but a different subtype had been expected.

knikolla:
@SahilTikale, I agree with Kyle. We are already having to handle all the switches in a separate code block. So we know what we support and can provide feedback to the user on the correct types.

gsilvis:
This logic should be in the switch driver---that way it's extensible.

gsilvis:
After some conversation with @SahilTikale, we agreed we don't need to do this nicely... yet.

However, please add a TODO or FIXME comment saying that this current design needs to be changed so that the CLI doesn't need to know about every backend.

gsilvis:
(and once you do that, +1)

gsilvis:
+1 and merging

PR #574, opened 2016-04-15
Add recommendation to persistently disable SELinux on CentOS/RHEL

henn:


gsilvis:
At first I thought 'Why are we now recommending this?'.  But then I noticed that we were already recommending it.  Given that it's there, I agree, we should say how to do it.  +1

SahilTikale:
All tests failing. Once fixed I will give a +1

gsilvis:
I don't know why these are failing, but this is a pure documentation change, so I don't think it's @henn's fault

gsilvis:
Rerunning tests...

gsilvis:
@SahilTikale, you gave a +1 conditional on tests passing, so I'll merge now.

Issue #580, opened 2016-04-26
Calls related to object creation in HaaS database fails 


SahilTikale:
Following is the behavior of the bug:

When trying to register new nodes or Users, it fails with the following error despite of giving correct inputs. 

INPUT: `haas node_register dummyNode01 ipmi host01 user01 pass1234`

OUTPUT:

```
Unexpected status code: 400
Response text:
{"msg": "The request body {u'obm': {u'host': u'host01', u'password': u'pass1234', u'type': 
u'http://schema.massopencloud.org/haas/v0/obm/ipmi', u'user': u'user01'}} is not valid for this request.", 
"type": "ValidationError"}
```

Other calls like: `haas user_create <username> <is_admin>` are also failing with the same error.

**Temporary Fix:**
For every api calls that are failing with the above described behavior do the following:

-- Modify the respective call in api.py and mention the parameter passed as URL explicitly in the json file.
**Example**
Modify the `node_register` call:
Original:

``` python
@rest_call('PUT', '/node/<node>', schema=Schema({
    'obm':{
        'type': basestring,
        Optional(object):object,
        },
}))
def node_register(node, **kwargs):
    """Create node.

    If the node already exists, a DuplicateError will be raised.
    The node is initially registered with no nics; see the method node_register_nic.
    """
```

Include the line `'node': basestring,` as follows:

``` python
@rest_call('PUT', '/node/<node>', schema=Schema({
    'node': basestring,
    'obm':{
        'type': basestring,
        Optional(object):object,
        },
}))
def node_register(node, **kwargs):
    """Create node.

    If the node already exists, a DuplicateError will be raised.
    The node is initially registered with no nics; see the method node_register_nic.
    """
```

After this change the code will work as expected.

henn:
@ravisantoshgudimetla and I just finished running git bisect with `haas node_register node1 ...`, and it pointed to ed42146817e0ad4bc3a04e4272a8638f70b2c141

It looks like @zenhack hit something similar in 9400f6953aa03d8952f4de812245a3e184262e90

henn:
FYI - I think that the automated apache tests that @knikolla is working on should catch these types of issues in the future.

naved001:
Looks like this was fixed by #581 see commit [message here](https://github.com/CCI-MOC/hil/pull/581/commits/973a176c22b957b7e89a3280d6d3dd8a62f3f557).

PR #581, opened 2016-04-28
Fix arg validation when passing arguments in the URL + body

henn:
We were checking both URL arguments & body arguments against the `Schema`, which only verifies body arguments.

Also address a log injection problem, where unsanitized user data was being used directly to raise an exception back to the original user.

### DIFF
--- a/docs/developer-guidelines.md
+++ b/docs/developer-guidelines.md
@@ -27,12 +27,21 @@ If you're looking to code, there are a few ways to help:
 * IRL (In Real Life): The MOC group office is on the BU campus, at 3 Cummington Mall, Boston, MA room 451. Anyone interested in HaaS is welcome to drop in and work there.
 * Email: HaaS developers or anyone else wishing to stay up to date should subscribe to haas-dev-list@bu.edu by sending a plain text email to majordomo@bu.edu with "subscribe haas-dev-list" in the body.
 
-# Coding style
+# Coding styles/conventions
 
 By default, HaaS (like many other python projects) uses
 [PEP8](https://www.python.org/dev/peps/pep-0008/) as its naming guide, and
 [PEP257](https://www.python.org/dev/peps/pep-0257/) for documentation.
 Departures are acceptable when called for, but should be discussed first.
+
+## REST API calls
+
+Arguments passed to externally-visible API calls are verified by a combination
+of the Flask API and
+
+### Body-args (Schema)
+
+
 
 ## Often-used code
 In certain cases, one will encounter heavily repeated code that gets run once per API call such as this:
--- a/haas/rest.py
+++ b/haas/rest.py
@@ -177,8 +177,8 @@ def _do_validation(schema, kwargs):
     body = flask.request.get_json(force=True)
 
     validation_error = ValidationError(
-        "The request body %r is not valid for "
-        "this request." % body)
+        "The request body is not valid for this request.")
+
     if not _validates(body):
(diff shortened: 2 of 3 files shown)


henn:
@zenhack, since you are the most familiar with this code, would you mind taking a look?

ravisantoshgudimetla:
Looks good to me +1

ravisantoshgudimetla:
@henn : I am getting validation error now for project_connect_node. It seems the error still persists.

zenhack:
I don't understand what the vulnerability is. The actual fix for #580 looks good.

zenhack:
Actually, there's a bit you're missing: if no schema is provided, we're building one that expects everything to be a string. That needs adjustment as well.

ravisantoshgudimetla:
@zenhack : I am trying to debug this issue and stuck at schema.validate() method. Is this validate method coming from dell.py(ext/switches) or from some other place. Also, for debugging purpose, I tried disabling this if condition(_validates(value)) and all the rest api calls are working fine. Can we try checking if this is causing problem. Also, will body be json?, should we do json.loads before sending to schema.validate?

zenhack:
The schema is an object from this library:

Https://pypi.python.org/pypi/schema

Body is json, flask does that much parsing for us; we don't call json.loads ourselves at all.

Also look at _make_schema_for, which builds the schema for anything where we don't explicitly specify. Problem is it includes all arts, including ones in the URL instead of the body. 

I'll likely be unresponsive for the rest of the day, but will try to check back tonight. Happy hacking.

ravisantoshgudimetla:
@zenhack : Asusual you are spoton Ian :100:  I did that change and it looks like it works. I will submit a pull request for it.

ravisantoshgudimetla:
I actually did a git diff and reverted few lines in rest.py to include the old changes. I am not sure about the flask routing but will it include werkzeug.routing as I had to include that exclusively.

henn:
> I don't understand what the vulnerability is. The actual fix for #580 looks good.

I can think of 2 possible issues from sending un-sanitized user data back to the user through an exception:
1. [log injection](https://www.owasp.org/index.php/Log_injection)
2. Imagine an load balancer sitting between the user and server, where a malicious user could cause arbitrary text to be sent from the server to the load balancer. That user could potentially cause the load balancer to think that one http transaction had completed and maybe merge some information into another pipeline by causing the server to transmit certain http headers back.

zenhack:
There isn't actually any unsanitized data here:

1. If we get this far, the body is already valid JSON, otherwise the call to get_json above would have rejected it. So any `\r` nonsense or the like is already going to be nicely escaped by virtue of being part of a JSON string.
2. We escape it *again* when constructing the string -- the `%r` format specifier calls `repr`, which in this case will give you a properly escaped representation of the python value. This is probably not to be relied on for arbitrary types, but for most of the built in types it's pretty safe -- strings, dicts, lists, etc. So that alone would be sufficient -- if it were an arbitrary string (which it's not) it would be escaped.
3. The result gets  packed up in another JSON object for reporting the error, per the usual APIError semantics. By the time the response gets on the wire, it's been escaped *three times.* The last layer doesn't help with the log injections, but the first two (each) more than cover it. A load balancer shouldn't even be looking under the last layer, as that stuff is in English.

I don't actually have an objection to removing it -- It's not  like including the body of the request in the response is going to be super informative anyway -- but I don't think there's an actual vulnerability here.

henn:
This is replaced by #596. Closing

PR #582, opened 2016-04-29
Body validation bug fix

ravisantoshgudimetla:
Fixes #581. Did as suggested by Ian.


ravisantoshgudimetla:
Is the test case still valid  where we have arguments coming from both URL and body?

ravisantoshgudimetla:
Closing this pull request, as the @henn suggested to have all the calls to have explicit body checks.

PR #586, opened 2016-05-04
Leasing script

ravisantoshgudimetla:
Leasing script with few parameters related to HaaS. This is for revoking nodes back from project to free pool.


ravisantoshgudimetla:
Wrong branch. Closing the pull request

Issue #589, opened 2016-05-06
Migration docs: be more explicit re: setting/not setting down_revision


zenhack:
The migrations doc says to change it to None only for new extensions, not for haas core. This is correct, but they're not very explicit about *not* changing this for existing extensions and/or haas core. Per the discussions in #554, this is likely to be misinterpreted. We should change it to be more explicit that this should be left-alone when not adding a new extension.

//cc @kylehogan, @henn

zenhack:
@kylehogan, I included this in the pr I sent you, so once #554 gets merged, we can close this.

matsuuran:
@zenhack I've been noticing there are a several issues that are still open, is it okay if I just do a little house keeping and close them?

zenhack:
If they're obviously fixed, go for it. Just leave a comment explaining why you're closing. I'll take care of this one.

PR #590, opened 2016-05-09
Updates to migration documentation

kylehogan:
@zenhack and @henn 

Here's some updates to the migration docs reflecting the troublesome bits from network ACL.

Do we want specifics on how branching works?

We might want some more detail on generating the new database - particularly where the migration script loads it from so developers can create a new db under a different name if they don't want to scrap their current one.

Also, more info on data vs. schema migrations might be helpful.

addresses #589

### DIFF
--- a/docs/migrations.md
+++ b/docs/migrations.md
@@ -96,8 +96,34 @@ To generate a new migration script, execute:
 Where `${branch_name}` is either the name of the extension's module (if the
 script is for an extension) or `haas` (if the script is for HaaS core).
 
-Alembic will generate a migration script in the appropriate directory. Again,
-sanity check the output; Alembic's guesses should not be trusted blindly.
+Alembic will look at the differences between the schema in the database and
+the one derived from the HaaS source, and generate its best attempt at a
+migration script. The script will be stored in `haas/migrations/versions`.
+
+The value of `down_revision` should be the identifier of the previous migration script.
+The value of `branch_labels` should be `('<branch_name>',)` where `branch_name` should match what was used in the command to generate the migration script.
+
+Sanity check the output; Alembic often does a good job generating scripts, but
+it should not be trusted blindly.
+
+## Notes on Generating Migrations and Checking the Output
+
+### State of the Database
+
+The database being loaded to generate the migration should match the schema of the cuttent master branch.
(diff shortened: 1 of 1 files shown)


docs/migrations.md
```diff
@@ -96,8 +96,34 @@ To generate a new migration script, execute:
...
 script is for an extension) or `haas` (if the script is for HaaS core).
 
-Alembic will generate a migration script in the appropriate directory. Again,
-sanity check the output; Alembic's guesses should not be trusted blindly.
+Alembic will look at the differences between the schema in the database and
+the one derived from the HaaS source, and generate its best attempt at a
+migration script. The script will be stored in `haas/migrations/versions`.
+
+The value of `down_revision` should be the identifier of the previous migration script.
+The value of `branch_labels` should be `('<branch_name>',)` where `branch_name` should match what was used in the command to generate the migration script.
```
zenhack: Please hard-wrap paragraphs to a maximum of 79 chars; without that it's annoying to read in a terminal.

docs/migrations.md
```diff
@@ -96,8 +96,34 @@ To generate a new migration script, execute:
...
+
+### [Data Migration][8] vs. Schema Migration
+
+The migrations Alembic generates are schema migrations: they will create/delete tables, columns, and relationships, but they do not populate these with data. This can be an issue, particularly in the case of a new relationship that would apply to existing data. None of the existing data will be accessible via the new relationship unless the data itself is specifically migrated as well. 
+
+This can be done by directly encoding data within the script and using a command like `bulk_insert()`, executing a SQL statement to SELECT the data and INSERT it into the new column, or by creating a live interaction with the database. 
+
+lines 28-32 in ``89630e3872ec_network_acl.py`` show an example of executing a SQL statement and then using `bulk_insert()` to migrate data
```
zenhack: I'd put the full path to this file, since there are migration scripts in a few different directories.

zenhack:
This looks good. A couple thoughts:

* Strictly speaking, you don't need to generate the script from a database on master, it's just that the auto-detection is more likely to work that way. Most of the other instructions assume you're doing this though, it's probably fine to leave it, but I might add something like "For automatic generation of scripts..."
* re: branches, a summary of how we're using them might make sense (but I think can be a separate pr). with this kind of thing I'm always wary of finding myself writing a tutorial on alembic; should probably refer to the alembic docs on what branches *are* and just stick to how we use them to handle extensions.

I'm happy with merging this once the two comments I made in-line are addressed.

kylehogan:
I think I got everything you asked (plus a few typos)

zenhack:
+1

PR #596, opened 2016-05-18
Explicit schema

zenhack:
Per our discussion, this pr removes the automatic schema generation, makes the schema argument to rest_call mandatory, and adds explicit schemas for all api calls.

### DIFF
--- a/haas/api.py
+++ b/haas/api.py
@@ -34,7 +34,7 @@ from haas.errors import *
                             ################
 
 
-@rest_call('GET', '/projects')
+@rest_call('GET', '/projects', Schema({}))
 def list_projects():
     """List all projects.
 
@@ -47,7 +47,8 @@ def list_projects():
     projects = sorted([p.label for p in projects])
     return json.dumps(projects)
 
-@rest_call('PUT', '/project/<project>')
+
+@rest_call('PUT', '/project/<project>', Schema({'project': basestring}))
 def project_create(project):
     """Create a project.
 
@@ -60,7 +61,7 @@ def project_create(project):
     db.session.commit()
 
 
-@rest_call('DELETE', '/project/<project>')
+@rest_call('DELETE', '/project/<project>', Schema({'project': basestring}))
 def project_delete(project):
     """Delete project.
 
@@ -89,7 +90,9 @@ def project_delete(project):
     db.session.commit()
 
 
-@rest_call('POST', '/project/<project>/connect_node')
+@rest_call('POST', '/project/<project>/connect_node', Schema({
+    'project': basestring, 'node': basestring,
+}))
 def project_connect_node(project, node):
     """Add a node to a project.
 
@@ -106,7 +109,9 @@ def project_connect_node(project, node):
     db.session.commit()
 
 
-@rest_call('POST', '/project/<project>/detach_node')
+@rest_call('POST', '/project/<project>/detach_node', Schema({
+    'project': basestring, 'node': basestring,
+}))
 def project_detach_node(project, node):
     """Remove a node from a project.
 
@@ -139,10 +144,11 @@ def project_detach_node(project, node):
 
 
 @rest_call('PUT', '/node/<node>', schema=Schema({
-    'obm':{
-	'type': basestring,
-	Optional(object):object,
-	},
+    'node': basestring,
+    'obm': {
+        'type': basestring,
+        Optional(object): object,
+    },
 }))
 def node_register(node, **kwargs):
     """Create node.
@@ -162,7 +168,7 @@ def node_register(node, **kwargs):
     db.session.commit()
 
 
-@rest_call('POST', '/node/<node>/power_cycle')
+@rest_call('POST', '/node/<node>/power_cycle', Schema({'node': basestring}))
 def node_power_cycle(node):
     auth_backend = get_auth_backend()
     node = _must_find(model.Node, node)
@@ -173,7 +179,7 @@ def node_power_cycle(node):
     node.obm.power_cycle()
 
 
-@rest_call('POST', '/node/<node>/power_off')
+@rest_call('POST', '/node/<node>/power_off', Schema({'node': basestring}))
 def node_power_off(node):
     auth_backend = get_auth_backend()
     node = _must_find(model.Node, node)
@@ -184,7 +190,7 @@ def node_power_off(node):
     node.obm.power_off()
 
 
-@rest_call('DELETE', '/node/<node>')
+@rest_call('DELETE', '/node/<node>', Schema({'node': basestring}))
 def node_delete(node):
     """Delete node.
 
@@ -201,7 +207,9 @@ def node_delete(node):
     db.session.commit()
 
 
-@rest_call('PUT', '/node/<node>/nic/<nic>')
+@rest_call('PUT', '/node/<node>/nic/<nic>', Schema({
+    'node': basestring, 'nic': basestring, 'macaddr': basestring,
+}))
 def node_register_nic(node, nic, macaddr):
     """Register existence of nic attached to given node.
 
@@ -217,7 +225,9 @@ def node_register_nic(node, nic, macaddr):
     db.session.commit()
 
 
-@rest_call('DELETE', '/node/<node>/nic/<nic>')
+@rest_call('DELETE', '/node/<node>/nic/<nic>', Schema({
+    'node': basestring, 'nic': basestring,
+}))
 def node_delete_nic(node, nic):
     """Delete nic with given name from it's node.
 
@@ -230,6 +240,8 @@ def node_delete_nic(node, nic):
 
 
 @rest_call('POST', '/node/<node>/nic/<nic>/connect_network', schema=Schema({
+    'node'             : basestring,
+    'nic'              : basestring,
     'network'          : basestring,
     Optional('channel'): basestring,
 }))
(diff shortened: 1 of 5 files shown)


zenhack:
//cc @henn, @gsilvis

haas/api.py
```diff
@@ -671,7 +703,9 @@ def switch_register_port(switch, port):
     db.session.commit()
 
 
-@rest_call('DELETE', '/switch/<switch>/port/<path:port>')
+@rest_call('DELETE', '/switch/<switch>/port/<path:port>', Schema({
```
henn: Will the framework handle decoding URL argument encodings? I'm just imagining that ports can have all sorts of special characters like spaces, colons and /'s.

haas/rest.py
```diff
@@ -53,23 +53,22 @@ def rest_call(method, path, schema=None):
     * path - the url-path to map the function to. The format is the same as for
             flask's router (e.g. '/foo/<bar>/baz')
     * method - the HTTP method for the api call (e.g. POST, GET...)
-    * schema (optional) - an instance of ``schema.Schema`` with which to
-            validate the body of the request. It should assume its argument is
-            a JSON object, represented as a python ``dict``.
+    * schema - an instance of ``schema.Schema`` with which to
+            validate the arguments to the function. The schema should expect
+            a dictionary, where the keys are the names of the arguments. If
```
henn: Could you add after `validate the arguments to the function`: , whether specified in the URL or in the JSON body.

haas/rest.py
```diff
@@ -141,49 +121,29 @@ def _do_validation(schema, kwargs):
...
-    # Without `force=True`, this will fail unless the
-    # "Content-Type" header is set to "application/json". We'll be
-    # a little more lienient here; we can discuss whether we want
-    # to tighten this up later, but this makes mucking around with
-    # curl less annoying, and is consistent with pre-flask
-    # behavior.
-    body = flask.request.get_json(force=True)
-
     validation_error = ValidationError(
         "The request body %r is not valid for "
```
henn: Could you please remove the `%r` and accompanying `% body`, per our discussion in #581?

zenhack: Can do, but would you please acknowledge the point I made at the end of that discussion? I think the fact that it confused you is reason enough to remove it, but I'd like to be clear on *why* we're doing it.

haas/rest.py
```diff
@@ -141,49 +121,29 @@ def _do_validation(schema, kwargs):
...
-    for k in body.keys():
-        if k in kwargs:
-            raise validation_error  # Duplicate key in body + url
-        kwargs[k] = body[k]
-    if not _validates(kwargs):
+        "this request." % flask.request.data)
+
+    try:
+        final_kwargs = json.loads(flask.request.data)
+    except ValueError:
```
henn: Will this exception trigger when the body is empty, in addition to poorly-formed JSON in the body?

haas/api.py
zenhack: Yes, it will. Schema doesn't see this; decoding happens in flask or the main web server.

haas/rest.py
zenhack: Yes; in that case data is an empty string, which will fail to parse.

haas/rest.py
zenhack: Will do.

zenhack:
@henn, still waiting on your ACK re: the '%r' issue. I've addressed everything else.

haas/rest.py
henn: Thanks, Ian.

You're probably right in terms of stream-injection (like the load balancer), but I'm still a little concerned that between UTF8 and embedded newlines, passing this data around may not be for the best.

While I'm thinking of it, I just realized that there are currently no restrictions on what folks can name things (like networks with embedded newlines, or JSON-y type of things which might be difficult to figure out), which may also be a problem.

haas/rest.py
```diff
@@ -141,49 +122,29 @@ def _do_validation(schema, kwargs):
...
-        "this request." % body)
-    for k in body.keys():
-        if k in kwargs:
-            raise validation_error  # Duplicate key in body + url
-        kwargs[k] = body[k]
-    if not _validates(kwargs):
+        "this request." % flask.request.data)
+
+    try:
+        final_kwargs = json.loads(flask.request.data)
```
henn: Could we first check to see if the body is empty before running this check? After that, we can make the ValueError fatal?

That way, if the body is empty we can continue with an empty `final_kwargs`, and we can cut down on unexpected behavior.

For example, the user could mistakenly pass invalid JSON as arguments to an API call that has only optional body arguments. The call would then succeed, but not using the arguments the user thought he had passed.

haas/rest.py
```diff
@@ -141,49 +122,29 @@ def _do_validation(schema, kwargs):
...
+    except ValueError:
+        final_kwargs = {}
+
+    for k in kwargs.keys():
+        if k in final_kwargs:
+            raise validation_error
```
henn: Maybe return a helpful message, like "URL argument passed in body of request".

haas/rest.py
zenhack: Like I said, if it's confusing to developers that's probably reason enough; I'll remove it.

Good call on the name restrictions.

haas/rest.py
zenhack: The weirdness around that is that the names of arguments in the path are implementation details; they're not visible in on the wire and `rest_api.md` doesn't necessarily refer to them by name.

Maybe log something at debug level?

henn:
2 others comments:

1) Would you mind adding something to the developer docs about API calls/validations?

2) Do we have a test case that contains only URL arguments, and not body? Are we testing the positive and negative cases for the 4 classes of arguments (none, url-only, body-only and mixed)? I started to work on some of this in [my other changes](https://github.com/CCI-MOC/haas/pull/581/files#diff-879164f6245b8b18f1dbb16ac4874c14). Feel free to copy & paste if you like any of it!

haas/rest.py
zenhack: Good call, done.

haas/rest.py
henn: Maybe `invalid argument passed in body` ?

zenhack:
@henn, dealt with the testing stuff. The docstrings in `haas.rest` are quite detailed; I'm inclined to just suggest that developers should be in the habit of reading the docs for functions they don't understand. Now that the schema stuff is actually visible at the call site, that strikes me as a bit more reasonable. If you have specific recommendations for things that aren't documented but should be, I'm all ears.

haas/rest.py
zenhack: Conceptually, it should be the same error message as if they just arbitrarily added an argument "fhqwhgads" the body -- they've specified an argument that *doesn't exist*. The name we give to the variable in the python code isn't part of the API; reporting it differently is just leaking implementation detail for no reason. If you think it would be helpful for developers, we can log it (at debug level), but I don't think anything else makes sense.

tests/unit/rest.py
```diff
@@ -231,74 +229,165 @@ def test_none_return(self):
...
+    we want to write test cases with invalid input as well.
+    """
+    client.post('/give-me-an-e', data=data)
+    # Annoyingly, the client doesn't give us a way to supply an arbitrary
+    # HTTP method, so we hack around it by using getattr to get the right
+    # function:
+    return getattr(client, method.lower())(path, data=data)
+
+
+@pytest.mark.parametrize('args', [
```
henn: I really like this parameterize!!

Would you mind also adding test cases to cover:
- [ ] duplicated args in the URL and body
- [ ] An empty body for `body_args`

haas/api.py
```diff
@@ -506,7 +535,7 @@ def headnode_detach_network(headnode, hnic):
                             ################
 
 
-@rest_call('PUT', '/network/<network>')
+@rest_call('PUT', '/network/<network>', Schema({'network': basestring}))
 def network_create(network, creator, access, net_id):
```
knikolla: Shouldn't creator, access and net_id also be part of the schema?

henn: That looks right. I've addressed it in https://github.com/zenhack/haas/pull/9/

zenhack:
@henn, merged your pr (with a few modifications), and added the tests.

henn:
Sorry - could we add one more arg test case? Other than this, I'm very happy.

Sending a valid JSON body to `/no/args` should result in an error. ie - if a user passes args to a function not expecting any, that should be an error.

zenhack:
Yeah, that doesn't work right now (nor did it before this pr). `haas.rest` automatically converts an empty body to `{}`. Changing this is actually kindof annoying, becuase, we'd need to be able to pass schema that recognize non-JSON. This isn't too bad if we're doing *just* the body, but shoving the URL arguments in there makes it not really feasable to pass a non dictionary. Could do something like `Schema({'body': ..., 'url': ...})`, but ug.

Options:

1. To heck with it, we'll allow `{}` for calls with no body arguments, like we've been doing -- it's not critical for fixing the current state of things. This is also zero work and not a regression.
2. Don't apply the schema to the URL at all. Again, not required for the basic fix, and so far we don't actually a have any calls that aren't just strings. Might make sense to check the ports, but that's switch specific, so we can't do that in the decorator anyway.
3. Do the `{'body': ..., 'url': ...}` thing.
4. Add an extra parameter for a second schema for the URL.

Opinion?

henn:
I was thinking of something more like passing a json body like `{"arg1":"foo"}` as an argument to `GET /projects`. My intuition is that the framework should correctly reject it now based on this quick test:
```python
a = schema.Schema({})
a.validate({1:2})
```


I wasn't as worried about an empty body like `{}`, though maybe we should create an issue for that?

zenhack:
Ah, I misunderstood. I'll add the test then. If you like you can open an issue.

knikolla:
+1. Looks good. We should get this in ASAP.

henn:
+1. Merging

Issue #583, opened 2016-04-29
CLI commands don't return with exit status 1 when they fail


knikolla:
Just noticed that my [ansible-haas script](https://github.com/knikolla/ansible-haas) was [still passing Travis CI](https://travis-ci.org/knikolla/ansible-haas/jobs/126498897#L440) even if the `haas init_db` command in the script section was failing (as it doesn't exist anymore). The CLI is exiting with 0 even if the command failed.

henn:
@knikolla , @gsilvis - how deep should the fix for this look for errors?

The easy fix for this exact problem (running a non-existent command) is to stick a `sys.exit(1)` before the `help()` in this:
```python
    if len(sys.argv) < 2 or sys.argv[1] not in command_dict:
        # Display usage for all commands
        help()
```

Fancier would be to stick a `sys.exit(1)` for non-success in the http status check in `check_status_code()`, though something could break if the `do_*()` call wasn't the last thing done. A quick glance makes me think that doesn't happen, though.

knikolla:
@henn I think that doesn't cover all the cases. Maybe better to put it here in the except block?

```python
def cmd(f):
    """A decorator for CLI commands.

    This decorator firstly adds the function to a dictionary of valid CLI
    commands, secondly adds exception handling for when the user passes the
    wrong number of arguments, and thirdly generates a 'usage' description and
    puts it in the usage dictionary.
    """
    @wraps(f)
    def wrapped(*args, **kwargs):
        try:
            f(*args, **kwargs)
        except TypeError:
            # TODO TypeError is probably too broad here.
            sys.stderr.write('Invalid arguements.  Usage:\n')
            help(f.__name__)
    command_dict[f.__name__] = wrapped
```

zenhack:
Closing, since we just merged a fix.

PR #597, opened 2016-05-18
Integration testing

knikolla:
This pull request configures apache2 on travis-ci and adds cli tests for project create/list/delete.


knikolla:
@zenhack need some help with this. apache/mod_wsgi doesn't seem to like psycopg2 at all. Unit tests run fine though.

```
[Sat Jun 04 17:40:45.583256 2016] [:error] [pid 4897:tid 140549225445120]     import psycopg2
[Sat Jun 04 17:40:45.583258 2016] [:error] [pid 4897:tid 140549225445120]   File "/home/travis/virtualenv/python2.7.10/lib/python2.7/site-packages/psycopg2/__init__.py", line 50, in <module>
[Sat Jun 04 17:40:45.583260 2016] [:error] [pid 4897:tid 140549225445120]     from psycopg2._psycopg import BINARY, NUMBER, STRING, DATETIME, ROWID
[Sat Jun 04 17:40:45.583262 2016] [:error] [pid 4897:tid 140549225445120] ImportError: /home/travis/virtualenv/python2.7.10/lib/python2.7/site-packages/psycopg2/_psycopg.so: undefined symbol: PyUnicodeUCS2_AsUTF8String
```

ci/install.sh
```diff
@@ -0,0 +1,30 @@
...
+sudo cp ci/haas.cfg.$DB /etc/haas.cfg
+sudo chown travis:travis /etc/haas.cfg
+
+# Database Setup
+if [ $DB = postgres ]; then
+    psql --version
+    psql -c 'CREATE DATABASE haas_tests;' -U postgres
+    psql -c 'CREATE DATABASE haas;' -U postgres
+elif [ $DB = sqlite ]; then
+    # Workaround to create an empty sqlite db file
```
zenhack: Would you elaborate on why this is necessary? It *should* create it automatically when you do a `db.create_all()`.

zenhack:
So, not certain what's going on, but there are a few suspicious things. Most notably, you seem to be using the virtualenv for the apace stuff. Even if that *should* work, it's not what `INSTALL.rst` tells you to do, so it's probably suboptimal as an integration test.

Also, a big red flag: Travis is claiming this passed despite the integration test failures. The problem is you're not checking the exit statuses in your shell script. If you add `set -e` at the top of the file, it will bomb-out at the first-non zero exit status.

A few things I'd suggest for making this easier to debug:

* Add `set -x` at the top of the shell scripts; this will cause the shell to print out the commands as its executing them.
* invoke the `env` command near the top of the scripts; that will dump all of the environment variables, which might give us some hints.

I don't really have a working theory as to why this is failing, but try the above and lets go from there.

Let me know if you want me to do a more thorough review, also.

ci/install.sh
knikolla: When removing it `haas-admin db create` throws a few warnings but everything still seems to work fine, having it there removed them.

knikolla:
@zenhack thanks for looking into this on a nice Saturday. 

Travis is saying it passed because `run_integration_tests.sh` returns 0 (since the last command it executes is `sudo cat /etc/apache2/error.log` and that makes the script succeed). Removing the `cat` makes it correctly say the integration tests failed.

I know `INSTALL.rst` doesn't say say using virtualenv, but also travis-ci recommends [against using travis-ci](https://docs.travis-ci.com/user/languages/python/#Travis-CI-Uses-Isolated-virtualenvs). I'll give it a try with system packages and see how things go.

I did some googling around and there seem to be an option when compiling python with either UCS2 or UCS4. I have a feeling that apache is using the system python with the virtualenv packages, bringing trouble for packages with C extensions.

knikolla:
@zenhack, I made the virtualenv use system site-packages and installed psycopg2 through apt-get. That seemed to fix the issue.

ci/install.sh
zenhack: What are the warnings?

Generally speaking, when you have something like this, you should be precise about the details of the problem it's working around. Too easy to get into a situation where you're looking at code that claims it fixes *something,* and you don't know what.

tests/integration/cli.py
```diff
@@ -0,0 +1,24 @@
...
+import pytest
+import subprocess
+
+PROJECT1 = 'foo'
+
+
+class TestCLI:
+    def test_add_list_delete_projects(self):
+        output = subprocess.check_output(['haas', 'list_projects'])
+        projects = ast.literal_eval(output)
```
zenhack: Should use `json.loads` here.

tests/integration/cli.py
```diff
@@ -0,0 +1,24 @@
...
+        size = len(projects)
+
+        subprocess.check_call(['haas', 'project_create', PROJECT1])
+        output = subprocess.check_output(['haas', 'list_projects'])
+        projects = ast.literal_eval(output)
+        assert PROJECT1 in projects
+        assert len(projects) == size + 1
```
zenhack: I'd just do `assert projects == [PROJECT1]` here and `assert projects == []` below.

tests/integration/cli.py
```diff
@@ -0,0 +1,24 @@
+import ast
+import pytest
+import subprocess
+
+PROJECT1 = 'foo'
+
+
+class TestCLI:
```
zenhack: This doesn't need to be in a class; pytest will pick up top-level functions called `test_*`. Do that unless there's a reason to introduce structure via a class.

tests/integration/cli.py
```diff
@@ -0,0 +1,24 @@
...
+    def test_add_list_delete_projects(self):
```
zenhack: Would add a docstring what the test is checking?

ci/haas.cfg.sqlite
```diff
@@ -0,0 +1,26 @@
+# This configuration file is used for testing HaaS with sqlite and apache
```
zenhack: Maybe name these files something that suggests the apache bit too?

tests/integration/cli.py
knikolla: I was trying to make no assumptions on the state of the system before the test is run.

zenhack: > I was trying to make no assumptions on the state of the system before
> the test is run.

Fair enough, but please add a comment to that effect.

gsilvis:
This looks pretty good to me.  One thing I'd like to check:  is there any way we can add a few negative CLI tests?  It'd be reassuring to know that the testsuite is catching CLI errors

henn:
@gsilvis: We agreed last week that we'd focus this PR on getting the infrastructure and a future PR for doing the comprehensive testing.

.travis.yml
```diff
@@ -1,31 +1,35 @@
...
+  system_site_packages: true
+
+sudo: required
+dist: trusty
+
+cache:
+  - pip
+
+addons:
+  postgresql: "9.3"
```
henn: Does 9.3 vs 9.4 or 9.5 matter much?

Are we specifying it here because we have to specify some version?

ci/haas.cfg.apache.postgres
```diff
@@ -0,0 +1,26 @@
+# This configuration files is used for testing HaaS with postgresql and apache
+[general]
+log_level = debug
+
+[auth]
+require_authentication = False
```
henn: Why disable auth? Is that future work?

ci/haas.cfg.apache.postgres
```diff
@@ -0,0 +1,26 @@
...
+[database]
+uri = postgresql://postgres@localhost/haas
+
+[devel]
+dry_run=True
+
+[extensions]
+haas.ext.switches.mock =
+haas.ext.auth.database =
+
```
henn: Do we need to add this?
```
haas.ext.obm.mock
```

ci/haas.cfg.apache.sqlite
```diff
@@ -0,0 +1,26 @@
...
+[database]
+uri = sqlite:////home/travis/haas.db
+
+[devel]
+dry_run=True
+
+[extensions]
+haas.ext.switches.mock =
+haas.ext.auth.database =
+
```
henn: Same 2 comments as for the other.

gsilvis:
@henn: Do we know that the CI /is capable/ of rejecting a patch?  Unless we know it /can/ output a negative result, I'm not sure even have infrastructure

henn:
Good point.

@knikolla - have you tried breaking something (like return bogus output
for `list_projects`) and seen CI mark the revision as failed?

@gsilvis - would that cover your concern?

.travis.yml
knikolla: We're not using any specific feature that's available only in one of these postgresql releases, but by default travis-ci uses 9.1, while ubuntu trusty ships with 9.3. I felt a bit safer as an integration test to pair trusty with 9.3.

knikolla:
@gsilvis Once #583 is resolved we can check if the CLI returns 1 on a wrong command.

@henn Yep. The test is parsing the json and looking for the project it just inserted there, if it isn't insterted it will throw an error. I added an assertion to verify that a bogus project2 is not in the project_list.

gsilvis:
@knikolla We might try and fix that one first, and test it here?  Waiting also seems fine to me.

ci/haas.cfg.apache.postgres
henn: Any thoughts on this?

knikolla:
@henn enabled authentication. i had missed that line somehow.

matsuuran:
@knikolla @gsilvis how is this going?

matsuuran:
@gsilvis @henn @SahilTikale @zenhack 
anyone have time to review?

gsilvis:
Sorry, I forgot about this PR.  Tests look good to me.  +1

zenhack:
+1, merging.

PR #609, opened 2016-06-03
Add mock OBM driver as well as null network allocator to default dev config

henn:
This way, most devs without real hardware should be able to just grab the
config and run without adjustments.

### DIFF
--- a/HACKING.rst
+++ b/HACKING.rst
@@ -45,8 +45,8 @@ Configuring HaaS
 
 Now the ``haas`` executable should be in your path.  First, create a
 configuration file ``haas.cfg``. There are two examples for you to work from,
-``examples/haas.cfg.dev-example``, which is oriented towards development, and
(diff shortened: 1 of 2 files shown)


rahulbahal7:
+1 Reviewed the changes. Also should there be a note in the README or somewhere regarding this?

henn:
@rahulbahal7 - We have this part of the README which points to the directory:
```rst
* `examples <examples/>`_ contains examples of config files, templates for creating headnode VM images and a script to register nodes with HaaS.
```

Also, HACKING.rst points to the config file location (which I just updated as part of the PR since it was outdated).

Do you think that is sufficient, or do you think we should display the config examples more prominently?

rahulbahal7:
I missed the pointer in README and the current update in HACKING.rst also makes it prominent. Thanks!

henn:
Merging due to +1.

PR #612, opened 2016-06-07
Disable native-vlan tagging on the Brocade driver

knikolla:
When a native-vlan is added to a port, disable native-vlan tagging.
This should be equivalent to running:

no switchport trunk tag native-vlan

### DIFF
--- a/haas/ext/switches/brocade.py
+++ b/haas/ext/switches/brocade.py
@@ -214,6 +214,7 @@ class Brocade(Switch):
             vlan: vlan to set as the native vlan
         """
         self._set_mode(interface, 'trunk')
+        self._disable_native_tag(interface)
         url = self._construct_url(interface, suffix='trunk')
         payload = '<trunk><native-vlan>%s</native-vlan></trunk>' % vlan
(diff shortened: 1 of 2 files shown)


ravisantoshgudimetla:
+1, Changes working as expected.

henn:
Kristi- thanks so much for writing this!

Question - do you know what will happen if we leave the port this way and subsequently remove the native VLAN? Will everything continue to be ok?

knikolla:
@henn yes, everything will continue to be ok. If the ports gets out of HaaS, native vlan tagging can just be reenabled from the switch console as expected.

henn:
+1

If you happen to know the command to reset the port to tag the native VLAN again, it'd be good to document. If not, no worries- I'll make a todo item to follow this up.

Actually- this brings up an interesting point. We don't really talk elsewhere about what state ports will be in when they leave HaaS.

knikolla:
@henn HaaS will have disconnected all the networks so no vlans will be active, native or otherwise. Since the network administrator will have to configure the port after that anyway, I don't think it really matters. 

With 2 * +1, can this be merged?

PR #614, opened 2016-06-09
Remove user-aware code from core, add per-request IDs to logs.

zenhack:
@gsilvis, @knikolla and I talked about this on IRC. The commit message describes the problem and my solution. I came across this while working on the keystone stuff and was bothered by it enough to insist on fixing it.

Could you two review? Opinions from others are also welcome (@henn?).

### DIFF
--- a/haas/auth.py
+++ b/haas/auth.py
@@ -42,14 +42,6 @@ class AuthBackend(object):
         """
 
     @abstractmethod
-    def get_user(self):
-        """Return a string with the currently authenticated user.
-
-        Return String with username of the currently authenticated user,
-        or None if no user is authenticated.
-        """
-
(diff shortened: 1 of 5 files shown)


zenhack:
I plan to build the keystone stuff on top of this; I'm doing it now because I didn't want to add `get_user` to another backend, as I strongly object to the construct.

gsilvis:
A little surprised that flask doesn't have this functionality.  +1 though.

zenhack:
If we someday feel ambitious we could pull that class into it's own package and distributed it separately; maybe have it listed on the flask extensions index. Out of scope for the time being.

knikolla:
Looks good, +1.

Issue #595, opened 2016-05-11
Project backends


knikolla:
Currently, we have an auth backend which does authentication and user/project assignment. In the eventuality that we want to have an Identity Provider (Keystone, or something else) do authentication and authorization for us, this is not enough and we must also offload all the project related logic to it.

In the Keystone example: We would be sending API calls to it for the list of projects, creating projects, deleting projects, assigning users to projects, etc. Currently HaaS assumes that there will be a project table with all this information, and it's a very strong assumption since it enforces a foreign key constraint.

To enable the use case above that assumption must be relaxed and an abstraction for projects created.

henn:
Kristi- how do the OpenStack code bases handle projects?

For example, do nova and cinder have their own Project table that they reference from others? If so, how do they manage the life cycle of those objects?

knikolla:
Nothing besides Keystone has a 'projects' table. The other services like nova and cinder only use (and care) about the project ID. That project ID comes from the scoped token and the project_id field in tables references it.

xuhang57:
PR merged. Close this now. Please reopen if necessary

PR #615, opened 2016-06-10
Keystone auth

zenhack:
WIP. Most of the effort on this one is going to be testing; yay openstack. I'm in the midst of debugging the CI setup. Would appreciate comments on the design, in particular the direction I'm going with the tests.

I'd hoped to work on this more today, but I've been running around doing last minitue prep for something that I half-forgot was this weekend. I'm going to be busy Monday too, but wil dive back into this as soon as I can. possible I can put a bit of time in Sunday night.

//cc @gsilvis, @henn, @knikolla


zenhack:
(It will probably be easier for everyone if you guys look at this *before* I've nailed down every last thing and am just blocked on review).

setup.py
```diff
@@ -70,7 +70,7 @@ def _get_readme():
                         'importlib==1.0.3',
                         'passlib==1.6.2',
                         'pexpect==3.3',
-                        'requests==2.4.1',
+                        'requests>=2.4.1,<3.0',
```
henn: Does requests 3.0 break something? If so, a comment mentioning that could help.

zenhack: It doesn't exist yet, but see <http://semver.org/>

Maybe worth having a comment about that in general.

haas/ext/auth/keystone.py
```diff
@@ -0,0 +1,22 @@
...
+    if not cfg.has_section(__name__):
+        logger.error('No section for [%s] in haas.cfg; authentication will '
+                     'not work without this. Please add this section and try '
+                     'again.' % __name__)
+        sys.exit(1)
+    keystone_cfg = {}
+    for key in cfg.options(__name__):
+        keystone_cfg[key] = cfg.get(__name__, key)
+
+    # Great job with the API design Openstack! </sarcasm>
```
gsilvis: factories!!!!

gsilvis:
Hm.  I'm not sure tests running against a real Keystone should be considered unit tests.  What I would do for unit tests is:  fake the Keystone calls, make sure that it determines authorization correctly, for a few different kinds of calls.  Then, save the actual running against Keystone for integration tests.

zenhack:
Yeah, it's in the wrong directory. Will fix

zenhack:
@gsilvis, it's been moved. I rebased; trying to not pollute the history too much. Perhaps useful for mitigating the negative effects of that:

https://zenhack.net/2015/09/27/github-code-review-workflow.html

gsilvis:
49caef5 looks good to me

henn:
Probably already on your radar, but would it be possible to include some documentation that talks about where projects get created and roles get assigned under keystone? (ie - is it in HIL or keystone; does it require being a HIL || keystone admin, etc)

zenhack:
@henn: @gsilvis, @knikolla and I talked about this on IRC for a while. I'm not 100% sure we actually came to an agreement, but I think the plan for the time being is to match projects up by name; an admin will have to create a project in hil corresponding to the openstack project name, and then you can authenticate with keystone.

@gsilvis, @knikolla, did I forget anything important from that conversation?

zenhack:
Of course whatever gets implemented will be clearly documented.

henn:
@zenhack, @gsilvis, @knikolla - is that how it works in other OpenStack projects, like Nova or Cinder? I'd be interested in documenting somewhere (maybe here?) the pros/cons for the different possibilities.

Also - If we implemented quotas (not sure if keystone stores that or the service?), all OpenStack projects could be referenced without being created in HIL first, but just the ones with HIL quotas would be able to do anything more than query stuff.

zenhack:
We're logging the IRC channel, correct? Could we fish out the logs for that conversation, rather than having it again? I don't know where those are.

zenhack:
@kamfonik, I remember you spending a bunch of time rigging up an irc logger; can you maybe fish out some logs for us? I think it was the same day that #616 got submitted/merged.

zenhack:
Or perhaps #614; not sure.

kamfonik:
[moc.2016-06-09.log.txt](https://github.com/CCI-MOC/hil/files/328882/moc.2016-06-09.log.txt)
I think this is the one you want?  The other day had no conversation.

kamfonik:
@zenhack If that's not the right one let me know, I can zip up all the logs from June for you

zenhack:
hm, I remember a longer discussion about how to do the project mapping, but apparently it didn't happen when I thought. @gsilvis, @knikolla: do you remember when that was? @henn, you said you thought you had something in your buffer that seemed relevant, when was that?

(as an aside, might be better if we could just access this ourselves in the future...)

zenhack:
Getting the whole month's logs might just be the easy thing to do.

zenhack:
Here are the relevant bits of the conversation That @gsilvis, @knikolla and I had on IRC:

https://gist.github.com/zenhack/5ea1b615681f43c14718a2a6e8e8629e

This was interleaved with discussion about what became #614. I've removed what clearly isn't relevant, but the topics overlapped in places.

@kamfonik, thanks for digging these out.

zenhack:
@henn, if you would read over that log and then weigh in, that would be appreciated.

@gsilvis, @knikolla, let's come to a consensus on this so I can plow ahead.

Also: regretably this hasn't gone in quite as quickly as I'd hoped. Most of that is to do with me; I've been dealing with sleep debt, and other commitments which I did a poor job of planning around. A thing to be aware of is that I'm going to be unavailable (and unable to work on this) starting Tuesday of next week through July 4th. Getting this in before then is going to be tight, but we come to a consensus on these issues tomorrow, I can probably have something working and ready for review by Monday. If we don't get it in before then it will have to wait another week.

zenhack:
Folks, can we come to a consensus on this?

zenhack:
Since it's perhaps easier than rifling through the IRC log, here's a summary of the questions at hand, and proposed solutions:

1. How do we map Openstack projects to HIL projects?
    1. Proposal: map HIL project name to Openstack project name.
        * Advantages:
           * Easy to work with from a UI perspective.
        * Disadvantages:
           * Openstack project names are only unique per domain. As discussed on
              IRC, one solution would be to specify the domain that HIL uses in the
              config file (which would restrict us to only one).
    2. Proposal: map HIL project name/label to Openstack project UUID
        * Advantages:
           * Globally unique
           * Corresponds to how openstack identifies resources generally.
        * Disadvantages:
           * Identifiers are opaque; would have to separately look up UUID to get
              human-understandable notion of what project we're talking about.
2. How does HIL learn about new Openstack projects?
    1. Proposal: Administrator must manually create a project in HIL with the
        identifier that corresponds to the Openstack project in question.
        * Advantages:
           * Simple
           * More administrative control over who has access to HIL.
        * Disadvantages:
           * Requires more work for administrators
    2. Proposal: When HIL sees a valid token for a project it doesn't know about,
        it creates the project on the fly.
        * Advantages:
           * Less work for administrators
           * Less bookkeeping in the overall system
        * Disadvantages:
           * Less administrative control
           * More complex implementation
           * Arguably inconsistent with the rest of the HIL api, which is very manual.
              * Note that the general philosophy of building higher-level tools suggests
                 users who want this should just have a tool that creates both Openstack
                 and HIL projects at the same time. Users who want (1) can't as easily
                 simulate it if we choose (2).

My preferences are to map projects to UUIDs, and do manual project creation,
but I don't feel super strongly (especially about the second issue).

@gsilvis has expressed that he strongly prefers manual project creation.

The IRC thread also discussed some database schema changes re: the
existence of the project table, but if nothing else, they strike me as entirely
orthoginal to the keystone work, and I'd rather leave them out of the present
discussion. @knikolla, if you feel strongly to the contrary feel free to make a
case.

//cc @henn

zenhack:
Re: UI issues around using UUIDs for projects: we can always write tools that look up projects by name before invoking the API, so the user doesn't have to deal with it. General philosophy has been to keep the server-side fairly dumb.

zenhack:
Quoting okrieg (2016-07-13 16:29:59)

>    I like the idea of manual project creation and mapping projects to
>    UUIDs

That's my inclination as well.

>    I am less clear about the domain issue for project names.

I don't really understand how domains are used in Openstack, but the
issue for us is that `(domain, project_name)` is globally unique, but
`project_name` is not, so if we were to use project names we'd have to
account for that.

>    is there problems with the uglyness of having huge UUIds in the project
>    name of HIL, not obvious why/if this is a problem?

Besides having to supply them instead of project names on the HIL
command line, there's no problem. I don't think it's an issue.

henn:
Two things to keep in mind:

1. We will soon have Network ACLs checked in as part of #554.
This means that in the keystone case, we'll want to be able to enable project access to projects that may not have been seen by this HIL instance yet. Does anyone know if the keystone client supports something like `validate_project_name()` or `validate_project_uuid()`?

2. We'll probably want to support keystone 2 keystone federation  (K2K)  at some point.
This may not make a huge difference since K2K means that the user is not necessarily known, but the project still needs to be native to the keystone service serving the HIL instance. @gsilvis @knikolla does this sound right to you?

henn:
I'm starting to think that Kristi's #595 is really the important thing to decide here.

### Name vs UUID
I'm not sure it makes a huge difference which we go with since a HIL would likely only ever exist within a single domain. How does the keystone client identify the project to the service?

In the unlikely case that we stuck with the name identifier and had to expand beyond a single domain, then we could use `domain,project` or `keystone,domain,project` tuples to describe projects, assuming the same `keystone,domain` by default to make things simple. Because of this, it makes me think that we could likely put this change off if we wanted to.

Also, if we abstracted the project implementation like #595 proposed (as we do for OBMs and Auth handlers), I don't think this is an issue, since locally-managed (and dev instances) of HIL can use names while keystone-enabled HILs could use whatever we decide on. 
Thoughts?

### Learning about projects

The thing about creating projects on demand is that there would also need to be a cleanup on demand if we don't want the table to become littered. Not impossible, but this is where tricky ref-counting bugs start to come into play. Abstracting out the projects could make it possible to avoid creating a table entry in the first place. Also keep in mind the #554 use case in the comment above.

One possibility is that we could work on quotas now in a separate PR, so that all keystone projects (assuming a `validate_project_*()`) could be recognized by default, but only those projects to which the admin gave an allotment to would be able to do more than query.

knikolla:
@henn quick reply to your questions while I keep reading the discussion.

1. With a valid admin token you can query for a specific project by name or uuid and see if it exists. HIL in this case would have an admin account in Keystone.
2. Keystone 2 Keystone will only assign users to already existing projects in Keystone, so the projects must be known beforehand.

knikolla:
If we use project UUIDs and manually register them in a projects table in HIL we are essentially "abstracting" the project definitions to Keystone and still providing a whitelist to allow us control over which projects are "enabled" for HIL access.

A configuration option could configure HIL in whitelist, blacklist, or no-list access; making it configurable on how HIL learns about the projects. However blacklist and no-list would require a change to the schema since the database enforces foreign keys.

zenhack:
> Also, if we abstracted the project implementation like #595 proposed (as we do for OBMs 
> and Auth handlers), I don't think this is an issue, since locally-managed (and dev instances) 
> of HIL can use names while keystone-enabled HILs could use whatever we decide on.
> Thoughts?

This is already the case actually; backends can interpret the labels however they like.

I'm now actually strongly leaning towards no auto-creation; as @henn points out it brings up the problem of needing to clean up projects, and while it's simple to auto-create projects when the project itself is making the call, if the project is an argument in someone else's API call, we'd have to either punt on that case or do a lookup with keystone. Currently the auth APIs don't provide an `is_valid_project` or similiar, and we'd have to design those... That's out of scope for  what I'd budgeted for - we'd planned some version of the project name thing and that's what I gave an estimate for. UUID is a slight variation on that, but mangling things to support auto-creation during a call to network_grant_access or similiar is totally out of scope.

Also, auto-creation is inconsistent with the design choices we've made *everywhere else*, and we already have resources that exist outside of HIL -- nodes don't evaporate when you do `node_delete`.


I'm also leaning towards the UUID thing just because it's simpler, and the arguments the other way aren't that compelling to me, but I'll go with whatever others prefer here.

okrieg:
Okay, so think its unanimous, project creation is manual

I am still concerned about UUID versus project name, but, dumb question, why don't we just add a domain to HIL, i.e., we have both project name and domain as fields on a HIL project.  In non keystone usage, domain is ignored, but if you are using keystone,  you have to specify both when you create the project.  From our perspective, isn't this just another opaque field we pass around that fully qualifies the project?

zenhack:
Quoting okrieg (2016-07-14 13:51:05)

>    I am still concerned about UUID versus project name, but, dumb
>    question, why don't we just add a domain to HIL, i.e., we have both
>    project name and domain as fields on a HIL project. In non keystone
>    usage, domain is ignored, but if you are using keystone, you have to
>    specify both when you create the project. From our perspective, isn't
>    this just another opaque field we pass around that fully qualifies the
>    project?


To add a separate field to the API we'd have to change HIL core, and the
db schema dictates that project.label needs to be unique, but we could
do something like what @henn suggested and have the project lables be
"projectname,domain" or such. We'd have to check what characters are
legal in openstack project names so we choose a good separator.

I like this; it keeps things humany-friendly without even the minor
configuration overhead required for setting a domain in the server
config.

What do others think?

knikolla:
Project names are mutable, that should be taken into consideration http://docs.openstack.org/admin-guide/cli_manage_projects_users_and_roles.html

zenhack:
@knikolla, good spot. That's a pretty strong argument for UUIDs; I don't see an obvioius, simple way to catch changes in the name without storing the UUID anyway.

gsilvis:
The domain could also be specified in the config file, because 90% of the time people won't care.  (You could also do a hybrid approach: use the config file domain if it's not specified in the HIL project name)

pns005:
UUID => Universally Unique Identifier.

Perhaps, I miss it, but for folks who have concern using UUID, that are the
issue(s) ?

zenhack:
Quoting pns005 (2016-07-14 22:24:40)

> UUID => Universally Unique Identifier.
> Perhaps, I miss it, but for folks who have concern using UUID, that
> are the issue(s) ?

Were there ever any other reasons besides easy readability? Particularly
in light of having to contend with *chaiging* names that seems a weak
argument.

I think at this point the choice of UUIDs and manual creation seems
clearly the way to go. Unless anyone has objections, I'm going to go
ahead and implement that.

pns005:
Ian wrote:


*I think at this point the choice of UUIDs and manual creation seemsclearly
the way to go. Unless anyone has objections, I'm going to goahead and
implement that.*

 +1. Definitely in favor of keeping things simple but works for both HIL
and
OpenStack.

zenhack:
This is done modulo a few things:

1. Tests. Working on these.
2. Design question: How do we get the initial user into theh database? We could just add another command like the existing `create_admin_user` to bypass the API for creating the first project. Does this sound reasonable to folks? Does anyone have any better ideas?

This also has the commits from #626 in it, which should be merged first.

If folks want to start looking at this, it would probably make sense to do so.

zenhack:
So it looks like the travis failure we're getting is because of this issue:

* https://github.com/python-greenlet/greenlet/issues/25

Our CI is still running on ubuntu 12.04. Travis does support 14.04, but not with containers, so startup would be slower:

* https://docs.travis-ci.com/user/ci-environment/
* https://docs.travis-ci.com/user/trusty-ci-environment

What do folks want to do with this?

zenhack:
This is now done modulo review, and resolving the travis issue I mentioned above. The tests I've written pass on my local machine.

#626 still ought to be merged first; this shouldn't be hard.

Would appreciate people doing some review.

//cc @henn, @knikolla, @gsilvis

zenhack:
I see the integration test patch that @knikolla wrote fixed the travis issue I was having, so that's moot.

gsilvis:
Looks pretty good to me.  I'm going to look through the testing part again, because that stuff was a bit complicated, then probably +1.

gsilvis:
This looks good to me.  +1

zenhack:
I just rebased now that #626 has been merged, to avoid having that patch in both prs; there are no code changes.

Also, I noticed that this actually got rid of the extra commits that have been merged, i.e. it solves the weird issue we were having where PRs showed commits that had been merged from other prs. @henn, take note.

knikolla:
@zenhack I'll review this evening.

.travis.yml
```diff
@@ -28,8 +28,10 @@ env:
   - DB=postgres
 
 install:
-  - sh -e ci/install.sh
+  - sh -e ci/apache/install.sh
```
henn: Could we cycle through the valid auth types? Want to be sure we don't let `database` and `none` become bitrotted.

INSTALL.rst
```diff
@@ -269,11 +269,22 @@ Authentication and Authorization
 --------------------------------
 
 HaaS includes a pluggable architecture for authentication and authorization.
-HaaS ships with two authentication backends. One uses HTTP basic auth, with
-usernames and passwords stored in the haas database. The other is a "null"
-backend, which does no authentication or authorization checks. This can be
-useful for testing and experimentation but *should not* be used in production.
-You must enable exactly one auth backend.
+HaaS ships with a handful of authentication backends; you must enable exactly
+one of them. The list of incuded backends is:
```
henn: Could you:
a) list these in the same order they appear below?
b) append the module name (like `haas.ext.auth.database`) to each line?

INSTALL.rst
```diff
@@ -300,6 +311,10 @@ To enable the null backend, make sure **[extensions]** contains::
 
   haas.ext.auth.null =
 
+Keystone Backend
```
henn: It may make sense to place this somewhere before `Null`, since that one is only for dev/testing

ci/keystone/keystone.sh
```diff
@@ -0,0 +1,105 @@
...
+#   http://docs.openstack.org/developer/keystone/developing.html
+
+# tunable variables:
+
+# So we can avoid repeatedly downloading the repo when developing this script:
+keystone_repo=${keystone_repo:-https://git.openstack.org/openstack/keystone}
+
+# git commit to use. Travis uses this to test against different releases of
+# keystone:
+keystone_commit=${keystone_commit:-master}
```
henn: Should we settle on a version of keystone that we're supporting, such that something changed in master doesn't break our tests?

docs/keystone-auth.md
```diff
@@ -0,0 +1,38 @@
+An authentication backend for Openstack's Keystone is maintained in this
+source tree as `haas.ext.auth.keystone`. This document describes its
+usage.
+
```
henn: Similar to my question in `ci/keystone/keystone.sh`, would it be prudent to claim to support only certain versions of keystone?

docs/keystone-auth.md
```diff
@@ -0,0 +1,38 @@
...
+Configuring HaaS to talk to Keystone deviates in the following ways:
+
+* The paste configuration is not used; you can simply ignore the
+  sections that refer to paste.
+* The options that the Keystone documentation puts in the section
+  `[keystone_authtoken]` should instead be placed in the extension's
+  section in `haas.cfg`, i.e. `[haas.ext.auth.keystone]`.
+
+# Usage
```
henn: Could we move this section above `Configuration`? I feel like it will likely be referred to more frequently.

docs/keystone-auth.md
```diff
@@ -0,0 +1,38 @@
...
+HaaS API. The HaaS project names must correspond to the Openstack UUIDs.
+For example, an administrator may execute the command:
+
+    haas project_create 00de7c85e594473db7461cdf7367166a
+
+To grant the Openstack project with that UUID access to HaaS.
+
+The HaaS command line interface will look for the same `OS_*`
+environment variables used by the Openstack command line tools; these
+may be set by a user to authenticate when using the CLI.
```
henn: Maybe make mention that a script to do this can be downloaded from the OpenStack Web UI via Access & Security.

haas/cli.py
```diff
@@ -130,9 +175,33 @@ def setup_http_client():
...
+        os_auth_url = os.getenv('OS_AUTH_URL')
+        os_password = os.getenv('OS_PASSWORD')
+        os_username = os.getenv('OS_USERNAME')
+        os_user_domain_id = os.getenv('OS_USER_DOMAIN_ID') or 'Default'
+        os_project_name = os.getenv('OS_PROJECT_NAME')
+        os_project_domain_id = os.getenv('OS_PROJECT_DOMAIN_ID') or 'Default'
+        if None in (os_auth_url, os_username, os_password, os_project_name):
+            # XXX: This error kinda makes sense, but I (zenhack) am somewhat
+            # dissatisfied.
+            raise KeyError
```
henn: Could you make this something to the effect of:
```python
raise KeyError("Required OpenStack environmental variable not included")
```

haas/cli.py
```diff
@@ -130,9 +175,33 @@ def setup_http_client():
...
+                           username=os_username,
+                           password=os_password,
+                           project_name=os_project_name,
+                           user_domain_id=os_user_domain_id,
+                           project_domain_id=os_project_domain_id)
+        sess = session.Session(auth=auth)
+        http_client = KeystoneHTTPClient(sess)
+        return
+    except (ImportError, KeyError):
+        pass
```
henn: Ahh - this is why it doesn't matter what you raise above :-)

Not sure how I feel about not informing the user that we couldn't validly configure an http client. At the same time, we don't have a good way of figuring out which auth method the client wants to do (which could include `none`)

haas/ext/auth/keystone.py
```diff
@@ -0,0 +1,64 @@
...
+                        project_id)
+            return False
+
+        return True
+
+    def _have_project_access(self, project):
+        return project.label == request.environ['HTTP_X_PROJECT_ID']
+
+    def _have_admin(self):
+        return 'admin' in request.environ['HTTP_X_ROLES'].split(',')
```
henn: Could someone have an admin role on a project, but not have global admin?

[This page](http://docs.openstack.org/developer/keystonemiddleware/api/keystonemiddleware.auth_token.html#what-auth-token-adds-to-the-request-for-use-by-the-openstack-service) seems to indicate that `HTTP_X_IS_ADMIN_PROJECT` might be something closer to what we want, but the caveats make me a little nervous to depend on it.
@gsilvis @knikolla - thoughts on this?

henn:
All set reviewing.

My only other thought is whether you can think of a way to reduce the CI test timing? It looks like we're adding 4-5 minutes here of test goodness and some folks have raised the long test times as a concern.

Maybe splitting some of the tests into other CI systems (like drone.io or circleci.com) would help? @knikolla and @SahilTikale - I know you have strong feelings in this area.

ci/keystone/keystone.sh
knikolla: That's a good point. Maybe stable/mitaka?

docs/keystone-auth.md
knikolla: AFAIK: For the backend we are supporting whatever Keystone Middleware is supporting, and for the CLI, Identity API v3.

haas/ext/auth/keystone.py
knikolla: @henn It depends on the Keystone configuration and policy file in use. With RBAC it is now possible to write policy files which restrict and scope the global admin. See [bug 968696](https://bugs.launchpad.net/keystone/+bug/968696)

`HTTP_X_IS_ADMIN_PROJECT` requires setting the configuration values for `admin_project_name` and `admin_project_domain_name`, which default to `None`. [[Source]](https://github.com/openstack/keystone/blob/master/etc/keystone.conf.sample#L1760)

Also the new sample policy configuration for v3 makes use of this new `cloud_admin` role [[Source]](https://github.com/openstack/keystone/blob/master/etc/policy.v3cloudsample.json#L3), but the default one does not [[Source]](https://github.com/openstack/keystone/blob/master/etc/policy.json#L2).

haas/cli.py
```diff
@@ -130,9 +175,33 @@ def setup_http_client():
...
-    http_client = RequestsHTTPClient()
-
+    # Next try keystone:
+    try:
+        from keystoneauth1.identity import v3
+        from keystoneauth1 import session
+        os_auth_url = os.getenv('OS_AUTH_URL')
+        os_password = os.getenv('OS_PASSWORD')
+        os_username = os.getenv('OS_USERNAME')
+        os_user_domain_id = os.getenv('OS_USER_DOMAIN_ID') or 'Default'
```
knikolla: The `default` domain id is lowercase. Same 2 lines below.

haas/cli.py
```diff
@@ -130,9 +175,33 @@ def setup_http_client():
...
+        os_project_name = os.getenv('OS_PROJECT_NAME')
```
knikolla: Aren't we using the OpenStack project id instead of the project name?

haas/cli.py
knikolla: Should we have a HIL_AUTH_METHOD env variable also?

This way OpenStack env variables can coexist with HIL env variables without having to unset.

knikolla:
Will do another pass at the testing bits after lunch. Besides the comments everything looks great.

knikolla:
@henn I'd prefer not to overcomplicate the testing infrastructure at this point in time by splitting it into several CI systems. 

If we want more performance we can move to our own hosted Jenkins, I think.

docs/keystone-auth.md
gsilvis: Stating we use the Keystone v3 api should be sufficient.  It's up to the Keystone people to make sure they actually maintain that API.

haas/ext/auth/keystone.py
gsilvis: @henn With typical Openstack policy files, there is no such thing as a per-admin project.  At some point we could move to having our own policy file, in which case we can do whatever we want.  But for now, we should work with the default policy that most Openstack installations will use.  (And the cloud_admin stuff is for more complicated domain stuff, which we also probably shouldn't use by default)

henn:
@knikolla - CircleCI says it will give open source projects up to 4 parallel containers.
In terms of test timing, @SahilTikale is the passionate one.

zenhack:
I few ideas re: ci improvements, but most of them are solidly out of scope. Much of the increase is from the ci patch that recently went in, though this depends on it: we're using VMs instead of containers, so startup times are significant. Let's discuss ci improvements on a different issue.

docs/keystone-auth.md
zenhack: Didn't know that. Does it just give you a script with `export OS_*=...`?

ci/keystone/keystone.sh
zenhack: Good spot; I meant to bring this up and forgot. There's also the option of testing multiple versions by adding stuff to Travis's env matrix, though that will add time. I think latest stable at a minimum is a good idea. Travis also let's you tag parts of the matrix as failures acceptable, so we could keep it testing master too, and get early warning if an upcoming version is going to break something.

zenhack: I don't have terribly strong opinons on this; once folks have come to a consensus I can implement.

haas/cli.py
zenhack: I don't feel strongly; if @gsilvis and @henn  think it's a good idea I can change it. The current implementation is what I'd talked to @gsilvis about on IRC.

I think I want to keep the environment variables as `HAAS_` for now, and have us switch all the identifiers at once.

haas/ext/auth/keystone.py
henn: I'm a little confused.

What is the default role for a user who has been given their own project?

My concern is that we're giving  _global_ admin rights here for someone who happens to have an admin role on a single project. Are those synonymous with how the default policy files function?

zenhack: Yeah, the caveat in that description made it fairly clear to me that we *shouldn't* use that variable. The current solution is what @gsilvis and I talked about on IRC, and his comment describes the trade-offs well.

haas/cli.py
henn: Do we expect a HIL entry will be installed in keystone's service directory? If so, then we technically don't even need HAAS_ENDPOINT when using keystone.

If that's the case, how about:
```
if HAAS_ENDPOINT:
    try username/password (DB auth)
    use null
else:
    try keystone
```
append appropriate helpful message to the keystone failure path (`If not using keystone,  perhaps you meant to set HAAS_ENDPOINT?`).

docs/keystone-auth.md
henn: Ya. And it prompts you for your password.

ci/keystone/keystone.sh
henn: I defer to @knikolla and @gsilvis

haas/ext/auth/keystone.py
zenhack: @henn, that is my memory (and I think what @gsilvis confirmed for me on IRC), and yeah, it's a weird default policy.

.travis.yml
zenhack: It's actually still using basic auth in the apache tests. Doing the full matrix would be ideal, but I think I'd like to have that separate discussion about CI first; as you've pointed out, running times are getting bad, and I'm wanting a strategy for keeping that under control.

zenhack:
I believe I've either implemented or replied to all comments; let me know if I missed something. travis is still running at the moment, but stuff passes on my local machine.

haas/ext/auth/keystone.py
gsilvis: @henn  The default role is '_member_', and openstack has no concept of the admin of a project

ci/keystone/keystone.sh
gsilvis: I don't think it matters too much---API stability is taken very seriously.  stable/mitaka is as good as anything (though even that changes sometimes, so it's not perfect).

henn: I'm not as concerned about the API being stable as I am about the checked in keystone.conf :-)
My concern is that a random future PR could experience test failures for something outside the submitter's control.

@zenhack - could we use `stable/mitaka` here?

haas/cli.py
henn: Thoughts on this?

haas/ext/auth/keystone.py
henn: Can we document our expectations on this (that we expect _only OpenStack-wide admins_ to have an `_admin_` role on any project) in docs/keystone-auth.md?

gsilvis: I guess we can say what the ``admin`` role means, sure.  It really is incredibly fundamental Keystone knowledge, though, and I think it would be quite reasonable to expect anyone deploying Openstack in production to know it.

haas/cli.py
henn: Could we add a portion to the `docs/keystone-auth.md` file giving an example command to register the HIL service in the keystone service directory?

I suspect it'll be something like:
```bash
source openstack-creds.sh # set OpenStack vars like OS_AUTH_...
keystone service-create  ...
keystone endpoint-create ...
```

haas/ext/auth/keystone.py
henn: You can always get newbs like me :)


> I guess we can say what the ``admin`` role means, sure.  It really is incredibly fundamental Keystone knowledge, though, and I think it would be quite reasonable to expect anyone deploying Openstack in production to know it.

ci/keystone/keystone.sh
zenhack: Works for me.

knikolla:
Looks good to me, +1.

haas/cli.py
zenhack: I'd considered service directory integration to be out of scope for this pr. Still haven't heard from anyone except @knikolla on the auth method variable.

zenhack:
Added the docs re: admin role and a thing to test against mitaka.

haas/cli.py
henn: I don't think we need an explicit variable - we can pick an order (either OpenStack or DB Auth first, the other second, and `Null` third), then check for the accompanying environmental variables. Just need to be sure to document it conspicuously (like in the command help).

BTW - the [keystoneclient documentation](http://docs.openstack.org/developer/python-keystoneclient/using-sessions.html#service-discovery) makes it sound like service discovery should be an easy addition, adding a single argument to the session.get/put/post calls. Of course, looks can be deceiving :-)

haas/ext/auth/keystone.py
henn: @zenhack - what do you think about documenting this?

docs/keystone-auth.md
```diff
@@ -16,6 +16,10 @@ For example, an administrator may execute the command:
 
 To grant the Openstack project with that UUID access to HaaS.
 
+Note that the plugin recognizes any user with an `admin` role on any
+project as a HaaS administrator, similar to the default policy for core
```
henn: could you say `as a global HaaS administrator, able to administrate all projects,`

docs/keystone-auth.md
```diff
@@ -0,0 +1,50 @@
...
+must manually add Openstack projects to HaaS before they can access the
+HaaS API. The HaaS project names must correspond to the Openstack UUIDs.
+For example, an administrator may execute the command:
+
+    haas project_create 00de7c85e594473db7461cdf7367166a
+
+To grant the Openstack project with that UUID access to HaaS.
+
+Note that the plugin recognizes any user with an `admin` role on any
+project as a HaaS administrator, similar to the default policy for core
```
henn: could you say `as a global HaaS administrator, able to administrate all projects,`

henn:
re: the test failures, it looks like keystone's `tools/sample_data.sh` [changed somewhat](https://github.com/openstack/keystone/commit/f7983d4bdf4da1c21da4a37229ea8fca28072cc6#diff-680f9627c165ae326e223e93c7e571ff) between stable/mitaka and now.

Perhaps using [tag 10.0.0.0b2](https://github.com/openstack/keystone/releases/tag/10.0.0.0b2) instead of stable/mitaka would be a good workaround?

docs/keystone-auth.md
zenhack: Quoting Jason Hennessey (2016-08-09 03:16:57)

>    could you say as a global HaaS administrator, able to administrate all
>    projects,

You are aware there's only one kind of HaaS administrator? I don't think
Project-level admins exist in any piece of software we've worked with as
a team.

zenhack:
Quoting Jason Hennessey (2016-08-09 03:40:29)

>    re: the test failures, it looks like keystone's tools/sample_data.sh
>    changed somewhat between stable/mitaka and now.
>
>    Perhaps using tag 10.0.0.0b2 instead of stable/mitaka would be a
>    good workaround?

Not a bad idea. I will fuss with it and find a version that works.

docs/keystone-auth.md
henn: The clarification was for keystone folks, where you could have a project
admin vs. a global admin.

zenhack: Quoting Jason Hennessey (2016-08-10 10:58:28)

>    The clarification was for keystone folks, where you could have a
>    project admin vs. a global admin.

Per @gsilvis's earlier comment, this is not this case; Openstack does
not have this concept either. @gsilvis, have I misunderstood?

gsilvis: That just isn't true.  In both Openstack and haas, 'admin' refers to a global admin (or at least the admin of a domain), and there is no concept of a project admin.

haas/cli.py
zenhack: It's a bit trickier than that, because we'd have to refactor the way HTTPClient works, or do the lookup separately; currently it expects a fully qualified URL, I'm sure there's a solution that isn't too bad, but I'd rather hash it out as we're desiging the client library/refactoring the CLI.

henn:
One final question  - what is the error given back to the user if the keystone client credentials are incorrect in some way (username, password, inaccessible project, etc)?

henn:
@zenhack - Just wanted to check that the error was something sane.

zenhack:
@henn, sorry for the delay. I poked at it a bit, most mistakes give you a 401. You get a connection error from keystone if `OS_AUTH_URL` is wrong

zenhack:
Also, a piece of this that I'd meant to deal with at some point, but that got forgotten about: how do we bootstrap the list of projects? We could add another command like create_initial_admin or whatever it was called, but for projects. I just rememberd because as part of testing this I ended up having to create a first project by turning on the null auth backend, then switching back to the keystone backend. That's not really a workable solution though.

Shouldn't be hard to just add another similar command, if folks think that's a good approach.

henn:
@zenhack I'm going to merge this now since the code is good and we want to prevent bitrotting. Two requests:

1) If it's straightforward, could make a PR that detects a keystone-related auth error and prints a helpful message?

2) If a user comes in as an admin to a non-bootstrapped HIL (thus, having no projects), will that still allow them to do stuff as a HIL admin?

henn:
+1; merging

zenhack:
Quoting Jason Hennessey (2016-08-18 23:32:46)
>    [1]@zenhack I'm going to merge this now since the code is good and we
>    want to prevent bitrotting. Two requests:
>
>    1) If it's straightforward, could make a PR that detects a
>    keystone-related auth error and prints a helpful message?

It's not immediately obvious to me how to differentiate different forms
of 401; it does at least print out the authentication required message.
The HIL server spits out the same message for basically any auth(Z)
related failure. It might be nice to at least see who yelled at us:
keystone or HIL, but the keystone client seems to throw the same
exception in both cases. I'd have to do some digging to figure out if
there's a good way to differentiate. Maybe another thing to think about
when designing client libraries?

>    2) If a user comes in as an admin to a non-bootstrapped HIL (thus,
>    having no projects), will that still allow them to do stuff as a HIL
>    admin?

You mean as an Openstack project that has admin rights? Not unless that
project is already in the DB; HIL rejects any requests by projects it
doesn't know about. You'll get a thing in the logs to the effect that
authentication checked out, but the project wasn't in the DB.

henn:
> You mean as an Openstack project that has admin rights? Not unless that
> project is already in the DB; HIL rejects any requests by projects it
> doesn't know about. You'll get a thing in the logs to the effect that
> authentication checked out, but the project wasn't in the DB.

Hmmm... it sounds like we do have a problem then. What if we allowed an admin from any  OpenStack project to be a HIL admin, regardless of whether the project was registered? In normal OpenStack, I don't believe you need to pre-register projects anyhow - we did it more for keeping our own internal tables straight and maybe in lieu of quotas.

@knikolla @gsilvis - appreciate your thoughts too.

gsilvis:
Agree with @henn here:  admins on projects not yet registered with HIL should still be able to perform any function (unless it's a function that is scoped to the user's project).  The authorization layer shouldn't care if a project has been created.  But, if the admin tries to do something like add a node to their project, then they should get the non-auth error "Project does not exist."

zenhack:
Quoting George Silvis, III (2016-08-19 08:34:36)

>    Agree with [1]@henn here: admins on projects not yet registered with
>    HIL should still be able to perform any function (unless it's a
>    function that is scoped to the user's project). The authorization layer
>    shouldn't care if a project has been created.

I'm fine with this. I can toss up another pr with this change soonish.

> But, if the admin tries to do something like add a node to their
> project, then they should get the non-auth error "Project does not
> exist."

Yep, and this should be hard to screw up -- in the code you need a
project object to act on, so if it's not in the db the code won't be
able to fetch the project at all.

henn:
Awesome - I think we have a plan here. 
lso, once we have the ability to be admins, the bootstrapping is taken care of.

PR #616, opened 2016-06-15
setup.py: Shore up dependency version specifications

zenhack:
In an attempt to keep the keystone auth pr as small as possible, I'm submitting this self-contained change. Per the commit message, python-keystoneclient requires a version of requests that conflicts with the one that we (needlessly) specified. It was easy enough to just fix the whole thing (and document the policy) so I went ahead and did it.

@gsilvis, @knikolla, @henn please have a look.


gsilvis:
+1

knikolla:
+1, looks good to me.

PR #618, opened 2016-06-23
Cli auth refactor

zenhack:
More infrastructure stuff for keystone auth. What I'll end up doing in the cli with the keystone stuff is just adding a bit of code to `setup_http_client` to check for the `OS_*` environment variables.


zenhack:
@knikolla, @gsilvis, @henn, would appreciate a review.

gsilvis:
How are you planning on detecting that the user wants to do keystone auth?

gsilvis:
Everything else looks good---in particular, I quite like the refactoring of do_request

zenhack:
I had just been planning on checking them one at a time; if `HAAS_*` is undefined we check for `OS_*`. I'm okay with having a discriminating environment variable too; that's probably a better idea, at least longer term. I'd say do it now if not for the whole "we need to redo the client stuff anyway" thing. The nice thing about this pr as is is it's a completely internal change; no risk of breaking scripts and such. If we check `HAAS_*` and then fall back to `OS_*`, we similarly get backwards compatibility.

I don't have strong feelings. I'd like to have other reviewers weigh in and come to a consensus before bothering to make changes.

re: getting the endpoint from the service catalog, I agree, but that's enough effort to sink into the client that I *really* think it should wait for the client rewrite.

haas/cli.py
```diff
@@ -59,6 +61,33 @@ def get_usage(f):
...
+
+        * Basic auth (username & password) via the HAAS_USERNAME and
+          HAAS_PASSWORD environment variables.
+        * No authentication.
+    """
+    global http_client
+    # First try basic auth:
+    basic_username = os.getenv('HAAS_USERNAME')
+    basic_password = os.getenv('HAAS_PASSWORD')
+    if basic_username is not None and basic_password is not None:
```
henn: Could `basic_password` being `None` signify a valid username with a null password?

gsilvis:
Checking for OS_PASSWORD seems reasonable, I suppose.  In the long run, you can detect when it's openstack because you got a OS_AUTH_URL instead of a HIL_API_URL or whatever.  So I guess no explicit flag is needed.

haas/cli.py
gsilvis: Is that a legal combination right now?

henn:
+1.

@SahilTikale - we'll need to be incorporate this logic into the client lib stuff.

gsilvis:
If @henn is satisfied, then +1 from me too

henn:
We can now [squash PR commits](https://help.github.com/articles/about-pull-request-merge-squashing/) into a single one!

zenhack:
@henn, I think that's generally a bad idea. I very purposefully split this into to so it was clear that the whitespace thing was a different logical change. Not really a big deal in this instance, but at the very least, let's continue to *not* do that until we've actually discussed it and  come to a consensus as a team.

haas/cli.py
zenhack: It's the same exact logic as we had before; nothing changed here.

zenhack:
@SahilTikale, the way I designed this, the backend needs to set `http_client` to an object that has a method `request` conforming to the interface described here:

http://docs.python-requests.org/en/master/api/#requests.Session.request

The object is responsible for modifying the object however necessary for authentication purposes.

The request library's `Session` class does this, and has provisions for setting up basic auth. You don't need to do the base64 stuff yourself; it will handle that.

The keystone client library's client objects also provide the same interface, so for the keystone backend we'd just set `http_client` to that instead.

Probably for a proper client library you'd want `http_client` not to be a global variable, probably a parameter to a class's `__init__` or something, but it would be the same idea.

Does that help clairfy?

I'm going to pop in to the `#moc` irc channel; feel free to ping me there.

Issue #606, opened 2016-05-31
Need list_nodes()


henn:
We should implement a list_nodes() as `GET /node`, which lists all nodes.
We should then transition list_free_nodes() to be `GET /node` with an optional argument `free=True` or something (no admin needed for this).

matsuuran:
I'm going to take this if that's cool

SahilTikale:
PR #619 addressed this.

PR #619, opened 2016-06-25
List nodes

matsuuran:
PR for #606 

first real pull so hopefully everything is in order

### DIFF
--- a/docs/apidesc.md
+++ b/docs/apidesc.md
@@ -93,7 +93,7 @@ one exception is NICs, where the label is unique only on a per-node basis.
 
     # Query interface
 
-    list_node                  <is_free>
+    list_nodes                 <is_free>
     list_project_nodes         <project>
     list_project_networks      <project>
     list_project_headnodes     <project>
--- a/docs/apidesc.md
+++ b/docs/apidesc.md
@@ -93,7 +93,7 @@ one exception is NICs, where the label is unique only on a per-node basis.
 
     # Query interface
 
-    list_free_nodes
+    list_node                  <is_free>
     list_project_nodes         <project>
     list_project_networks      <project>
     list_project_headnodes     <project>
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -329,11 +329,12 @@ Authorization requirements:
 
 * Access to the project to which `<node>` is assigned (if any) or administrative access.
 
-### list_free_nodes
+### list_nodes
 
-`GET /free_nodes`
+`GET /nodes/<is_free>`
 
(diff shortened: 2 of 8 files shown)


shwsun:
I tested it in devhaas and it looks fine, so +1.

docs/apidesc.md
```diff
@@ -93,7 +93,7 @@ one exception is NICs, where the label is unique only on a per-node basis.
 
     # Query interface
 
-    list_free_nodes
+    list_node                  <is_free>
```
knikolla: list_nodes

haas/api.py
```diff
@@ -796,20 +796,22 @@ def port_detach_nic(switch, port):
     port.nic = None
     db.session.commit()
 
-
-@rest_call('GET', '/free_nodes', Schema({}))
-def list_free_nodes():
-    """List all nodes not in any project.
+@rest_call('GET', '/node/<is_free>', Schema({'is_free' : basestring}))
```
knikolla: Should this be `/node` or `/nodes`? The API description says `nodes` but here you are registering it to `node`.

matsuuran: I think it should be /node so if for whatever reason we wanted to grab nodes by name or whatever we could do so with /node/<name>.  I'm not terribly committed to it so if you feel strongly about it and have reason, I can change it.

docs/rest_api.md
```diff
@@ -329,11 +329,12 @@ Authorization requirements:
 
 * Access to the project to which `<node>` is assigned (if any) or administrative access.
 
-### list_free_nodes
+### list_nodes
 
-`GET /free_nodes`
+`GET /nodes/<is_free>`
```
SahilTikale: make this consistent with the API call. `node/<is_free>`

haas/api.py
knikolla: I don't feel strongly about either, but for some reason list_projects is registered to `GET /projects`. https://github.com/CCI-MOC/hil/blob/master/docs/rest_api.md#list_projects so I'd prefer to be consistent. But we should discuss about this.

SahilTikale: I checked online and people use both plural and singular forms equally. I prefer idea of a resource name without getting into "English" of its plurality. As it is, many plural words do not have well defined patterns. child - children, foot -feet, sheep -sheep. So to be consistent and avoid any such problems in the future, lets stick with the singular form of resources. Whatever api is plural should be changed as we go.

matsuuran:
is everyone cool with this now?

SahilTikale:
This works. merging and closing it.

Issue #605, opened 2016-05-31
Need list_port() + show_port()


henn:
Currently, we have list_{projects/nodes,...}. To help the admin, we also need list_port and show_port.

We should also include the port in `show_node`.

All of this should only be available for admins, since regular users don't manage ports.

shwsun:
Since I am working on #567 and show_port() is already implemented by @kylehogan . I will take list_port() as well.

henn:
Addressed by #767

PR #621, opened 2016-06-30
Revert port

shwsun:
This PR is a continue work of #567 (since Kyle wont be able to finish it)

It addresses #508 and #605 . I have tested on devhaas, unit test is missing since I need code review and advice on tests.

(Need label: waiting on reviewer, do not merge yet)

- [ ] The commits are consistent with our contribution guidelines
- [ ] You've added tests (if possible) to cover your change(s)
- [ ] All tests are passing (tests/unit, pep8)
- [ ] You've updated the documentation (if adding/changing api calls)


shwsun:
Need #554  to resolve migration problem.

shwsun:
This PR is hanging there forever just because some network ACL was merged by accident. I think it would make much sense to start from a clean slate. 

Also, I am against the idea of having too many issues addressed in a single PR(, unless they are deeply related). This will hugely populate the number of lines of code and causing more troubles later. As far as I know, even within a huge code base like Google, change lists can not have more than 1k lines of code. 

//cc @pjd-nu @henn @zenhack

zenhack:
How much has this actually diverged from #554? if it's going to be easy enough to rebase it on top of *that*, then the thing might just be to wait for that to get merged and then rebase; the network acl changes will go away.

It may well be easier to just start fresh though. Whatever makes it easiest to ship the feature.

henn:
FYI - we discussed this at the weekly HIL meeting. We agreed that we should move `show_switch` from here to its own PR.

zenhack:
So here's a "fun" measurement:

```
[isd@rook haas]$ git diff shwsun/revert_port me/acl | wc -l
9133
[isd@rook haas]$ find haas tests *.py -name '*.py' | xargs cat | wc -l
10805
```

```python
>>> 9133.0 / 10805
0.8452568255437297
```

The *diff* between this tree and the version of the network acl branch with conflicts resolved is 85% the size of the *source tree*. *maybe* we can cherry-pick some things from this, but I think we should just close and work from a fresh branch once #665  is merged.

shwsun:
@zenhack That is my concern a while ago. I will start a new PR soon.

zenhack:
Okay, I'm going to go ahead and close this.

haas/api.py
```diff
@@ -726,6 +867,89 @@ def switch_delete_port(switch, port):
...
+    port = _must_find(model.Port, port)
+    nic = model.Nic.query.filter(model.Nic.port_id == port.id).first()
+
+    if nic is None:
+        return '', 404
+
+    #delete old entry and create new oned
+    nic = _must_find(model.Nic, nic.label)
+    for a in nic.attachments:
+        db.session.query(model.NetworkAttachment)\
```
henn: Could this be simplified to `a.delete()`?

haas/api.py
```diff
@@ -726,6 +867,89 @@ def switch_delete_port(switch, port):
...
+                .filter_by(nic=a.nic, channel=a.channel)\
+                .delete()
+        db.session.add(model.NetworkingAction(nic=a.nic,
+                                      new_network=a.network,
+                                      channel=a.channel))
+        port.owner.revert(port)
```
henn: I'm thinking that the revert() has to happen before the for loop above, since we want the switch to have no VLANs attached, which the for loop above restores.

haas/ext/switches/_console.py
```diff
@@ -61,6 +61,13 @@ def disable_native(self, vlan_id):
         """
 
     @abstractmethod
+    def revert(self):
```
henn: Could we call this "reset"?

haas/api.py
```diff
@@ -726,6 +867,89 @@ def switch_delete_port(switch, port):
...
+    get_auth_backend().require_admin()
+    port = _must_find(model.Port, port)
+    nic = model.Nic.query.filter(model.Nic.port_id == port.id).first()
+
+    if nic is None:
+        return '', 404
+
+    #delete old entry and create new oned
+    nic = _must_find(model.Nic, nic.label)
+    for a in nic.attachments:
```
henn: Rather than running this logic in api.py, could we add an entry to the NetworkingAction queue, so that this operation is run like all of the other switch operations in deferred.py?

Also, if we treat this as any other network operation, we should at least check to see if there is another operation pending, then maybe fail this operation. If the queue is stuck because it's waiting for revert_port(), then maybe we can add a `force` parameter or something that does something sane.

zenhack: If the queue has a revert_port in it, this would be conceptually a noop anyway, no?

Per my comment above, the prospects for fixing the merge conflicts are pretty hopeless, so I'm starting fresh, referencing this as I go. Current implementation (which I haven't pushed yet) just raises a BlockedError like everything else, but we can discuss/rehash.

Issue #474, opened 2015-08-06
Check for port in node_connect_network()


henn:
In node_connect_network(), we currently do not verify that a NIC is connected to a port before adding the connection request to the Network Action Service queue.

According to Ian in #455:
> Started working on this, and found there was actually a test that checks for the behavior you're pointing out. Talked to @gsilvis, I think the reasoning at the time was that you should never actually have nics in the DB that aren't connected to the switch, and that this was an administrator error. The driver writes a warning to the log and just noops. I'm not sure I still think this is the right behaviour, but let's not add unrelated api changes to this patch.

I think this is pretty straightforward (even if it breaks compatibility). Anyone disagree? @gsilvis @zenhack

zenhack:
I think complaining loudly and as soon as possible is the right thing here (and usually). Should be straightforward to implement.

henn:
One tricky part of doing this is that we will want to verify that a NIC is connected to a PORT on all of the real drivers, but not on the mock or null switch drivers we use for testing.

One way around this that keeps the api.py calls driver-neutral would be to check `dev_support.py:have_dry_run`

gsilvis:
I don't see any reason to change the behavior here.

henn:
This just bit us in one of our setups -- some folks spent hours trying to debug why their connected nodes couldn't talk to each other.

If we know ahead of time that a NIC is not connected to a PORT (and it's not in dev mode), then why not do the check so that we can tell the user immediately that the operation will fail?

There currently is an error message, but it's in the network service thread output:
```
WARNING:haas.deferred:Not modifying NIC eno1; NIC is not on a port.
```

gsilvis:
...hmm, that's pretty convincing.

How about returning 200, but also including in the body "Warning---this NIC is not connected to any port, and will not provide any network connectivity.  Contact your HaaS administrator to get this fixed." or something like that?

Whether the NIC has a PORT can be checked without looking into the driver at all.  Now, there's no way to check that the port makes /sense/ without looking into the driver... but I don't know that that's doable at all.

henn:
Would there ever be a case (outside of `dry_run`) where trying to connect a NIC without a port would succeed? If not, I like returning an error right away rather than delaying it.

gsilvis:
It "succeeds" fine.  It just doesn't do anything.

shwsun:
[...]

PR #625, opened 2016-07-12
Check port

shwsun:
This PR solves #474 . Basically it does a check port action before every ``node_connect_network``. This behavior is useful cause  when using HIL in real life it is hard to debug whether a port is already attached to the node's nic.

It also did ``tests/unit/api/`` adjustments since check port is not taken into consideration before.

### DIFF
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -799,7 +799,6 @@ Authorization requirements:
 Possible errors:
 
 * 404, if no port is connected to the given nic.
-
 * 409, if the nic or port is already attached to something.
 
 ### port_detach_nic
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -333,7 +333,7 @@ Authorization requirements:
 
 `GET /node/<is_free>`
 
-Return a list of all nodes or free/available nodes. The value of `is_free` 
+Return a list of all nodes or free/available nodes. The value of `is_free`
 can be `all` to return all nodes or `free` to return free/available nodes.
 
 Response body:
@@ -798,6 +798,8 @@ Authorization requirements:
 
 Possible errors:
 
+* 404, if no port is connected to the given nic.
+
 * 409, if the nic or port is already attached to something.
 
 ### port_detach_nic
@@ -884,4 +886,3 @@ Remove a user from a project.
 Authorization requirements:
 
 * Administrative access.
-
--- a/haas/test_common.py
+++ b/haas/test_common.py
@@ -287,6 +287,35 @@ def headnode_cleanup(request):
     request.addfinalizer(undefine_headnodes)
 
 
+def additional_db():
+    """ Populated database with additional objects needed for testing.
+
+    The database setup in initial_db is required to remain static as
+    a starting point for database migrations so any changes needed for
+    testing should be made in additional_db.
+    """
+
+    initial_db()
+
+    switch = db.session.query(Switch).filter_by(label="stock_switch_0").one()
+
+    with app.app_context():
+        for node_label in ['runway_node_0', 'runway_node_1',
+                           'manhattan_node_0', 'manhattan_node_1']:
+
+            node = db.session.query(Node).filter_by(label=node_label).one()
+            nic = db.session.query(Nic).filter_by(owner=node, label='boot-nic').one()
+
+            port = Port('connected_port_0', switch)
+            port.nic = nic
+            nic.port = port
+
+            db.session.add(node)
+
+        db.session.add(switch)
+        db.session.commit()
+
+
 def initial_db():
     """Populates the database with a useful set of objects.
 
--- a/haas/api.py
+++ b/haas/api.py
@@ -284,7 +284,7 @@ def node_connect_network(node, nic, network, channel=None):
     allocator = get_network_allocator()
 
     if nic.port is None:
-        raise NotFoundError("No port is connected to this nic.")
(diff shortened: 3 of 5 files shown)


henn:
@zenhack @knikolla In order to check for connectivity of a NIC to a switch port, we had to modify the default database to include a connection to a port.

What do you think of modifying the database migration check in 4fe5706 to make this work?

zenhack:
@henn, I think you linked to the wrong commit; that's just a typo fix.

Why (and how) would you need to modify the database migration check for that?

This patch doesn't appear to touch the database at all. I am highly confused.

haas/test_common.py
```diff
@@ -388,45 +394,66 @@ def initial_db():
                             username='bob',
                             password='password',
                             type=MockSwitch.api_name)
+        db.session.add(switch)
```
shwsun: @zenhack What I did is that I changed the behavior of the ``test_common.py``, and @knikolla helped me figure out that it actually caused conflicts with ``tests/unit/migrations/falsk.sql``. 

PS: looks like doing auto code format here is really a bad idea.

zenhack:
Don't change initial_db; it needs to create the exact same set of objects in `flask.sql`. If you need more objects, add them to another fixture. See also: https://github.com/CCI-MOC/hil/pull/554/files#diff-ac219157b3b73e2fefb95fc88cec5213R294

shwsun:
@zenhack Thx, I got a clear point now.

Prepare to fix the test and fixture soon.

shwsun:
Finally my first PR is done. Besides modified the ``node_connect_network``, basically I add fixtures to ``tests/unit/api/auth.py`` and modified ``main.py``.  Now it need code review @henn 

I apologize for some messy code due to using YAPF to auto format the code.

shwsun:
@knikolla do you feel like having some time to do a code review? A large part of the code is because of the code formatting so it should not take too much time.

gsilvis:
Can you make the code formatting step be a separate pull request?

shwsun:
@gsilvis you are right and I just revert the irritating formatting stuff. Code formatting should be in the pep8 PR.

gsilvis:
Thanks, that makes it a lot easier to read

gsilvis:
Most of the changes here are adding identical lines to several different tests.  Is it possible to use a fixture or something to make it available to any test that needs it?

shwsun:
@SahilTikale @henn do you guys got a minute to take a look and do a code review?

shwsun:
Waitting for code review.
// cc: @SahilTikale @zenhack @henn

zenhack:
@shwsun, I believe @gsilvis's comment about test fixtures still needs addressing. Also, you should update `docs/rest_api.md` with the new possible error reported by node_connect_network.

shwsun:
> @shwsun, I believe @gsilvis's comment about test fixtures still needs addressing. Also, you should update docs/rest_api.md with the new possible error reported by node_connect_network.

Sorry that I missed @gsilvis 's comment. will work on test fixtures.

shwsun:
To solve the identical lines problem, I think it is possible to do a shorthand and put sth like this in it.
```
def register_switch_port_simple(switch, port):
    api.switch_register(switch,
                        type=MOCK_SWITCH_TYPE,
                        username="switch_user",
                        password="switch_pass",
                        hostname="switchname")
    api.switch_register_port(switch, port)
```

However, this kind of shorthand will reduce the readability of ``tests/unit/api/main.py``. The ``switch_register`` and ``switch_register_port`` are there because they should be registered per function. 

Anythoughts? @zenhack @gsilvis

zenhack:
@shwsun, almost all of those lines actually have the same values for `switch` and `port` as well, so there's no need to even parametrize it. As @gsilvis suggested, I think the right solution is to put it into a fixture, so you're not even calling it explicitly from the tests.

The way the tests in that file are written is something we've recognized as having been a mistake; Ideally we would move most of the rest of the object creation out of the tests and into fixtures as well (though that's out of scope for this pr).

We want to keep the content of the test functions themselves down to the property they're actualy testing; setup code goes elsewhere.

shwsun:
> @shwsun, almost all of those lines actually have the same values for switch and port as well, so there's no need to even parametrize it. As @gsilvis suggested, I think the right solution is to put it into a fixture, so you're not even calling it explicitly from the tests.
> 
> The way the tests in that file are written is something we've recognized as having been a mistake; Ideally we would move most of the rest of the object creation out of the tests and into fixtures as well (though that's out of scope for this pr).
> 
> We want to keep the content of the test functions themselves down to the property they're actualy testing; setup code goes elsewhere.

That philosophy is actually sth I agree with.

shwsun:
I wrote a small pytest fixture to cover these identical test code. Need code review.
//cc: @SahilTikale @zenhack @gsilvis

shwsun:
Needs code review. @zenhack @SahilTikale @gsilvis @henn anyone has some time?

docs/rest_api.md
```diff
@@ -798,6 +798,8 @@ Authorization requirements:
 
 Possible errors:
 
+* 404, if no port is connected to the given nic.
+
```
zenhack: There should be no blank line here.

zenhack:
I made one last very minor comment; once that's addressed I'm happy.

knikolla:
Looks good, +1.

zenhack:
+1, merging.

PR #626, opened 2016-07-20
Specify interface for http_client

zenhack:
Apparently I glossed over a (small, but important) difference between the APIs provided by the requests library and the keystone client library. See the commit message.

//cc @henn, @gsilvis, @knikolla

### DIFF
--- a/haas/cli.py
+++ b/haas/cli.py
@@ -23,6 +23,7 @@ import requests
 import sys
 import urllib
 import schema
+import abc
 
 from functools import wraps
 
@@ -31,7 +32,54 @@ usage_dict = {}
 MIN_PORT_NUMBER = 1
 MAX_PORT_NUMBER = 2**16 - 1
 
+
+class HTTPClient(object):
+    """An HTTP client.
+
+    Makes HTTP requests on behalf of the HaaS CLI. Responsible for adding
+    authentication information to the request.
+    """
+
+    __metaclass__ = abc.ABCMeta
+
+    @abc.abstractmethod
+    def request(method, url, data=None):
+        """Make an HTTP request
+
+        Makes an HTTP request on URL `url` with method `method` and request
+        body `data` (if supplied). May add authentication or other
+        backend-specific information to the request.
+
+        Parameters
+        ----------
+
+        method : str
+            The HTTP method to use, e.g. 'GET', 'PUT', 'POST'...
+        url : str
+            The URL to act on
+        data : str, optional
+            The body of the request
+
+        Returns
+        -------
+
+        requests.Response
+            The HTTP response
+        """
+
+
+class RequestsHTTPClient(requests.Session, HTTPClient):
+    """An HTTPClient which uses the requests library.
+
+    Note that this doesn't do anything over `requests.Session`; that
+    class already implements the required interface. We declare it only
+    for clarity.
+    """
+
+
+# An instance of HTTPClient, which will be used to make the request.
 http_client = None
+
 
 def cmd(f):
     """A decorator for CLI commands.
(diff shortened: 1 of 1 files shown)


gsilvis:
Hm.  I'm all right with this, I guess, but it's a little odd to rely on that coincidence.

zenhack:
It's not really a coincidence; the python-keystoneclient docs say the API was very much inspired by requests. It also has requests as a dependency, so my guess is it's just a thin (by openstack standards) wrapper around requests anyway.

What would the alternative be? We need to prescribe *some* interface for different backends to implement. Backends that don't get this "for free" may have to write a wrapper class.

gsilvis:
How do you feel about making a trivial wrapper here, just to be explicit?

zenhack:
@gsilvis, how about just adding a comment explaining it?

zenhack:
I'd say just methodically document the interface; adding a wrapper is just going to trigger "why is there a function that doesn't do anything?" type reactions.

gsilvis:
Okay, sounds good to me.

henn:
I'm happy. +1

Assume we are waiting for the documentation part of @gilsvis 's request.

zenhack:
Ack, I missed another difference: the keystone client throws an
exception on 500 responses. I'm just going to go ahead and write that
wrapper.

zenhack:
@henn, @gsilvis have another look. I changed the name of the pr accordingly, as we're no longer doing the pass-by-name thing (we'll have to add a real wrapper for keystone anyway).

haas/cli.py
```diff
@@ -31,8 +32,55 @@
...
+
+        Returns
+        -------
+
+        requests.Response
+            The HTTP response
+        """
+
+
+class RequestsHTTPClient(requests.Session, HTTPClient):
```
knikolla: I understand the reasoning behind the wrapper, however why not just use the plain `requests.Session` for the fallback case? It will still work through duck typing.

zenhack: Quoting Kristi Nikolla (2016-07-20 13:13:56)

> > +class RequestsHTTPClient(requests.Session, HTTPClient):
>
>    I understand the reasoning behind the wrapper, however why not just use
>    the plain requests.Session for the fallback case? It will still work
>    through duck typing.

I don't have a strong opinion, but it does seem like we should document
somewhere explicitly that it satisfies the interface, and this seemed as
good a place as any. If folks have a preference for not wrapping
requests.Session, I'm happy to make that change. I'll wait to hear from
@henn and @gsilvis first though.

henn:
@SahilTikale is implementing a new client library, so I think it'd be good to get his input.

haas/cli.py
henn: Either way is fine by me

henn:
In thinking about this more, this is a lot of infrastructure we are putting in for a single call in do_request. I'm still fine with the changes, but wonder if there's an easier way to do the whole thing.

zenhack:
@henn, ther'es almost no code in this patch at all; it's mostly docs, and the codethat is there is just there as a convienient way of documenting the interfaces.

The `do_*` functions are entirely superfluous at this point; `do_request` used to handle the atuh logic itself, but now `do_*` just uselessly wrap the `http_client.request`.  We could swap out all the calls to `do_get(...)` with `http_client.request('GET', ...)`, for example.

zenhack:
Can we come to a decision on this? @gsilvis, @knikolla?

pjd-nu:
It looks like this could also provide a good place to hook in the flask unit test functions that we talked about during the meeting today.

zenhack:
Quoting Peter Desnoyers (2016-07-25 20:15:48)

>    It looks like this could also provide a good place to hook in the flask
>    unit test functions that we talked about during the meeting today.

Care to explain what those are, for those of us who weren't there?

henn:
@zenhack: @SahilTikale was looking for a way to implement unit tests for his new client library.

One thought had been to use a similar mechanism to what was done in `tests/unit/rest.py`, where flask pretends to start an http server and you use its fake http client to make your REST calls.

henn:
I agree with @pjd-nu: this will provide a perfect place to hook the unit tests

zenhack:
Yeah, absolutely; just use flask's mock client. It may need an actual wrapper; iirc it doesn't actually have the `request` method, just `get`, `put`, etc. I think there's already something like the wrapper you'd need in the test suite somewhere; I remember making use of `getattr`.

zenhack:
Ping? this is still waiting on you all.

// cc @henn, @gsilvis, @knikolla, @pjd-nu

matsuuran:
@henn, @gsilvis, @knikolla, @pjd-nu

Do any of you have time to review this?

gsilvis:
Looks fine to me. +1

knikolla:
Sorry for the late review, was on vacation. Looks good to me, +1 and merging.

SahilTikale:
I was wondering why no unit tests or any tests were written for this ?

zenhack:
@SahilTikale, fair question. The CLI is a blind spot re: unit tests already, and we seem to keep not dealing with that. The refactoring will have to come with tests, and from there we'll be more strict. Also, this stuff should get folded into the client library.

Issue #337, opened 2014-11-04
haas.cfg should be loaded from more sensible location(s).


zenhack:
Right now, the wsgi script loads it from `/etc/haas.cfg`, and the haas command loads it from `./haas.cfg`. This is terrible.

matsuuran:
@zenhack are you looking for wsgi/haas to check `/etc/haas.cfg` and then to check locally or did you have something else in mind?

zenhack:
exact details are up for debate, but it should (a) be at least somewhat consistent between the two, and (b) be fairly idiomatic for unix daemons.

Loading from the current directory is nice from a development standpoint, but we really shouldn't be assuming anything about where apache executes the wsgi script from.

One common (but not universal) way of finding config files is to search for them in priority order, with something like this as the ordering:

1. File specified on the command line, via a `--config` option or the like. we don't do this now partially because the way our cli is set up would require a bit of refactoring, but we want to do that anyway.
2. relative to the current directory, e.g. `./haas.cfg`
3. relative to the users home directory, e.g. `$HOME/.haas.cfg`
4. an absolute path, e.g. `/etc/haas.cfg`

Step (3) doesn't make much sense for the server, but might be nice from the client's perspective. Step (2) makes no sense when run behind apache, and step (1) isn't even possible in that case.

shwsun:
IMO, ``~/.haas.cfg`` would not be necessary. config file should be either in ``/etc/haas.cfg`` or ``./haas.cfg``. More choice would only make it harder to debug if wrong ``haas.cfg`` is loaded.

zenhack:
I don't feel strongly about `~/.haas.cfg`. @shwsun has a point. I'll go along with whatever others think is best.

matsuuran:
for now I'll leave it out and if we decide later that it is nice to have it is easy to add in

matsuuran:
@zenhack if this is the case, because my PR doesn't implement the option to pass in as a command line argument, should we even change this?  Initially I was going to wait for the larger CLI refactoring to add the `--config` option but I could do it now if you see no issue.

zenhack:
Quoting Nicholas Matsuura (2016-08-02 13:24:05)

>    [1]@zenhack if this is the case, because my PR doesn't implement the
>    option to pass in as a command line argument, should we even change
>    this? Initially I was going to wait for the larger CLI refactoring to
>    add the --config option but I could do it now if you see no issue.

The reason it needs to wait for the refactoring is that the way the
[...]

PR #632, opened 2016-07-29
Load haas config

matsuuran:
Issue #337 

This will load the haas.cfg from different locations in this order:

1. Current directory `haas.cfg` or the filename passed into the function
2. `~/.haas.cfg`
3. `/etc/haas.cfg`

They are in reverse order in the `load()` function because each subsequent overrides the previous.  I left the filename as a pass in because it gives a little more flexibility and if we wanted to add a way to pass in a filename from the command line (as @zenhack mentioned in #337) in the future it is there.  Obviously I can take it out if anyone has any real issue.


zenhack:
The behavior of configparser is to actually *merge* the files, so it won't delete options that were there in configs it loaded first. This is probably not what we want, particularly with the way extensions work.

The `read` method returns the list of successfully loaded files, so you could also do:

```python

if cfg.read(filename) != []:
     return

```

with similar clauses for the other locations.

Also, need to update the docs re: new behavior.

shwsun:
@matsuuran @zenhack 
We have discussed this one during today's hil meeting. The issue existed a long time ago, and from our perspective, loading haas config from different locations will only benefit the developers. Right now we already have the ``HAAS_ENDPOINT`` variable and the haas.cfg location should no longer be an issue.

zenhack:
I'm fine with whatever folks want to do with this; if that just means closing it, go for it.

knikolla:
Closed as requested.

PR #635, opened 2016-08-11
fixes the show_node issue and updates the documentation

SahilTikale:
`show_node` and `list_nodes` erroneously shared the same URI.
This caused `show_node` to list all nodes.
This PR fixes it.
Updated the documentation `rest_api.md` accordingly.


zenhack:
You forgot to update the tests. Other than that this looks good.

zenhack:
Something I noticed (which isn't part of this pr, since it was already there): we're not actually checking the value of the parameter in the server-side api call; if someone does `GET /nodes/banana` they'll get the same thing as `GET /nodes/all`. We probably want to be more strict than this. I think with the schema library you could actually do something along the lines of:

    Schema({"is_free": Or("free", "all")})

shwsun:
@zenhack That is probably because ``api.py`` does not check ``all`` at all.
```
def list_nodes(is_free):
    """List all nodes or all free nodes
    Returns a JSON array of strings representing a list of nodes.
    Example:  '["node1", "node2", "node3"]'
    """
    if is_free == "free":
        nodes = model.Node.query.filter_by(project_id=None).all()
    else:
        nodes = model.Node.query.all()
```

@SahilTikale since this pr is working against ``list_nodes`` can you fix this as well? And change the doc if necessart.

SahilTikale:
I agree with @zenhack about the keywords 'all' and 'free' I attempted to change it, but later postponed thinking it needs its own PR. Will do that once this is merged. 
I checked the unit tests were passing on my system. this must be CI or stress.py will check and fix that.

SahilTikale:
@shwsun will fix what you suggested.

shwsun:
@SahilTikale I think I just tracked down what's wrong here. 

Github will compare your PR against the newest HIL main repo, and your repo is 11 commits behind the upstream. Fix it then it will be ready to ship.

zenhack:
@shwsun, the problem is actually just that he needs to update the tests to match the code; they currently expect list_free_nodes to be at the old path (look at the error log). Pulling in master is good policy anyway.

zenhack:
And yes, it's stress.py that's failing.

SahilTikale:
@zenhack @shwsun 
Fixed the errors, this is ready for merge.

tests/stress.py
```diff
@@ -10,6 +10,11 @@
 @pytest.fixture
 def configure():
     config_testsuite()
+    config_merge({'extensions': 
+            { 'haas.ext.obm.ipmi': '',
+              },
+                    })
+
```
gsilvis: This indentation is kind of wonky

zenhack: Indeed. @SahilTikale, since the pep8 patch hasn't landed yet, would you run one of the pep8 linters on this branch for us?

zenhack:
Meta: where did this "Ready to merge" label come from? If it were ready to merge, a reviewer would have merged it. "Wating on reviewer" is probably the appropriate label, no? No one has signed off on this so tagging it "Ready to merge" is presumptuous at best.

gsilvis:
+1 once it passes pep8.

(And, I have no clue where that tag came from.)

zenhack:
@SahilTikale, fix the indentation issue @gsilvis pointed out and then I'm happy.

shwsun:
Indention is fixed. 

+1

zenhack:
+1, merging.

zenhack:
Also, any objections if I just remove that label from the list of available labels? I doesn't seem to have a legitimate use case, and nothing is currently even using it -- we should just be using waiting on * instead.

PR #636, opened 2016-08-15
Define auth behavior for Project 'None'

henn:
Fall back to have_admin()

Appreciate any suggestions for ways to test. This is needed before #554.

@zenhack - would appreciate your input.

### DIFF
--- a/haas/auth.py
+++ b/haas/auth.py
@@ -75,9 +75,19 @@ class AuthBackend(object):
         ``project`` will be a ``Project`` object, *not* the name of the
         project.
 
-        Note that have_admin implies have_project_acccess.
+        Note: that have_admin implies have_project_acccess.
+
+        Note 2: If ``project`` is ``None``, then we return the result of
+        ``have_admin()``. This is to prevent undefined behavior for public
+        networks, where ``None`` in the creator or access fields are valid.
         """
-        return self._have_admin() or self._have_project_access(project)
+
+        if self._have_admin():
+            return True
+        elif project is None:
+            return False
+        else:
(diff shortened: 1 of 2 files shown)


zenhack:
I don't follow; why is this needed for #554?

This seems questionable to me; I don't see any good reason for us to be passing `None` to this at all.

henn:
The code in #554 currently has cases where it passes `None` to the auth calls for networks owned by the admin (which we represent using `owner = None`).

The alternative is to change those uses and stick an `assert project is not None` in the auth code.

Either way, we should define the behavior one way or the other in case this happens again.

zenhack:
My preference would be to check in the client code. In general, when desiging APIs I think desiging for the call site is a good idea; it won't be obvious from looking at the client code that it's not just doing something dangerous/wrong, unless you actually go read the docstring for that function.

I think adding an assert is a good idea, and I'd actually suggest `assert isinstance(project, model.Project)`; I've seen a few slip-ups (not that have been merged, to my knowledge) where people have passed in the name of the project, instead of the project model. This would cover both cases.

shwsun:
IMO, either way would be fine. If we want to use this ``None`` term as a default behavior of HIL, then probably do the ``assert`` at the beginning would be better.

However, I do thhink this Project ``None`` needs to be a little more clear in the doc.

henn:
> My preference would be to check in the client code

I think adding the assert and doing the check in the client code is probably more internally consistent, since in one case (for `owner`) `None` means _admin-only_, and in the other (for `access`) is means _everyone_.

Closing this in favor of what we discussed above.

PR #637, opened 2016-08-15
Fix our broken, lonely badge

henn:


shwsun:
I spotted it a while ago. I thought we have to fix it when we do the name swapping thing anyway.

+1

knikolla:
+1, merging.

PR #638, opened 2016-08-15
Fix copy/paste error where "import logging" wasn't brought over

henn:
This is causing an error in _ipmtool() if `status != 0`.


shwsun:
This is a bug pep8 test should spot. Thanks BMI team for spotting it.

+1

knikolla:
+1, merging.

PR #639, opened 2016-08-16
Verify that project argument is actually a project.

henn:
Per suggestion from @zenhack in #636


zenhack:
Looks good to me, +1

knikolla:
Looks good, +1, merging.

PR #640, opened 2016-08-18
Update dbinit for latest APIs

henn:
I think we just need one reviewer for this, since it's not part of the main codebase.

### DIFF
--- a/ci/apache/run_integration_tests.sh
+++ b/ci/apache/run_integration_tests.sh
@@ -11,3 +11,4 @@ cd $TRAVIS_BUILD_DIR
 
 # Test commands
 py.test tests/integration/cli.py
+python examples/dbinit.py
--- a/examples/dbinit.py
+++ b/examples/dbinit.py
@@ -19,22 +19,25 @@ It could be used in an environment similar to the one which
 
 from subprocess import check_call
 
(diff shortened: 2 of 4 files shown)


henn:
Actually, now that I am hooking dbinit.py into the CI testing, this should probably have the regular review process.

The reason for hooking it is that it would be good to have a script that devs can run to populate some objects into their HIL instance that's always ready to go.

@SahilTikale - you're welcome for adding more to the testing :-)

Also - please send any comments of cli.py or scripts/haas to #642. I included them into here because the integration tests don't do anything without the CLI commands returning an appropriate status code. I can take those out if reviewers would prefer.

zenhack:
Two things:

1. re: #642, once that's merged, if you rebase the extra diffs should go away.
2. I like the idea of running example scripts with travis where possible, but I feel like it should be called out a little more explicitly as testing the *examples*, rather than the server.

gsilvis:
Why is #642 merged into this at all?  As far as I can tell, the changes are independent

gsilvis:
A little late, but, I finally noticed why you included #642 in this.

gsilvis:
This has to be redone now...

gsilvis:
Replaced by #664

PR #642, opened 2016-08-19
Return cli errors

henn:
This will make  CI testing better and more reliable.

### DIFF
--- a/haas/cli.py
+++ b/haas/cli.py
@@ -116,7 +116,6 @@ class KeystoneHTTPClient(HTTPClient):
 # An instance of HTTPClient, which will be used to make the request.
 http_client = None
 
-
 def cmd(f):
     """A decorator for CLI commands.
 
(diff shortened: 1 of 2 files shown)


gsilvis:
I'm pretty sure you can get rid of the global by modifying ``wrapped`` to either return an error value or raise an exception.  @zenhack, thoughts?

zenhack:
The alternative is to just leave the existing call to `sys.exit` where it is, and add another one in that `except TypeError`. I think that's a little simpler.

But yeah, that global is gross and needs to be removed.

gsilvis:
Oh, good point.  Can we also get a test of this function, to run under keystone and plain auth?

haas/cli.py
```diff
@@ -233,19 +245,19 @@ def do_request(method, url, data={}):
...
-    return do_request('POST', url, data=json.dumps(data))
+    do_request('POST', url, data=json.dumps(data))
 
 def do_get(url):
-    return do_request('GET', url)
+    do_request('GET', url)
 
 def do_delete(url):
-    return do_request('DELETE', url)
+    do_request('DELETE', url)
```
shwsun: Again, we lost those ``return``s

shwsun:
I agree that the global variable is a little gross. We'd better find an another way since it only serves as a signal.

haas/cli.py
henn: That was on purpose - those functions never actually returned anything,
and nobody was checking anyway.

shwsun: I will buy it. Since I knew that theoretically python will return ``-1`` if something fails, and nothing complains while I testing it on devhaas.

henn:
FYI - I believe this should also address #583

zenhack:
Yep, we should close #583  when we merge this. There are still conflicts and outstanding review comments though.

shwsun:
@henn I would say that now is a good time to deal with this one. I will get #655 done this week.

gsilvis:
@henn wasn't getting to this, so I rewrote it myself in #660

zenhack:
Closing, since merged @gsilvis's version.

PR #628, opened 2016-07-25
add pep8 tests and fix pep8 errors

matsuuran:
This PR addresses #539.  It will allow people to make sure there code is pep8 compliant with the command `py.test --pep8 -m pep8`.  I can add this to the unit tests if people desire, so when you run `py.test test/unit` it will run automatically.

PR #643, opened 2016-08-19
Enable the passing of GET arguments from the cli

henn:
This was needed for #554

### DIFF
--- a/haas/cli.py
+++ b/haas/cli.py
@@ -233,7 +233,10 @@ def do_request(method, url, data={}):
         `url` - The url to make the request to
         `data` - the body of the request.
     """
-    return check_status_code(http_client.request(method, url, data=data))
+    if method == "GET":
+        return check_status_code(http_client.request(method, url, params=data))
+    else:
+        return check_status_code(http_client.request(method, url, data=data))
 
 def do_put(url, data={}):
     return do_request('PUT', url, data=json.dumps(data))
@@ -241,8 +244,8 @@ def do_put(url, data={}):
 def do_post(url, data={}):
     return do_request('POST', url, data=json.dumps(data))
 
-def do_get(url):
-    return do_request('GET', url)
+def do_get(url, data={}):
+    return do_request('GET', url, data=data)
 
 def do_delete(url):
     return do_request('DELETE', url)


henn:
This was needed for #554

shwsun:
I am just about to ask why those ``return`` are missing and it is fixed. That's good!  Another question is: I dont think ``request`` has this ``param``, and what will be the difference between ``data`` and ``param``?

henn:
>  Another question is: I dont think request has this param, and what will be the difference between data and param?
[param](http://docs.python-requests.org/en/master/user/quickstart/#passing-parameters-in-urls) allows you to pass GET parameters, like what you see after the question mark in `http://foo.com/bar?param1=1`.

[data](http://docs.python-requests.org/en/master/user/quickstart/#more-complicated-post-requests) passes parameters via the body. This is valid for the `PUT` and `POST` methods, but is not standard for `GET`, which uses URL parameters.

gsilvis:
Looks good to me.  +1

zenhack:
-1. This breaks keystone; the HTTPClient class doesn't expect a params argument. The implementations for basic/no authentication will accept it because they're just request.Session, but KeystoneHTTPClient should throw a type error.

The good news is I think the underlying keystone session *does* support this, and I'm fine with updating our docs and passing that argument through, but if that's the approach we take you need to actually do that.

gsilvis:
Oh, good point.  Can we also get a test of this functionality, to be run under both keystone and plain auth?

zenhack:
@henn, you've got merge conflicts now as well (we *finally* merged #628).

shwsun:
@zenhack Hi Ian, I guess I will be working on this PR for a while. 

Do you still think that we need some extra work, i.e. to let the HTTPClient class accept the param argument? If so, does it mean I will have to touch the keystone session and the design we have?

zenhack:
@shwsun, yeah, HTTPClient and the keystone subclass will need to be updated to support a `params` argument. I *believe* they underlying keystone session object we're using supports the exact same parameters, so it should just be a matter of passing the parameter through and updating the docstrings for HTTPClient, but do check the openstack docs for this.

henn:
Needs to support httpclient. Closing in favor of #655

Issue #641, opened 2016-08-19
Generalize schema helpers to encompass query parameters


zenhack:
This came up on the mailing list; @henn is working on an API call that uses query parameters as part of #554. Our rest_call decorator doesn't currently check those. We can still pull them out of `request.args`, but it might be nice to have this fit into the same validation framework we're using for everything else. Let's bounce around ideas about how to do this.

zenhack:
#644 has been merged, closing.

PR #644, opened 2016-08-19
Add support for receiving GET parameters on the server

henn:
Doing this now in order to support list_network_attachments() in #554 

Should address #641

### DIFF
--- a/haas/rest.py
+++ b/haas/rest.py
@@ -175,7 +175,8 @@ def rest_call(methods, path, schema):
 
         if isinstance(methods, list):
             # Methods can be either passed as a single string
-            # or a list.
+            # or a list. Use a separate local variable because
+            # modifying `methods` gives an "UnboundLocalError"
             meths = methods
         else:
             meths = [methods]
--- a/haas/rest.py
+++ b/haas/rest.py
@@ -110,9 +110,9 @@ def rest_call(methods, path, schema):
 
     * path - the url-path to map the function to. The format is the same as for
             flask's router (e.g. '/foo/<bar>/baz')
-    * method - the HTTP method for the api call (e.g. POST, GET...).
-               Can also pass a list of methods to support multiple (used in
-               testing)
+    * method - string representing the HTTP method (e.g. POST, GET...) for the
+               api call or a list of such strings if the api call supports
+               multiple HTTP methods.
     * schema - an instance of ``schema.Schema`` with which to
             validate the arguments to the function, whether specified in the
             URL or as JSON in the request body. The schema should expect
--- a/haas/rest.py
+++ b/haas/rest.py
@@ -110,7 +110,9 @@ def rest_call(method, path, schema):
 
     * path - the url-path to map the function to. The format is the same as for
             flask's router (e.g. '/foo/<bar>/baz')
-    * method - the HTTP method for the api call (e.g. POST, GET...)
+    * method - the HTTP method for the api call (e.g. POST, GET...).
+               Can also pass a list of methods to support multiple (used in
+               testing)
     * schema - an instance of ``schema.Schema`` with which to
             validate the arguments to the function, whether specified in the
             URL or as JSON in the request body. The schema should expect
@@ -122,6 +124,10 @@ def rest_call(method, path, schema):
             a JSON object in the body of the request. It is an error for an
             argument to appear in both the request body and the path; such
             requests will be rejected prior to invoking the schema.
+
+            Because GET query parameters have no type, ensure that any GET
+            arguments of non-string type implement Schema's "Use" to perform
+            type validation and conversion.
 
     For example, given::
 
@@ -164,10 +170,13 @@ def rest_call(method, path, schema):
           whose second is an integer (the status code).
     """
     def register(f):
+
+        methods = method if isinstance(method, list) else [method]
+
         app.add_url_rule(path,
                          f.__name__,
                          _rest_wrapper(f, schema),
-                         methods=[method])
+                         methods=methods)
         return f
     return register
 
@@ -186,18 +195,35 @@ def _do_validation(schema, kwargs):
     `ValidationError`.
     """
 
-    if flask.request.data == '':
-        # No request body
-        final_kwargs = {}
+    final_kwargs = {}
+
+    if flask.request.method == "GET":
+        # GET requests can use URL and query parameter arguments
+        if flask.request.data != '':
+            raise ValidationError("GET request made with a non-empty request"
+                                  " body")
+
+        for key,value in flask.request.args.iteritems():
+            if final_kwargs.has_key(key):
+                raise ValidationError("Parameter specified more than once")
+            elif value is None or value == '':
+                # TODO: if we want to take flags (ie - an option that has no
+                # value), change this check and add logic in the next else to
+                # take care of that. We'll also need to pick a standard value
+                # to represent a flag (like 'True', 'None' or '').
+                raise ValidationError("Empty parameter specified")
+            else:
+                final_kwargs[key] = value
     else:
-        try:
-            final_kwargs = json.loads(flask.request.data)
-        except ValueError:
-            raise ValidationError("The request body is not valid JSON")
+        # Methods other than GET use URL and body arguments
+        if flask.request.data != '':
+            try:
+                final_kwargs = json.loads(flask.request.data)
+            except ValueError:
+                raise ValidationError("The request body is not valid JSON")
 
     validation_error = ValidationError(
-        "The request body is not valid for "
-        "this request.")
+            "Request arguments are not valid for this request")
 
     for k in kwargs.keys():
         if k in final_kwargs:
--- a/haas/rest.py
+++ b/haas/rest.py
@@ -180,7 +180,6 @@ def rest_call(methods, path, schema):
         else:
             meths = [methods]
 
-
         app.add_url_rule(path,
                          f.__name__,
                          _rest_wrapper(f, schema),
@@ -211,8 +210,8 @@ def _do_validation(schema, kwargs):
             raise ValidationError("GET request made with a non-empty request"
                                   " body")
 
-        for key,value in flask.request.args.iteritems():
-            if final_kwargs.has_key(key):
+        for key, value in flask.request.args.iteritems():
+            if key in final_kwargs:
                 raise ValidationError("Parameter specified more than once")
             elif value is None or value == '':
                 # TODO: if we want to take flags (ie - an option that has no
--- a/haas/rest.py
+++ b/haas/rest.py
@@ -99,7 +99,7 @@ class ValidationError(APIError):
     """An exception indicating that the body of the request was invalid."""
 
 
-def rest_call(method, path, schema):
+def rest_call(methods, path, schema):
     """A decorator which registers an http mapping to a python api call.
 
     `rest_call` makes no modifications to the function itself, though the
@@ -120,14 +120,16 @@ def rest_call(method, path, schema):
             a function argument has a default value, it may be marked as
             optional in the schema.
 
-            Any arguments not found in the path will be assumed to be keys in
-            a JSON object in the body of the request. It is an error for an
-            argument to appear in both the request body and the path; such
+            Any arguments not found in the path will be assumed to be keys in a
+            JSON object in the body of the request for non-GET requests. For
+            GET requests, query parameters (ie - those appearing after '?') are
+            also examined. It is an error for an argument to appear in two
+            places (ie - the query parameters or body and the path); such
             requests will be rejected prior to invoking the schema.
 
-            Because GET query parameters have no type, ensure that any GET
-            arguments of non-string type implement Schema's "Use" to perform
-            type validation and conversion.
+            Because GET query parameters are initially strings, ensure that any
+            GET arguments of non-string type implement Schema's "Use" to
+            perform type validation and conversion.
 
     For example, given::
 
@@ -171,12 +173,18 @@ def rest_call(method, path, schema):
     """
     def register(f):
 
-        methods = method if isinstance(method, list) else [method]
+        if isinstance(methods, list):
+            # Methods can be either passed as a single string
+            # or a list.
+            meths = methods
+        else:
+            meths = [methods]
+
 
         app.add_url_rule(path,
                          f.__name__,
                          _rest_wrapper(f, schema),
-                         methods=methods)
+                         methods=meths)
         return f
     return register
 
@@ -198,7 +206,7 @@ def _do_validation(schema, kwargs):
     final_kwargs = {}
 
     if flask.request.method == "GET":
-        # GET requests can use URL and query parameter arguments
+        # GET requests can use path AND query parameter arguments
         if flask.request.data != '':
             raise ValidationError("GET request made with a non-empty request"
                                   " body")
@@ -215,7 +223,7 @@ def _do_validation(schema, kwargs):
             else:
                 final_kwargs[key] = value
     else:
-        # Methods other than GET use URL and body arguments
+        # Methods other than GET can use path and body arguments
         if flask.request.data != '':
             try:
                 final_kwargs = json.loads(flask.request.data)
--- a/tests/unit/rest.py
+++ b/tests/unit/rest.py
@@ -24,7 +24,7 @@ from abc import ABCMeta, abstractmethod
 import unittest
 import json
 
-from schema import Schema, Optional
+from schema import Schema, Optional, Use
 
 # We don't directly use this, but unless we import it, the coverage tool
 # complains and doesn't give us a report.
@@ -238,12 +238,12 @@ def validation_setup():
     # We have four kinds of calls we want to validate here:
 
     # 1. No arguments in the URL or body (no_args)
-    @rest.rest_call('POST', '/no/args', Schema({}))
+    @rest.rest_call(['GET','POST'], '/no/args', Schema({}))
     def no_args():
         pass
 
     # 2. Argument in the URL and not in the body (url_args)
-    @rest.rest_call('POST', '/url/args/<arg1>/<arg2>', Schema({
+    @rest.rest_call(['GET','POST'], '/url/args/<arg1>/<arg2>', Schema({
         'arg1': basestring,
         'arg2': basestring,
     }))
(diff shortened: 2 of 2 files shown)


haas/rest.py
```diff
@@ -186,18 +195,35 @@ def _do_validation(schema, kwargs):
     `ValidationError`.
     """
 
-    if flask.request.data == '':
-        # No request body
-        final_kwargs = {}
+    final_kwargs = {}
+
+    if flask.request.method == "GET":
+        # GET requests can use URL and query parameter arguments
```
zenhack: s/URL/path/

the URL is the whole business:

    https://foo.bar.com:5353/some/path/elements?baz=quux&name=bob

The path is just:

    /some/path/elements

The same goes for the comment below.

haas/rest.py
```diff
@@ -186,18 +195,35 @@ def _do_validation(schema, kwargs):
...
+
+        for key,value in flask.request.args.iteritems():
+            if final_kwargs.has_key(key):
+                raise ValidationError("Parameter specified more than once")
+            elif value is None or value == '':
+                # TODO: if we want to take flags (ie - an option that has no
+                # value), change this check and add logic in the next else to
+                # take care of that. We'll also need to pick a standard value
+                # to represent a flag (like 'True', 'None' or '').
+                raise ValidationError("Empty parameter specified")
```
zenhack: I don't see any reason to add this restriction? supplying an empty string or explicit `null` is something we can do for PUT/POST as part of a JSON object.

haas/rest.py
```diff
@@ -164,10 +170,13 @@ def foo(bar, baz, quux):
           whose second is an integer (the status code).
     """
     def register(f):
+
+        methods = method if isinstance(method, list) else [method]
```
zenhack: Suggestion:

1. rename the parameter method to methods
2. Just have:

```python
if not isinstance(methods, list):
    methods = [methods]
```

...I find python's if expressions really hard to read; they look inside-out to me.

One alternative is to just get rid of the if entirely, and go change all of the calls to this; more annoying to do in the short term, but it makes our own glue code smaller/simpler, and my preference is to push as much of the misc. web framework goop down into flask as we can. I don't feel strongly about doing this now though.

zenhack: Also, I want to point out: the presence of this change threw me for a bit, since it has nothing to do with the stated purpose of this pr. Then I remembered the context in which we decided we needed this and see why you added it. I'm fine leaving it here, but please try to be more clear about the contents of a pr in its description in the future.

haas/rest.py
```diff
@@ -110,7 +110,9 @@ def rest_call(method, path, schema):
 
     * path - the url-path to map the function to. The format is the same as for
             flask's router (e.g. '/foo/<bar>/baz')
-    * method - the HTTP method for the api call (e.g. POST, GET...)
+    * method - the HTTP method for the api call (e.g. POST, GET...).
+               Can also pass a list of methods to support multiple (used in
+               testing)
```
zenhack: Wait is this just for testing then? I'd figured you were planning on using it in #554. but perhaps I've misremembered the details of that pr. If it's just for testing, I'd prefer you leave it out and try to find a way to write the tests without it.  You can stack the decorator, which may make things easier:

```python
@rest_call(...)
@rest_call(...)
def foo():
    ...
```

haas/rest.py
```diff
@@ -123,6 +125,10 @@ def rest_call(method, path, schema):
             argument to appear in both the request body and the path; such
             requests will be rejected prior to invoking the schema.
 
+            Because GET query parameters have no type, ensure that any GET
+            arguments of non-string type implement Schema's "Use" to perform
+            type validation and conversion.
```
zenhack: Get parameters do not "have no type;" their type is string. 

I think this particular warning doesn't need to be here, but you *do* need to update the rest of the description to describe the modified behavior; the preceeding paragraph says path -> body still.

zenhack:
About to step out the door, so I'll finish up my review later.

haas/rest.py
henn: With JSON, it's not really an issue since the caller explicitly sets a value and our verification system checks its type (or potentially could even check values). So if the API call expects `null` as the "flag", that's ok.

As the comment mentions, we'd need to adopt a standard for GETs, and I figured that we can do that if the need arises.

Do you feel strongly that we should pick something now? If so, which value do you propose?

haas/rest.py
henn: If I use the argument error from the parent function, I get the following error. Any ideas?

In the meantime, I'll create a version with an intermediate variable, though I do think python inline if saves a ton of space in instances like this and is the most "pythonic" way to do this.

```python
tests/unit/rest.py:33: in <module>
    from haas.test_common import config_testsuite
venv/local/lib/python2.7/site-packages/_pytest/assertion/rewrite.py:171: in load_module
    py.builtin.exec_(co, mod.__dict__)
haas/test_common.py:19: in <module>
    from haas import api, config
venv/local/lib/python2.7/site-packages/_pytest/assertion/rewrite.py:171: in load_module
    py.builtin.exec_(co, mod.__dict__)
haas/api.py:37: in <module>
    @rest_call('GET', '/projects', Schema({}))
haas/rest.py:174: in register
    if not isinstance(methods, list):
E   UnboundLocalError: local variable 'methods' referenced before assignment
```

haas/rest.py
henn: The list option is indeed to support testing, but others might find it useful. I think it saves us a bunch of space while keeping the code very understandable and adhering to DRY.

When I tried the stacking option I got an error:
Code:
```python
    @rest.rest_call('GET', '/mixed/args/<arg1>', Schema({
        'arg1': basestring,
        'arg2': basestring,
    }))
    @rest.rest_call('PUT', '/mixed/args/<arg1>', Schema({
        'arg1': basestring,
        'arg2': basestring,
    }))
    def mixed_args(arg1, arg2):
        return json.dumps([arg1, arg2])
```
Error:
```python
            if old_func is not None and old_func != view_func:
                raise AssertionError('View function mapping is overwriting an '
>                                    'existing endpoint function: %s' % endpoint)
E               AssertionError: View function mapping is overwriting an existing endpoint function: mixed_args

venv/local/lib/python2.7/site-packages/flask/app.py:984: AssertionError
```

haas/rest.py
henn: I think taking a list from everyone would be ok too, except it would be adding characters to everyone whereas this functionality is mostly useful for testing.

There are a number of API calls that differ only in method (like show_network() and network_delete()), though the benefits of having them clearly defined probably outweigh the combining of the common code (which would most likely be auth and fetching the accompanying objects).

haas/rest.py
henn: It took a little digging to figure out the "Use" function.

Take a look at the new verbiage and let me know what you think.

henn:
Made some changes per @zenhack's feedback. Appreciate any more input!

haas/rest.py
zenhack: Ack, forgot about the naming thing. You can *almost* stack the decorators :P 

I'm fine with keeping it, since it's a step closer to what flask does anyway, and I'm a fan of not writing more of our own web framework than actually makes sense. Maybe down the line we'll put a pair of braces around everything and just pass it straight through to flask.

haas/rest.py
zenhack: I just don't see why this is a special case. The empty string is still a string, and anyway all our code is already set up to deal with input of different types. If the call expects query parameters for the GET, it *needs* to include those in it's schema (that's the whole point of this patch!) So this feels both redundant and completely arbitrary.

tests/unit/rest.py
```diff
@@ -231,51 +231,58 @@ def test_none_return(self):
...
-    # arbitrary.
+    # There's a mix of GET, PUT and POST in these; mixing it up may give us
+    # better coverage. The particulars of which PUT and POST calls get which
+    # methods are arbitrary, though GET calls have some specificity.
 
     # We have four kinds of calls we want to validate here:
 
     # 1. No arguments in the URL or body (no_args)
-    @rest.rest_call('POST', '/no/args', Schema({}))
+    @rest.rest_call(['GET','POST'], '/no/args', Schema({}))
```
zenhack: IIRC, PEP8 expects you to have a space after that comma. Want to run a linter on this before we merge it? Soon travis will do that for us, but...

tests/unit/rest.py
```diff
@@ -231,51 +231,58 @@ def test_none_return(self):
...
+    def one_optional(arg1, arg2=-42):
         return json.dumps([arg1, arg2])
 
     # Let's also make sure we're testing something with a schema that isn't
     # just basestring:
-    @rest.rest_call('PUT', '/non-string-schema', schema=Schema({
-        "the_value": int,
-    }))
+    @rest.rest_call(['GET','PUT'], '/non-string-schema',
+                    schema=Schema({"the_value": Use(int)}))
```
zenhack: You should just add an extra test for this; this modification actually causes this to not cover cases it did before. It was intended to check that e.g. `{"the_value": 42}` was accepted but `{"the_value": "42"}` was not. By trying to coalesce these you've broken that.

tests/unit/rest.py
```diff
@@ -287,15 +294,15 @@ def _do_request(client, method, path, data):
...
     """
-    client.post('/give-me-an-e', data=data)
-    # Annoyingly, the client doesn't give us a way to supply an arbitrary
-    # HTTP method, so we hack around it by using getattr to get the right
-    # function:
-    return getattr(client, method.lower())(path, data=data)
+    if method == "GET":
+        return client.get(path, query_string=query, data=data)
+    else:
+        return client.open(method=method, path=path, data=data)
```
zenhack: Didn't know about this one, good find. Questions:

1. Can we just not have the if/else and pass the (possibly empty) query string to both?
2. If not, could we at least be consistent about passing path by name vs. position?

tests/unit/rest.py
```diff
@@ -231,51 +231,58 @@ def test_none_return(self):
...
-        "the_value": int,
-    }))
+    @rest.rest_call(['GET','PUT'], '/non-string-schema',
+                    schema=Schema({"the_value": Use(int)}))
     def non_string_schema(the_value):
         return json.dumps(the_value)
 
 
-def _do_request(client, method, path, data):
+def _do_request(client, method, path, data={}, query={}):
```
zenhack: Please add a description of `query` to the docstring, with the rest of the arguments.

tests/unit/rest.py
```diff
@@ -287,15 +294,15 @@ def _do_request(client, method, path, data):
...
-    # function:
-    return getattr(client, method.lower())(path, data=data)
+    if method == "GET":
+        return client.get(path, query_string=query, data=data)
+    else:
+        return client.open(method=method, path=path, data=data)
 
 
 @pytest.mark.parametrize('args', [
-    # Legal cases:
```
zenhack: Why did you remove this line?

tests/unit/rest.py
```diff
@@ -304,20 +311,93 @@ def _do_request(client, method, path, data):
                  'data': ''},
      'expected': {'status': 200,
                   'body_json': ['hello', 'goodbye']}},
+    {'request': {'method': 'GET',
+                 'path': '/url/args/hello/goodbye'
+                 },
+     'expected': {'status': 200,
+                  'body_json': ['hello', 'goodbye']}},
+    # Should fail because arg is expected in URL
```
zenhack: s/arg/arg2/

tests/unit/rest.py
```diff
@@ -304,20 +311,93 @@ def _do_request(client, method, path, data):
...
+    {'request': {'method': 'GET',
+                 'path': '/just/args',
+                 'data': json.dumps({'arg1': 'hello',
+                                     'arg2': 'goodbye'})},
+     'expected': {'status': 400,
+                  'body_json': None}},
+    {'request': {'method': 'GET',
+                 'path': '/just/args',
+                 'query': {'arg1': '',
+                                     'arg2': 'goodbye'}},
```
zenhack: Indentation here is a bit wonky.

tests/unit/rest.py
```diff
@@ -231,51 +231,58 @@ def test_none_return(self):
...
         'arg1': basestring,
         'arg2': basestring,
     }))
     def mixed_args(arg1, arg2):
         return json.dumps([arg1, arg2])
 
-    # 4. Arguments in body and not in the URL.
-    @rest.rest_call('POST', '/body/args', Schema({
+    # 4. Arguments in body (query parameters for GET) and not in the URL.
+    @rest.rest_call(['GET','POST'], '/just/args', Schema({
```
zenhack: The path name here is confusing (particularly when looking at the call site; I keep having to look back up to remind myself what this is). How about `/no/path/args` or something?

Also, in the comment above, same note about URL vs. path terminology as before.

tests/unit/rest.py
```diff
@@ -231,51 +231,58 @@ def test_none_return(self):
...
-    def body_args(arg1, arg2):
+    def just_args(arg1, arg2):
+        return json.dumps([arg1, arg2])
+
+    # 5. One optional argument.
+    @rest.rest_call(['GET', 'POST'], '/one/optional', Schema({
+        'arg1': basestring,
+        Optional('arg2'): Use(int),
+    }))
+    def one_optional(arg1, arg2=-42):
```
zenhack: This is another one that's hard to understand from the call site. Things that aren't obvious from that perspective:

1. *which* argument is optional?
2. There's no indiciation that there are any other requirements (i.e. it's not obvious you need to pass it something that is convertable to an int).

tests/unit/rest.py
```diff
@@ -411,10 +493,12 @@ def test_validation_status(client, validation_setup, args):
     # HTTP method, so we hack around it by using getattr to get the right
     # function:
```
zenhack: I think this comment has been orphaned; it (a) probably was supposed to be in `_do_request`, but got separated from the code it was talking about somehow, and (b) is obsolete now that you've found client.open. Should probably just remove it.

haas/rest.py
```diff
@@ -110,19 +110,27 @@ def rest_call(method, path, schema):
 
     * path - the url-path to map the function to. The format is the same as for
             flask's router (e.g. '/foo/<bar>/baz')
-    * method - the HTTP method for the api call (e.g. POST, GET...)
+    * method - the HTTP method for the api call (e.g. POST, GET...).
```
gsilvis: pluralize

haas/rest.py
```diff
@@ -164,10 +172,19 @@ def foo(bar, baz, quux):
           whose second is an integer (the status code).
     """
     def register(f):
+
+        if isinstance(methods, list):
```
gsilvis: Is there any reason to keep the 'single method' option?  Why not a list with one element---checking if something is a list is not terribly "pythonic" or whatever

shwsun: I assume not changing every ``method`` into ``list`` can help us keep the rest of codebase untouched.

haas/rest.py
```diff
@@ -186,18 +203,35 @@ def _do_validation(schema, kwargs):
...
+    final_kwargs = {}
+
+    if flask.request.method == "GET":
+        # GET requests can use path AND query parameter arguments
+        if flask.request.data != '':
+            raise ValidationError("GET request made with a non-empty request"
+                                  " body")
+
+        for key,value in flask.request.args.iteritems():
+            if final_kwargs.has_key(key):
```
shwsun: I think this should be ``if key in final_kwargs:``

tests/unit/rest.py
```diff
@@ -304,20 +311,93 @@ def _do_request(client, method, path, data):
...
+                 'query': {'arg2': 'goodbye'}},
+     'expected': {'status': 200,
+                  'body_json': ['hello', 'goodbye']}},
+    # Should fail because GET doesn't take body args
+    {'request': {'method': 'GET',
+                 'path': '/mixed/args/hello',
+                 'query': {'arg2': 'goodbye'},
+                 'data': json.dumps({'arg2': 'goodbye'})},
+     'expected': {'status': 400,
+                 'body_json': None}},
```
shwsun: indentation

tests/unit/rest.py
```diff
@@ -304,20 +311,93 @@ def _do_request(client, method, path, data):
...
+    {'request': {'method': 'GET',
+                 'path': '/mixed/args/hello',
+                 'data': json.dumps({'arg2': 'goodbye'})},
+     'expected': {'status': 400,
+                 'body_json': None}},
```
shwsun: indention

tests/unit/rest.py
```diff
@@ -304,20 +311,93 @@ def _do_request(client, method, path, data):
...
+                 'path': '/just/args',
                  'data': json.dumps({'arg1': 'hello',
                                      'arg2': 'goodbye'})},
      'expected': {'status': 200,
                   'body_json': ['hello', 'goodbye']}},
 
+    {'request': {'method': 'GET',
+                 'path': '/one/optional',
+                 'query': {'arg1': 'foo',
+                                     'arg2': 'bar'}},
```
shwsun: indention

tests/unit/rest.py
```diff
@@ -304,20 +311,93 @@ def _do_request(client, method, path, data):
...
+    {'request': {'method': 'GET',
+                 'path': '/mixed/args/hello',
+                 'data': json.dumps({'arg2': 'goodbye'})},
+     'expected': {'status': 400,
+                 'body_json': None}},
+
+    {'request': {'method': 'GET',
+                 'path': '/just/args',
+                 'query': {'arg1': 'hello',
+                                     'arg2': 'goodbye'}},
```
shwsun: indention

henn:
Re: having a separate variable for `methods` in register(), the reason I did it that way is because assigning to `methods` in the way you suggest causes a compilation error (see below).

Happy to receive any suggestions.

```python
_________________________ ERROR collecting tests/unit/rest.py _________________________
tests/unit/rest.py:33: in <module>
    from haas.test_common import config_testsuite
venv/local/lib/python2.7/site-packages/_pytest/assertion/rewrite.py:171: in load_module
    py.builtin.exec_(co, mod.__dict__)
haas/test_common.py:19: in <module>
    from haas import api, config
venv/local/lib/python2.7/site-packages/_pytest/assertion/rewrite.py:171: in load_module
    py.builtin.exec_(co, mod.__dict__)
haas/api.py:37: in <module>
    @rest_call('GET', '/projects', Schema({}))
haas/rest.py:183: in register
    if not isinstance(methods, list):
E   UnboundLocalError: local variable 'methods' referenced before assignment
```

zenhack:
That's right, python's scoping is utter garbage, I'd forgotten. Carry on.

tests/unit/rest.py
```diff
@@ -276,11 +276,16 @@ def one_optional(arg1, arg2=-42):
...
-    @rest.rest_call(['GET', 'PUT'], '/non-string-schema',
-                    schema=Schema({"the_value": Use(int)}))
+    @rest.rest_call('PUT', '/non-string-schema',
+                    schema=Schema({"the_value": int}))
     def non_string_schema(the_value):
         return json.dumps(the_value)
 
+    @rest.rest_call('GET', '/non-string-schema2',
+                    schema=Schema({"the_value": Use(int)}))
+    def non_string_schema2(the_value):
```
zenhack: Can we come up with better variable names to distinguish these (and maybe paths to match)? maybe something that calls out the fact that one of them converts the value (with `Use`) and the other won't.

shwsun:
I think this one is good to go.

tests/unit/rest.py
henn: Will update. At first I was nervous about making assumptions as to what the underlying defaults were (it turns out that they use `None` as a default instead of `{}`), but then I realized that there's only one difference for combining them (add `query_string` to client.open()).

haas/rest.py
henn: Agree with @shwsun - compatibility was to avoid updating everything else as well as reducing the amount of code. This is primarily used for testing anyhow.

henn:
Made a couple of updates, so it'd be good if you guys could give the last commit a quick look.

shwsun:
I am happy with the arg naming.
+1.

zenhack:
+1, Looks good to me. Merging

PR #649, opened 2016-09-03
Add VirtualHost option that includes IPv6 localhost

henn:
Somebody was ssh-forwarding to a HIL server such that their local port was forwarded to the remote HIL server. The remote port was specified as "localhost:80".

What ended up happening was that the remote host interpreted `localhost` using IPv6, which caused the connections to fail using our default setup. Adding the IPv6 localhost to the default will prevent this from happening in the future.

### DIFF
--- a/docs/INSTALL.rst
+++ b/docs/INSTALL.rst
@@ -332,7 +332,7 @@ former is a WSGI application, which we recommend running with Apache's
   LoadModule wsgi_module modules/mod_wsgi.so
   WSGISocketPrefix run/wsgi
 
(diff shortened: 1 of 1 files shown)


knikolla:
Should we just leave it as ``*:80`` ?

henn:
Right now, we don't recommend putting the server on the Internet (which is why we do localhost).

I'd prefer to keep it that way for now, till we have some confidence (like a 1.0).

knikolla:
+1, doc change, merging.

PR #655, opened 2016-09-15
Enable query parameter in HTTPClient and KeystoneHTTPClient

shwsun:
This will address the concerns in #643 and support it. In the current version of `HTTPClient` and `KeystonHTTPClient` the `params` is not handled. 
[reference official doc](http://docs.openstack.org/developer/python-keystoneclient/api/keystoneclient.html).

//cc @zenhack @henn 

### DIFF
--- a/haas/cli.py
+++ b/haas/cli.py
@@ -43,12 +43,12 @@ class HTTPClient(object):
     __metaclass__ = abc.ABCMeta
 
     @abc.abstractmethod
-    def request(method, url, data=None):
+    def request(method, url, data=None, params=None):
         """Make an HTTP request
 
-        Makes an HTTP request on URL `url` with method `method` and request
-        body `data` (if supplied). May add authentication or other
-        backend-specific information to the request.
+        Makes an HTTP request on URL `url` with method `method`, request body
+        `data`(if supplied) and query parameter `params`(if supplied). May add
+        authentication or other backend-specific information to the request.
 
         Parameters
         ----------
@@ -59,6 +59,9 @@ class HTTPClient(object):
             The URL to act on
         data : str, optional
             The body of the request
+        params : dictionary, optional
+            The query parameter, e.g. {'key1': 'val1', 'key2': 'val2'},
+            dictionary key can't be `None`
 
         Returns
         -------
@@ -95,7 +98,7 @@ class KeystoneHTTPClient(HTTPClient):
         """
         self.session = session
 
-    def request(self, method, url, data=None):
+    def request(self, method, url, data=None, params=None):
         """Make an HTTP request using keystone for authentication.
 
         Smooths over the differences between python-keystoneclient's
@@ -110,7 +113,8 @@ class KeystoneHTTPClient(HTTPClient):
             # we expect, but the names are the same:
             return self.session.request(method=method,
                                         url=url,
-                                        data=data)
+                                        data=data,
+                                        params=params)
         except HttpError as e:
             return e.response
 
--- a/haas/cli.py
+++ b/haas/cli.py
@@ -231,34 +231,29 @@ def object_url(*args):
     return url
 
 
-def do_request(method, url, data={}):
-    """Helper function for making HTTP requests against the API.
-
-    Uses the global variable `http_client` to make the request.
-
-    Arguments:
-
-        `method` - the http method, as a string: 'GET', 'PUT', 'POST'...
-        `url` - The url to make the request to
-        `data` - the body of the request.
-    """
-    return check_status_code(http_client.request(method, url, data=data))
-
+# Helper functions for making HTTP requests against the API.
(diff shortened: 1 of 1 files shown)


haas/cli.py
```diff
@@ -43,12 +43,12 @@ class HTTPClient(object):
...
     @abc.abstractmethod
-    def request(method, url, data=None):
+    def request(method, url, data=None, params=None):
         """Make an HTTP request
 
-        Makes an HTTP request on URL `url` with method `method` and request
-        body `data` (if supplied). May add authentication or other
-        backend-specific information to the request.
+        Makes an HTTP request on URL `url` with method `method`, request body
+        `data`(if supplied) and GET parameter `params`(if supplied). May add
```
zenhack: Should call them query parameters, not GET parameters; they've got nothing to do with GET per se, it's just that you can't have a body in GET so...

This is just the bit after the question mark in the URL, e.g. `https://example.com/foo/bar?baz=1&quux=hello`

haas/cli.py
```diff
@@ -59,6 +59,8 @@ def request(method, url, data=None):
             The URL to act on
         data : str, optional
             The body of the request
+        params : str, optional
+            The GET parameter
```
zenhack: This should be a dictionary; see <http://docs.python-requests.org/en/master/user/quickstart/#passing-parameters-in-urls>

Also, s/GET parameter/query parameters/ as above.

zenhack:
I made a couple comments. Ideally we'd add some tests for this as well.

shwsun:
@zenhack Thanks for pointing out some mistakes. I agree that we should get some unit tests or sth else for it, but it seems we kinda need a discussion regarding it? We need this done to put our hands on network ACLs.

My $0.02 would be open an issue and deal with it later on. 
//cc @henn

zenhack:
I'm happy. Since CLI coverage is already a problem and this isn't making it any worse, I'm fine merging this without the tests, though @shwsun, could you at least sanity check it manually and report that youv'e done so?

Also would like @henn's thoughts on the tests.

henn:
I agree with @zenhack. If @shwsun says taht he did a quick sanity check, I'm happy.

Longer term, maybe we can switch tests/integration/keystone.py to use `KeystoneHTTPClient`; then we just need to ensure something uses a GET parameter. Thoughts?

gsilvis:
I'm also okay skipping tests here for now

zenhack:
Re: using HTTPClient objects in the tests, my thinking is that we'll end up moving that class to the client library, which would be a good time to add it.

henn:
When looking at the code, I realized we need to plumb `params` through to do_get(). Then I remembered some of the ugly things that were broken there.

To address these, I made a PR to @shwsun's branch: shwsun/hil#1

Feel free to comment there!

shwsun:
I ended up hacking keystone and integration test @zenhack wrote for a couple of hours. I managed to replace the `_do_get` api call with our `KeystoneHTTPClient.request` and it seems like the query params works fine. [Here is the travis log](https://travis-ci.org/shwsun/hil/jobs/161999735)

Something I can do in a blink is to replace the `_do_get` with our `KeystoneHTTPClient.request` api call. OR we can just merge this one and save the integration changes later. Thoughts? @henn

shwsun:
@henn I merged your PR, which solved your concern. Could you take a look and approve it?
@zenhack I did the sanity check and took advantage of the integration testing framework we have for keystone, plz take a look. Do you think it is a good idea to make the `_do_get` a wrapper around `KeystoneHTTPClient`, which I don't think we have a way to test it anyway? o/w we can just merge this one and do the testing framework thing later.

gsilvis:
@henn, have all the changes you requested been fixed?

gsilvis:
@shwsun , can you rebase this down to one commit?

zenhack:
I'm happy. @gsilvis, I can just do the squash via github. I agree in this case it's the right approach.

@shwsun, re: using KeystoneHTTPClient in the integration tests, I'd actually rather not; those tests are meant to check the server-side plugin, and I don't want to conflate that with the client side. When you do that you run the risk of having an internally consistent system that isn't actually to spec: this way we know we're working against the actually HTTP semantics we wanted, rather than just that our server works with *our* client.

When we get the client library put together, well make sure to add proper tests then. For now I'm happy with the manual checking we've done.

@henn, this is just blocked on you now I believe.

gsilvis:
Oh wow, I didn't know about that feature!  Merge, squash+merge, rebase... why no squash+rebase, though?

gsilvis:
@henn is still sick, and his changes have clearly been addressed.  I'm merging this.

zenhack:
Quoting George Silvis, III (2016-09-26 22:49:51)

>    Oh wow, I didn't know about that feature! Merge, squash+merge,
>    rebase... why no squash+rebase, though?

The feature is a recent addition.

PR #657, opened 2016-09-19
Fixed broken links

naved001:
links to Install.rst and Hacking.rst have been fixed.


shwsun:
This PR does not fix the problems. `HACKING.rst` and `UPGRADE.rst` no longer exist, and I would expect to change them to correct names.

gsilvis:
Oh wow, I didn't realize how out of date that file was

gsilvis:
In that case, the best course is probably to rewrite those sections entirely

naved001:
@shwsun 
> Things like keystone doc should also be added.

You want to add [keystone-auth.md](https://github.com/CCI-MOC/hil/blob/master/docs/keystone-auth.md) to readme.rst?

Issue #572, opened 2016-04-12
show_switch API call


gsilvis:
This needs to show details of a switch.  Since different switches have different details, this will require support from the switch backend---probably another function call in the plugin API there.

xuhang57:
So, along with #550, we also need a show_switch API call for HaaS. Not sure if anyone working on these two calls? If not, I could work on them. @SahilTikale

kylehogan:
#550 handles both of them. there is also a show_switch implementation in #567

xuhang57:
Thanks for pointing it out. @kylehogan

shwsun:
#658 solved this, closing

PR #658, opened 2016-09-19
Show switch

shwsun:
This PR addresses part of the #572 's need. Per discussion with @henn and @gsilvis , right now we have `show_switch` api call, however we still need support from the switch backend. 

One quick question is: do we want a auth test for it? It doesn't seem like we have one for `list_switches`.

### DIFF
--- a/haas/api.py
+++ b/haas/api.py
@@ -731,6 +731,30 @@ def switch_delete_port(switch, port):
     db.session.commit()
 
 
+@rest_call('GET', '/switch/<switch>', Schema({
+    'switch': basestring,
+}))
+def show_switch(switch):
+    switch = _must_find(model.Switch, switch)
+    attachments = {}
+    get_auth_backend().require_admin()
+    for port in switch.ports:
+        attachments[port.label] = local.db.query(model.NetworkAttachment). \
+                                filter(model.NetworkAttachment.Nic.port == port)
+
+    get_auth_backend().require_admin()
+    return json.dumps({
+        'switch': switch.label,
+        'ports': [{'name': port.label,
+                   'attachments': [{'nic': a.nic_id,
+                                    'network': a.network_id,
+                                    'channel': a.channel,
+                                    'node': a.nic.owner,
+                   } for a in attachments[port.label]],
+        } for port in switch.ports],
+    }, sort_keys=True)
+
+
 @rest_call('GET', '/switches', Schema({}))
 def list_switches():
     """List all switches.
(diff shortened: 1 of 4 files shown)


haas/api.py
```diff
@@ -731,6 +731,28 @@ def switch_delete_port(switch, port):
...
+    """Show details of a switch.
+
+    Returns a JSON object regrading the switch. See `docs/rest_api.md`
+    for a full description of the output.
+
+    FIXME: Ideally this api call should return all detail information about
+    this switch. right now it needs support from the switch backend.
+    """
+    switch = _must_find(model.Switch, switch)
+    get_auth_backend().require_admin()
```
gsilvis: The authorization check should be run /before/ anything that might return or throw an exception.

tests/unit/api/main.py
```diff
@@ -1343,6 +1343,23 @@ def test_list_switches(self):
...
+        assert json.loads(api.show_switch('sw0')) == {
+            'name': 'sw0',
+            'ports': [{'label': '3'}]
+        }
+
+        api.switch_register_port('sw0', 'test_port')
+        assert json.loads(api.show_switch('sw0')) == {
+            'name': 'sw0',
+            'ports': [{'label': '3'},
+                      {'label': 'test_port'}]
```
shwsun: I agree with that, however I am not that concerned. The test cases are fixed and we will obey the orders, besides I don't think there is a good solution to sort the ports without breaking the json object.

haas/api.py
shwsun: Oops, will fix it.

tests/unit/api/main.py
gsilvis: Just making a comment about this issue is fine too

haas/cli.py
```diff
@@ -679,6 +679,13 @@ def list_project_networks(project):
 
 
 @cmd
+def show_switch(switch):
+    """Display information about <network>"""
```
knikolla: Docstring references network instead of switch.

tests/unit/api/main.py
```diff
@@ -1343,6 +1343,25 @@ def test_list_switches(self):
         ]
 
 
+class Test_show_switch:
```
gsilvis: CamelCase is the standard for classes

knikolla:
@shwsun can you update the branch?

gsilvis:
Squashing the commits too might be nice

PR #660, opened 2016-09-23
Return non-zero on unsuccesful API calls

gsilvis:
Rewrite of #642.


zenhack:
Looks good to me. I see @knikolla has already approved, so I'll go ahead and merge.

PR #663, opened 2016-09-29
Throw and catch a new exception on invalid arguments

### LINKED ISSUE #662: Improve CLI
Currently, cli.py throws a `FailedAPICallException` when encountering errors, which is caught in _main()_ and causes an exit with status 1.

An error case is missed by this, where _cmd()_  can also throw an uncaught exception that doesn't necessarily lead to a `sys.exit(1)`:
```python
    def wrapped(*args, **kwargs):
        try:
            f(*args, **kwargs)
        except TypeError:
            # TODO TypeError is probably too broad here.
            sys.stderr.write('Invalid arguements.  Usage:\n')
            help(f.__name__)
   command_dict[f.__name__] = wrapped
```
I believe the right thing to do here is to also catch `TypeError` in _main()_.

@gsilvis @knikolla - appreciate your input.

gsilvis:
Fixes #662

### DIFF
--- a/haas/cli.py
+++ b/haas/cli.py
@@ -123,6 +123,10 @@ class KeystoneHTTPClient(HTTPClient):
 http_client = None
 
 
+class InvalidAPIArgumentsException(Exception):
+    pass
+
+
 def cmd(f):
     """A decorator for CLI commands.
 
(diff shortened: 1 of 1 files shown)


haas/cli.py
```diff
@@ -123,6 +123,10 @@ def request(self, method, url, data=None, params=None):
 http_client = None
 
 
+class InvalidAPIArgumentsException(Exception):
```
gsilvis: If that one should, so should the other exception there.  I'd rather keep client-side errors separate from server-side errors.

knikolla: Sure. I don't have a strong opinion.

PR #664, opened 2016-09-29
Update and test dbinit

gsilvis:
Supersedes #640

### DIFF
--- a/ci/apache/run_integration_tests.sh
+++ b/ci/apache/run_integration_tests.sh
@@ -11,3 +11,6 @@ cd $TRAVIS_BUILD_DIR
 
 # Test commands
 py.test tests/integration/cli.py
+
+# Test dbinit script
+python examples/dbinit.py
--- a/examples/dbinit.py
+++ b/examples/dbinit.py
@@ -19,10 +19,11 @@ It could be used in an environment similar to the one which
 
 from subprocess import check_call
 
-N_NODES = 10
+N_NODES = 6
 
 ipmi_user = "ADMIN_USER"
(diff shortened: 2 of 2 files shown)


gsilvis:
@henn, please review

knikolla:
@henn I don't agree having this as part of the CI.

If you want CI to test this, write a integration test that tests it and its correct output. Just running it is not enough.

henn:
Kristi - my goal with including this in CI was to ensure that a future
change to the API wouldn't break the example script accidentally; the
script isn't intended to be part of our testing per se.

Is your opinion on this strong? How do others feel?

scripts/haas
```diff
@@ -14,5 +14,8 @@
 # express or implied.  See the License for the specific language
 # governing permissions and limitations under the License.
 
+import sys
+
 from haas import cli
-cli.main()
+
+sys.exit(cli.main())
```
henn: cli.main() doesn't return a value currently.

I was thinking that maybe cli.main() should do try/except with the proper exceptions, then call sys.exit() itself.

What do you think?

gsilvis: Good point.  Right now, ``cli.main`` calls ``sys.exit`` manually on errors, but not on normal results.  So we can just revert this part of the change.  (I copied it over from your PR without thinking about it very hard.)

shwsun:
@knikolla Is your concern addressed?

gsilvis:
It would be good to actually test that the script does the right thing, but just testing that it runs at all is a start.  (Note that we had to fix it here because previous changes to HIL made it invalid---adding it to CI in this form will prevent that from happening again.)

PR #665, opened 2016-09-30
Network ACLs (take 2)

zenhack:
So I fixed the merge conflicts and got the tests to pass again. I haven't yet started to clean up the history, and while I'll take a whack at it, I think I'll hit deminishing returns before it becomes anything you might call clean. @gsilvis, were there specific offenses in the history you especially wanted addressed?


zenhack:
@gsilvis, I'm noticing that a lot of the "simple typo" commits don't actually apply cleanly on top of the commits where the typos were made, due no doubt in part to the fact that there are often 10+ commits between them. I don't know how much cleanup can realistically be done.

gsilvis:
The biggest thing to fix is to remove all the merge commits, IMO.  Alternatively, do you think it's too big to squash into one commit?

zenhack:
There's ~1000 sloc worth of churn here; I'm kindof grossed out by that being one commit.

re: removing merge commits, I don't even know where to start wth rebasing across merges; that is currently above my git-fu rank. If you specify a commit for rebase where there's an intervening merge commit, what does it even *do*?

zenhack:
Also, it's apparent to me that we need to (a) be sticklers about smaller, incremental PRs in the future, and (b) teach the interns to do some very basic rebasing.

gsilvis:
If I were doing this, and I had enough time, I would squash it all down to one commit, then manually split it into some reasonable number of individual commits (and discard all the old history)

shwsun:
IMO, we should at least not adding 69 commits because of one PR. Squashing it down to some major changes would be great. I don't have a strong opinion regarding either way though.

zenhack:
I'm having a hard time justifying the time expenditure of constructing a coherent alternate history; if others are okay with just squashing the whole thing I'm game.

gsilvis:
I'm fine with that.  I do have one request, though:  can we change 'owner' back to 'creator'?  That gets rid of a lot of extra noise in the diff.

gsilvis:
And one question:  right now we have ``pytestmark = ...`` in every single class... is there some way we can abstract that out a bit?

shwsun:
@zenhack Yeah constructing a coherent history can be annoying, as long as we can reduce the number of commits I would agree with it.

@gsilvis Yes that is the way we are doing right now, it is just for some specific classes in which we want to use more features than the universals, the `pytestmark` must be a full set of features, only declaring the extra ones won't work.

zenhack:
It looks like changing the name was a concious decision:

https://github.com/CCI-MOC/hil/pull/554#issuecomment-209618385

Though the logic changes that you mentioned were never made. Thoughts?

Also, my memory was that we were more-or-less ready to merge this one; I'd rather not introduce non-essential changes at this stage.

shwsun:
I think what @gsilvis means is that switching back to `creator` would reduce the number of lines of code changes in this PR. I would say that I prefer avoiding the open a new PR just to solve that problem solution.

zenhack:
I understood what was meant. It's not as if the word `owner` is only used in place it was changed though, so reverting it may not be totally trivial. I suspect enough has changed in the interm that I can't just revert the relevant commit either (though I haven't actually checked). I don't really want to revert a change that it seemed like everyone was fine with, after it's been tested, just to make a diff smaller.

zenhack:
I fixed the latest merge conflicts. If we're just going to squash the whole thing, we can do that via github.

shwsun:
I did a quick check on devhaas, seems working fine. However, I am not sure those questions are answered:
* Do we have unit tests for all those api/functionalities? I did a quick look and there seems not
* Per those PR that @henn started, I think we might need to have proper way to test them
* And how to squash the whole PR after all those things are done.

@zenhack what do you think?

gsilvis:
Github can squash the PR for us

zenhack:
@shwsun, I remember spending a fair amount of time discussing and reviewing test code. I haven't gone back through the whole thing, but this is basically the same patch as what's in #554, just with the merge conflicts fixed (which wasn't trivial). If there are specific areas you think could use improvement re: coverage, please point them out.

kylehogan:
Thanks getting this sorted. 
 
@shwsun, @zenhack is correct that the test coverage was reviewed in the previous request, though i know that the comments are a bit much to read

@gsilvis, regarding the change from 'creator' to 'owner' - networks no longer have to be created by the project that owns them (admin could create it to be owned by a project) so the change was made to reflect that 'creator' isn't necessarily the parameter we care about (I don't think this was explicitly stated in the original comment)

gsilvis:
@kylehogan That's actually always been true, I believe... it was always a slightly weird term.  I objected at the time, saying that it wasn't a necessary change, but I was overruled then.

gsilvis:
Is anyone reviewing this right now?  @knikolla @henn @shwsun

shwsun:
@gsilvis I will review it soon. @henn do you have time to do a code review?

docs/rest_api.md
```diff
@@ -175,6 +172,102 @@ Additionally, the `show_networks` api call may return the channel identifier
...
+			"projects": ["qproj-01", qproj-02"]
+			},
+		"netB": {
+			"network_id": "102",
+			"projects": None}
+	}
+
+Authorization requirements:
+
+Administrative access is required
```
shwsun: missed a `*`

docs/rest_api.md
```diff
@@ -175,6 +172,102 @@ Additionally, the `show_networks` api call may return the channel identifier
...
+
+#### list_network_attachments
+
+`GET /network/<network>/attachments`
+
+List all nodes that are attached to network <network>.
+
+If optional argument 'project' is supplied will only list attached nodes
+belonging to the specified project.
```
shwsun: *paraphrase:*
If optional argument 'project' is supplied, only attached nodes belonging to the specified project will be listed.

docs/rest_api.md
```diff
@@ -175,6 +172,102 @@ Additionally, the `show_networks` api call may return the channel identifier
...
+		"node2": {
+			 "nic": "nic2",
+			 "channel": "vlan/235",
+			 "project": "projectB"
+			 }
+	}
+
+Authorization requirements:
+
+Admins or the project that is the owner can list all attached nodes. Other projects can only list their own nodes.
```
shwsun: * Admins or the owner project can list all attached nodes. 
* Other projects can only list their own nodes.

docs/rest_api.md
```diff
@@ -175,6 +172,102 @@ Additionally, the `show_networks` api call may return the channel identifier
...
+
+#### network_grant_project_access
+
+`PUT /network/<network>/access/<project>`
+
+Add <project> to access list for <network>.
+
+Authorization requirements:
+
+Only admins or the network owner can grant a project access to the network.
```
shwsun: missed `*`

docs/rest_api.md
```diff
@@ -175,6 +172,102 @@ Additionally, the `show_networks` api call may return the channel identifier
...
+
+#### network_revoke_project_access
+
+`DELETE /network/<network>/access/<project>`
+
+Remove <project> from access list for <network>.
+
+Authorization requirements:
+
+Only admins, the network owner, or the project itself can revoke a project's access to the network.
```
shwsun: `*`

haas/api.py
```diff
@@ -137,6 +137,82 @@ def project_detach_node(project, node):
...
+def network_grant_project_access(project, network):
+    """Add access to <network> to <project>.
+
+    If the project or network does not exist, a NotFoundError will be raised.
+    If the project already has access to the network a DuplicateError will be
+    raised.
+    """
+    network = _must_find(model.Network, network)
+    project = _must_find(model.Project, project)
+    auth_backend = get_auth_backend()
```
shwsun: auth check should be at the top.

haas/api.py
```diff
@@ -137,6 +137,82 @@ def project_detach_node(project, node):
...
+    # <project>.
+
+    if network.access:
+        if not (auth_backend.have_admin() or
+                (network.owner is not None and
+                    auth_backend.have_project_access(network.owner)) or
+                (project in network.access and
+                    auth_backend.have_project_access(project))):
+            raise AuthorizationError("You are not authorized to remove the "
+                                     "specified project form this network.")
```
shwsun: s/form/from

haas/api.py
```diff
@@ -536,25 +612,97 @@ def headnode_detach_network(headnode, hnic):
...
+            net = {'network_id': n.network_id, 'projects': None}
+        result[n.label] = net
+
+    return json.dumps(result, sort_keys=True)
+
+
+@rest_call('GET', '/network/<network>/attachments', schema=Schema({
+    'network': basestring, Optional('project'): basestring,
+}))
+def list_network_attachments(network, project=None):
```
shwsun: Can we get rid some part of the comments or make theme shorter ?

tests/unit/api/main.py
```diff
@@ -51,12 +52,6 @@ def server_init():
 with_request_context = pytest.yield_fixture(with_request_context)
 
 
-pytestmark = pytest.mark.usefixtures('configure',
-                                     'fresh_database',
-                                     'server_init',
-                                     'with_request_context')
```
shwsun: We should keep this one and make the  test code DRY.

zenhack:
@shwsun, think I responded to everything. Just waiting on travis, but stuff passed locally.

shwsun:
@zenhack I am happy with those changes. Only two more I want to get resolved:
* There are some `creator`s left in `rest_api.md`, pls take care of them as well.
* in `network_grant_project_access` the auth check is not at the top of the code.

zenhack:
@shwsun, fixed rest_api.md. Re: the auth checks in `network_grant_project_access`, the auth check *is* at the top of the function, it's just that it's ~10 lines of code. The if/else makes use of the network and project objects, so it can't be moved before those are fetched. Am I missing something?

shwsun:
@zenhack I am not convinced why we can't do auth check before fetching `network` and `project`, I assume it wouldn't matter. And we did it for the `network_revoke_project_access` as well.

shwsun:
@gsilvis just pointed out that it is not actually doing auth check. I am happy with this PR. 
+1

PR #666, opened 2016-10-03
Remove errant duplicate switch_register in CLI

gsilvis:


shwsun:
@knikolla Could you take a quick look, I think it should be fix ASAP.
+1

Issue #231, opened 2014-09-02
Test suite should detect warning/error level logs, and fail.


zenhack:

zenhack:
I'm going to grab this one. I'll have a pr soon.

PR #668, opened 2016-10-11
Test fail on log warnings

zenhack:
Per #231

This is  build on top of #665, because:

1. It probably would have caused *more* merge conflicts otherwise, and I didn't want to deal with that.
2. There's no urgency to getting this in, so we should merge #665 and then I'll rebase.

The actually content of this pr is just one commit, so if you want to get a sense before #665 is merged, you can just look at the last commit.

### DIFF
--- a/haas/test_common.py
+++ b/haas/test_common.py
@@ -154,6 +154,39 @@ def with_request_context():
         yield
 
 
+def fail_on_log_warnings():
+    """Raise an exception if a message is logged at warning level or higher.
+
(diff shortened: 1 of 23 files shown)


zenhack:
Since this does touch `tests/deployment/` (although not in a terribly meaningful way), someone should probably run those.

zenhack:
Just rebased now that #665 is merged. Waiting on travis, but stuff passed locally (including the pep8 check).

//cc @gsilvis, @henn, @shwsun

Issue #584, opened 2016-05-02
Need to validate "network_id"


henn:
running `haas network_create foo admin '' yes` creates a public network `foo`with a VLAN-id of `yes`.

We should validate that the VLAN id is an integer

gsilvis:
One option is:  Add a method to the NetworkAllocator class that checks for validity, and call that.

gsilvis:
(Note that the null backend, in particular, uses UUIDs)

naved001:
I would like to work on this issue after we decide what type of VLAN-ID is considered valid.

naved001:
Here's a PR #670  that addresses this issue.

gsilvis:
Fixed by #670

PR #669, opened 2016-10-11
Validates "network_id"

naved001:
Added a try and except block to make sure that network_id is an int. 
So running `<haas network_create foo admin '' yes>` will return a Value Error. 
This PR is to fix issue #584


shwsun:
I would say it makes more sense to validate it from the api side.

naved001:
Yes, makes sense. I am on it.

Issue #314, opened 2014-10-27
record successful application of networks one at a time.


zenhack:
Right now `serve_networks()` clears pending actions in one transaction, at the very end of its operation. This is suboptimal, since it means a failure in the middle will cause _all_ actions to be reapplied, even ones which are complete.

The code should be reorganized a bit to facilitate this; in particular, doing them all at once allows us to avoid logging into the switch multiple times. Getting the best of both works would require a bit of refactoring.

henn:
I think this is actually a pretty simple change, and would increase the robustness of HaaS by not allowing one error in any operation to cause all operations to fail. Operations in the same queue as the failing one will also reflect incorrectly in the DB (and thus the future ``show_network()``). Optimization (like having multiple service threads or connection pooling) should be treated as a separate issue.

It could be done within deferred.py's ``apply_networking()`` by placing the bulk of the body inside of a while loop that is keyed off of:
```python
action = db.query(model.NetworkingAction).\
             order_by(model.NetworkingAction.id).first()
```

See SQLAlchemy's [first()](http://docs.sqlalchemy.org/en/rel_0_9/orm/query.html#sqlalchemy.orm.query.Query.first) function for info on semantics.

It might make sense to move to some sort of Reliable Message Queue when it's needed @zenhack, however I think for now using a DB table in a sensible way will get us a lot of the way there.

zenhack:
I'd want to actually measure the time it takes to log into and out of the switch before making this change -- It may well be simply too big not to batch the applications.

If this is the case, the solution is easy enough: expose connect/disconnect. On my vlan-trunking branch these are already separated out (just for cleanliness purposes) so it wouldn't be hard to make it a public api. then we can just put the connect/disconnect around your proposed loop.

zenhack:
Observation: the implementation is changed such that connections to the switches are created on first use, so the overhead we were discussing for not batching the actions is no longer there. doing a commit after each action seems sensible. I'd like to pin down where #508 is going, but this seems straightforward in and of itself.

matsuuran:
would it be okay to close the sessions after the db commits, so we'd go from two loops on actions to one?

zenhack:
@matsuuran, that's what I had in mind.

naved001:
@henn this is the only 0.2 bug left now

zenhack:
I'm going to take @matsuuran off of this. Does anyone else want to pick this one up?

naved001:
@zenhack @henn 

```python
[...]

Issue #54, opened 2014-04-29
Check for well formed port names on registration


zenhack:
We're making use of pexpect to ensure the state of the shell is what we think it is, but if pexpect should throw an exception, it will crash the HaaS. This shouldn't be possible; we should be catching the exception somewhere.

henn:
I've run into this, specifically when the port names are entered incorrectly. We should address this for 0.2

henn:
Gold star if you can catch my specific error and report something like "Check that port names agree with the switch"

zenhack:
@henn I think the right approach to port names would be to check them on the way in - there's no reason the driver can't know about their format.

Obviously we still need to do error handling for the general case.

matsuuran:
@zenhack @henn  I've been googling a bit to see if I can find a way to ask the switch for a list of its port names so that we can check to see if the port they provided is completely valid.  I've had no luck, just wanted to see if: 
 - either of you knew how to ask the switch for its port names?
 - is it an okay idea?

matsuuran:
@shwsun @naved001 do you know the answer to my question above?

zenhack:
I think what we'd been discussing was validating the /format/ of the port name. e.g. on the powerconnects, you have things like gi1/0/23; I believe @henn was talking about the situation where someone accidentally types e.g. gk1/0/23; we should be able to inspect this when the API call is made without interacting with the switch.

What the original issue is about though, is that we just don't handle communication failures with the switch *at all.* If a connection to the switch breaks, pexpect will throw an exception, which is currently not handled *anywhere* so the daemon will just crash until a sysadmin goes and restarts it.

At a minimum we ought to add a try/except at some appropriate level, which cleans up state and goes back to trying to process `NetworkingAction`s, ideally with some kind of exponential backoff.

Note that there *are* still things an admin can do to introduce persistent failures, e.g. if an admin mis-types the port as gi1/2/23 and that particular port doesn't exist, we don't currently have a way of detecting that and handling that. This is at least something that only an admn can do though.

Dealing with errors in the network daemon is hard because because we don't currently have a way for it to communicate any information to the user. That's probably a more general issue that needs addressing.

henn:
@matsuuran - do you think you could lead the discussion on this one?

matsuuran:
@henn what do you mean by lead discussion?  I have a branch somewhere that has most of this figured out and could probably knock it out after apply networking is closed

henn:
[...]

PR #673, opened 2016-10-24
Apply networking

matsuuran:
addresses #314


matsuuran:
@zenhack @gsilvis @henn @shwsun 

One thing I did want to ask is if anyone thought it would be good remove the looping from `deferred.py` because the loop in `apply_networking` will handle this anyway.  I'm not sure if there are any performance implications that I'm not seeing.  Otherwise, it seems like it would make the code smaller.

haas/deferred.py
```diff
@@ -49,32 +51,30 @@ def apply_networking():
...
-
-    # Close all of our sessions:
-    for session in switch_sessions.values():
-        session.disconnect()
+            logger.warn('Not modifying NIC %s; NIC is not on a port.' %
+                        nic.label)
 
-    # Then perform the database changes and delete them
-    for action in actions:
+        # Then perform the database changes and delete them
```
zenhack: Probably should get rid of the word "Then" since the first comment is gone now.

zenhack:
The problem with removing the loop in `apply_networking` is that we're still doing teardown of the switch connections at the very end (as we should), so there's state that's maintained across loop iterations. The loop in `cli.py` is there to keep us from tight-looping; if we don't do that os.sleep(2) call, then when there are no actions in the queue we'll just spin, wasting resources. So neither loop can go away.

zenhack:
This looks like basically the right logic, but (a) I'd like to see some tests, and (b) we should run this against some real switches, since travis won't do that for us.

Also,  there are a handful of what look like (good!) stylistic changes here that aren't otherwise related to this patch. I'm fine with this, but in the future could you make those their own commits, so the diffs are easier to understand?

Probably with this pr we should squash it when we merge.

shwsun:
It looks fine, and I am also waiting for some testing on real switch.

matsuuran:
@zenhack when you say you'd like to see some tests, are you referring to output from testing on a real switch or adding some pytest tests?

zenhack:
Quoting Nicholas Matsuura (2016-10-26 10:37:42)

>    @zenhack when you say you'd like to see some tests, are you
>    referring to output from testing on a real switch or adding some pytest
>    tests?

The latter

matsuuran:
Apologies for so many questions but wouldn't the testing in `vlan_networks.py` cover this?

matsuuran:
To do this I'm envisioning injecting several valid actions and then an invalid action to see if the valid actions were correctly committed to the db.  This however, will crash right now because #54 (which I should have a PR open for this weekend).  

Are you thinking of a different way of testing this, because to me what happens in `apply_networking()` is not transparent to pytest?

zenhack:
hm, one approach would be to add another switch driver for testing that inspect the database when it's called. Maybe we can see if things are being changed incrementally by creating a new Session (and thus a separate transaction), which shouldn't see any changes until commit, so we can do something like:

```python
current_count = session.query(NetworkingAction).count()
if last_count is None:
    last_count = current_count
else:
    assert current_count == last_count - 1, "network daemon did not commit previous change!"
    last_count -= 1

**Comment by zenhack** [comment]
@gsilvis, would appreciate your thoughts on the above approach.

**Comment by shwsun** [comment]
@matsuuran  any updates?
@zenhack Unfortunately George is now full time working in Google, I guess he might not be able to weigh in from time to time.

**Comment by naved001** [comment]
@matsuuran any updates on this pr?

**Comment by matsuuran** [comment]
@naved001 yep Jason and I are going to talk about this on Monday

**Comment by henn** [comment]
FYI - this is waiting on deployment tests.

**Comment by matsuuran** [comment]
What I'm finding is that once I call `deferred.apply_networking()` it clears the NetworkingActions almost immediately.

Right now what I'm doing is creating two networking actions in a mock switch test driver then calling `deferred.apply_networking()` and then doing essentially what @zenhack said to do above.

To troubleshoot I put a `sleep()` call in the `apply_networking()` for the mock switch and it still runs through.  This suggests that it is all running on one thread, so I'm looking for advice on how to move forward.

**Comment by zenhack** [comment]
Can you publish your work in progress somewhere (maybe as a gist)? I'm not sure we're 100% on the same page re: the approach and it might be more expedient for me to just look at what you've got than for us to try to go back and forth and understand what's being done in English.

**Comment by matsuuran** [comment]
https://gist.github.com/matsuuran/dc1def08682da1c1ea1ca6541a545070

This is just what I was debugging with to see if at any point I would see the `current_count` equal 1

**Comment by zenhack** [comment]
Ah, I think you've misunderstood what I was suggesting. deferred.apply_networking *is* going to still commit everything before returning, but we want to verify that it does so in individual commits. So checking after it returns isn't going to work.

My thought was to do the check *in the mock switch driver* (or possibly a new test switch driver for this purpose, since we don't necessarily want it to do this in other tests). deferred.apply_networking is going to call the switch driver in order to make the changes on the actual switch, so we can add a check to the switch driver that verifies that the last networking action is gone. If apply_networking is doing the right thing, namley call the switch, commit, call the switch, commit... then this should work.

Does that make sense?

**Comment by matsuuran** [comment]
Ohhhhhhh gotcha.  What I'll probably do is create a test mock switch in the same file and use that for the checks

**Comment by zenhack** [comment]
Sounds good, happy hacking.

**Comment by matsuuran** [comment]
So I've been banging my head on this for a little bit and I'm hoping someone can enlighten me.

Here is the [error](https://gist.github.com/matsuuran/f2a076cdc44d1bcca1d34e98ddf3fb93)

Here is the [test](https://gist.github.com/matsuuran/41fd0bac6391701d4218fedfdc7b5883)

Funny thing is, when I move `from haas.ext.obm.mock import MockObm` on lines 56 and 71 up to the imports at the top it works, but that breaks `test_dont_pollute_other_tests_extensions()` from `/hil/tests/unit/test_common.py`

**Comment by matsuuran** [comment]
So I figured that last error out, I wasn't including the mock extension in the configure_merge().  This small victory was short lived as now I'm struggling with this.  

```E       assert ["{u'alembic_...: None,", ...] == ["{u'alembic_v...: None,", ...]
E         At index 182 diff: " u'network': {'rows': [(1, u'stock_int_pub', None, True, u'100')," != " u'mock_test_switch': {'rows': [],"
E         Right contains more items, first extra item: '                    (297, 495, True),'
E         Use -v to get the full diff
tests/unit/migrations.py:249: AssertionError```

I'm only getting this on the postgreSQL tests.  I'm not really sure why I'm getting it beyond I'm putting a mock_test_switch which is part of my test into the database.  Anyone have advice?

File: tests/unit/ext/switches/mock_switch.py [hunk 2]
```diff
@@ -0,0 +1,146 @@
...
+import pytest
+
+
+from haas import config, deferred, model
+from haas.ext.obm.ipmi import Ipmi
+from haas.model import db
+from haas.test_common import config_testsuite, config_merge, \
+                             fresh_database, fail_on_log_warnings
+
+from sqlalchemy import Column, Integer, ForeignKey, String
```
zenhack: You don't need to (and shouldn't) import these from the sqlalchemy module; they're re-exported by `db`, so you can just do:

```python

foo = db.Column(db.Integer, ...)
bar = db.Column(...)
```

And so forth. Have a look at what the other database objects do.

tests/unit/ext/switches/mock_switch.py
```diff
@@ -0,0 +1,146 @@
...
+# software distributed under the License is distributed on an "AS
+# IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language
+# governing permissions and limitations under the License.
+
+import pytest
+
+
+from haas import config, deferred, model
+from haas.ext.obm.ipmi import Ipmi
```
zenhack: Don't import this at top-level; true of anything that starts with haas.ext.

zenhack:
I don't see the exact mistake in your code, but it looks like your switch module is getting loaded when it shouldn't be, and is contaminating the migration tests. That would explain why there's an extra table there (same kind of problem as what prompted the original fix).

Do you have the full output somewhere?

zenhack:
Ah, I bet I know what's happening: Because you define the model at top-level in a test file, it's going to get registered at test collection time, so it's going to be defined for every test. This is going to have the side-effect of creating a table for the model, which is going to break the migration tests.

The reason we don't import haas.ext at top-level is to avoid exactly this; they can't define database models at collection time.

The solution is probably to move the model definition somewhere under haas.ext.

Also, it could use a better name -- we already have a switch called 'mock'.

matsuuran:
Currently the `assert` is in the `MockTestSwitch` so if I move it we will have an assertion outside of the test subdirectory.  Are you good with that?

matsuuran:
@zenhack just wanted to see if you had a chance to take a look at this

haas/ext/switches/migrations/test/e87572b8fa00_create_test_switch_table.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = 'e87572b8fa00'
+down_revision = '57f4c30b0ad4'
+branch_labels = ('haas.ext.switches.test',)
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
zenhack: Should get rid of these comments (I'm assuming any needed adjusting has been done).

haas/ext/switches/test.py
```diff
@@ -0,0 +1,65 @@
...
+        }).validate(kwargs)
+
+    def session(self):
+        return self
+
+    def disconnect(self):
+        pass
+
+    def apply_networking(self, action):
+        current_count = db.session.query(model.NetworkingAction).count()
```
zenhack: We need to be creating a new session here, rather than using the request-local session (db.session), otherwise this will run in the same transaction as deferred.apply_networking's loop, which defeats the purpose of this test.

haas/ext/switches/test.py
```diff
@@ -0,0 +1,65 @@
...
+
+from collections import defaultdict
+from haas.migrations import paths
+from haas import model
+from haas.model import db, Switch
+from os.path import dirname, join
+
+paths[__name__] = join(dirname(__file__), 'migrations', 'test')
+
+LOCAL_STATE = defaultdict(lambda: defaultdict(dict))
```
zenhack: You're not using this anywhere; I assume it's copypasta from the other driver?

tests/unit/ext/switches/test.py
```diff
@@ -0,0 +1,113 @@
...
+        '00:11:22:33:44:55')
+
+
+@pytest.fixture()
+def network():
+    project = model.Project('anvil-nextgen')
+    return model.Network(project, [project], True, '102', 'hammernet')
+
+pytestmark = pytest.mark.usefixtures('configure',
+                                     'fresh_database')
```
naved001: you might want to add `fail_on_log_warnings` here.

matsuuran:
> The tests are in a file that suggests they're for a switch, but they're not, they're testing the incremental commit behavior.

This is because initially I had the switch code in there as well.  I've been thinking about it a bit and the only thing that I think makes sense is calling the test file `../unit/deferred.py` but wanted to see if you had thoughts

zenhack:
I like `../unit/deferred.py`.

> What do you think about throwing a new class in in the mock file and having a more specific class name?

I don't like it. At the least we'd have to write a migration script that adds the new table, since it would get imported by migration tests that previously didn't have that table.

One other option: we could define the class in a fixture, which we'd need to ensure was run before the fixture that initializes the db. e.g.

```python

DeferredTestSwitch = None

@pytest.fixture()
def _deferred_test_switch_class():
    global DeferredTestSwitch
    class DeferredTestSwitch_(db.Model):
        ...
    DeferredTestSwitch_.__name__ = 'DeferredTestSwitch'
    DeferredTestSwitch = DeferredTestSwitch_

```

It's more trickery that I'd like, but it means we can have it close to where it's actually used, and I'm not really any more of a fan of anything else we've discussed. There may be a slightly better way to handle the naming issue than the `__name__` stuff that I did above.

zenhack:
Would appreciate others' thoughts on the proposal above.

matsuuran:
@zenhack sorry I've been trying to implement the DeferredTestSwitch in a fixture as suggested.  I think that it is a better solution (albeit hacky) just because all the test code will reside within the test/ directory. 

Question, once I get it working will I be able to just delete the added migration files or will I have to do some sort of rollback

zenhack:
If you've got a local postgres db that you'd like to not trash, you can use alembic to revert that migration before deleting the scripts. I end up blowing away my local dev dbs on a regular basis though, esp. the sqlite ones (and the postgres one gets trashed hundreds of times if you run the test suite against it), so maybe not worth fussing with for a dev setup.

matsuuran:
@zenhack I'm getting an AssertionError on `db.session.add(action_native1)` because Port isn't recognizing DeferredTestSwitch as a switch, is there a solid way around this?  I tried to implement DeferredTestSwitch as a subclass of Switch but then we run into there not being a table for DeferredTestSwitch.  Relevant error message below:

`AssertionError: Attribute 'owner' on class '<class 'haas.model.Port'>' doesn't handle objects of type '<class 'deferred.DeferredTestSwitch'>'`

zenhack:
Probably something to do with the order in which the fixtures are being run. Is your code published somewhere that I can look at?

matsuuran:
Wanted to play with the fixture order to see if I could figure it out, but no dice.  This is the code I referenced last night along with the full error output.

[code](https://gist.github.com/matsuuran/4244b5c964643ab2668f927e2d665d25)
[error](https://gist.github.com/matsuuran/f2a076cdc44d1bcca1d34e98ddf3fb93)

matsuuran:
hey @zenhack just wanted to make sure you saw this

zenhack:
@matsuuran, sorry for the delay. You need to inherit from Switch, not db.Model. You'll also need to set API name. Have a look at the docstrings for Switch and the other switch implementations.

matsuuran:
@zenhack no worries.  So after adding what you suggested worked and the last thing I'm working on is your comment on using a new session.

I just wanted to let you know that when I use the old `deferred.py` code the test does not pass. When I use the old code it will have two networking actions each time because the database doesn't commit the change until the end, where as with the updated `deferred.py` code it will drop incrementally (as expected).  Just wanted to point this out to see if you still thought we needed a new session.

matsuuran:
hey @zenhack just wanted to check your reasoning for using a new session

zenhack:
The thing that's important is to make sure it's running in its own transaction, which it sounds like it is, but I'm uncomfortable with being unable to identify where the boundaries of the transactions are, and why. Could you do some digging?

matsuuran:
@zenhack call me dig dug

matsuuran:
It sounds like after each commit the session releases the transaction, so we should be good.

Here are the [docs](http://docs.sqlalchemy.org/en/latest/orm/session_transaction.html) and a relevant portion from them:

>For each Connection, the Session also maintains a Transaction object, which is acquired by calling Connection.begin() on each Connection, or if the Session object has been established using the flag twophase=True, a TwoPhaseTransaction object acquired via Connection.begin_twophase(). These transactions are all committed or rolled back corresponding to the invocation of the Session.commit() and Session.rollback() methods. A commit operation will also call the TwoPhaseTransaction.prepare() method on all transactions if applicable.

>When the transactional state is completed after a rollback or commit, the Session releases all Transaction and Connection resources, and goes back to the “begin” state, which will again invoke new Connection and Transaction objects as new requests to emit SQL statements are received.

zenhack:
@matsuuran, that was my understanding, but my question was about whether the loop in apply_networking and the check in the switch driver are *sharing* a transaction, regardless of when it starts or ends. If they were, you'd should "correct" behavior even if the transaction was never committed. What we want is for the check to be running in a different transaction (probably on its own connection). And if it is, I don't understand where the switch's transaction is being created.

i.e., what we want to happen conceptually is:

```python

# in the apply networking loop:
t1 = new_transaction()
do_stuff_with_transaction(t1)
call_switch()
commit(t1)

...

# in the switch driver:
do_other_stuff_with_transaction(t2)
```

What I want to make sure is *not* happening is the switch is *also* using `t1`, in which case it may see uncommitted changes.  If it *is* doing the right thing, I don't understand where `t2` is being created; we're using db.session in both places.

naved001:
@matsuuran any updates?

matsuuran:
Sorry I just got to Ohio will look at this weekend

naved001:
@matsuuran Hey NIck, any updates on this PR? Thanks!

zenhack:
So, we haven't seen any actual movement on this since March. This has merge conflicts now. Unless @matsuuran is actually around and ready to work on this, I think someone else needs to take over.

shwsun:
Since we need to close this PR and start from today's branch anyway, I would suggest some else to take it. (I dont think it can get done before the name changing

henn:
Closing, in preparation of the name change merge conflict coming up.

Issue #671, opened 2016-10-17
systemd service for `create_bridges`


SahilTikale:
The `create_bridges` command responsible for making bridges for all the vlans that HIL manages is currently run as a script from /etc/rc.local. We have observed that this does not always get executed. Sometimes we have to run this manually after a server reboot. 

Suggestion is to convert this to a systemd service and sequence it to be executed before the libvirtd service, so that VM depending on HIL networks find the requisite bridges on boot.

naved001:
@SahilTikale I made the script and made the changes in docs/INSTALL.rst on how to run the script. Do I have to write some tests for it?

shwsun:
I am nervous to these kind of changes, please make sure you have tested it on various environments.

naved001:
Yes, I'll test it before I create a pull request. You may look at the script [here](https://github.com/CCI-MOC/hil/compare/master...naved001:fix_671)

naved001:
This issue can be closed. It was fixed by #674

PR #674, opened 2016-10-28
Service to start create_bridges script

naved001:
This PR is to fix issue #671

### DIFF
--- a/docs/INSTALL.rst
+++ b/docs/INSTALL.rst
@@ -199,7 +199,50 @@ contains ``haas.cfg``. This pre-allocation is easier to reason about
 than on-demand creation, and allows HaaS to be run as an unprivileged user,
 but it also causes some limitations.  For instance, because of this, headnodes
 can only be connected to networks with allocated VLANs.  The bridges must also
-be pre-allocated again on each boot. For now, the recomended approach is to add::
+be pre-allocated again on each boot. For now, the recommened method is to use 
+systemd.  A systemd service for running the create_bridges script is available 
+in the 'scripts' directory.
+
+Name of the service is: create_bridges.service
+Name of the script is: create_bridges
+
+Centos:
+^^^^^^^
+
+Centos uses systemd to controll all its processes.
+
+Place the file create_bridges.service under:
+``/usr/lib/systemd/system/``
+
+And, place the file create_bridges and haas.cfg under /etc/
+
+Ubuntu:
+^^^^^^^
+Systemd is available from Ubuntu 15.04 onwards and LTS version 16.04 will ship with systemd by default.
+
+Place the file create_bridges.service under:
+``/lib/systemd/system/``
+
+And, place the file create_bridges and haas.cfg under /etc/
+
+
+Starting the service:
+^^^^^^^^^^^^^^^^^^^^^
+
+Following commands will start the daemon:
+``systemctl daemon-reload``
+``systemctl start create_bridges.service``
+
+You can check the status using:
+``systemctl status create_bridges.service``
+
+To auto-start the service on boot (recommended):
+``systemctl enable create_bridges.service``
+
+For systems that do not support systemd:
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+You can add the following line::
 
   (cd /etc && create_bridges)
 
--- a/docs/INSTALL.rst
+++ b/docs/INSTALL.rst
@@ -204,6 +204,7 @@ systemd.  A systemd service for running the create_bridges script is available
 in the 'scripts' directory.
 
 Name of the service is: create_bridges.service
+
 Name of the script is: create_bridges
 
 Centos:
@@ -214,7 +215,8 @@ Centos uses systemd to controll all its processes.
 Place the file create_bridges.service under:
 ``/usr/lib/systemd/system/``
 
-And, place the file create_bridges and haas.cfg under /etc/
+Place the file create_bridges and haas.cfg under:
+``/etc/``
 
 Ubuntu:
 ^^^^^^^
@@ -223,7 +225,8 @@ Systemd is available from Ubuntu 15.04 onwards and LTS version 16.04 will ship w
 Place the file create_bridges.service under:
 ``/lib/systemd/system/``
 
-And, place the file create_bridges and haas.cfg under /etc/
+Place the file create_bridges and haas.cfg under:
+``/etc/``
 
 
 Starting the service:
--- a/docs/INSTALL.rst
+++ b/docs/INSTALL.rst
@@ -215,19 +215,16 @@ Centos uses systemd to controll all its processes.
 Place the file create_bridges.service under:
 ``/usr/lib/systemd/system/``
 
-Place the file create_bridges and haas.cfg under:
-``/etc/``
-
 Ubuntu:
 ^^^^^^^
 Systemd is available from Ubuntu 15.04 onwards and LTS version 16.04 will ship with systemd by default.
 
+Edit the create_bridges.service file and change the ExecStart
+to 
+`` /usr/local/bin/create_bridges``
+
 Place the file create_bridges.service under:
 ``/lib/systemd/system/``
-
-Place the file create_bridges and haas.cfg under:
-``/etc/``
-
 
 Starting the service:
 ^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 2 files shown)


scripts/create_bridges.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Service to create bridges
+After=libvirtd.service
```
henn: This should run before libvirtd.

We should also make this run after the networks target.

docs/INSTALL.rst
```diff
@@ -199,7 +199,50 @@ contains ``haas.cfg``. This pre-allocation is easier to reason about
 than on-demand creation, and allows HaaS to be run as an unprivileged user,
 but it also causes some limitations.  For instance, because of this, headnodes
 can only be connected to networks with allocated VLANs.  The bridges must also
-be pre-allocated again on each boot. For now, the recomended approach is to add::
+be pre-allocated again on each boot. For now, the recommened method is to use 
+systemd.  A systemd service for running the create_bridges script is available
```
shwsun: I think it would be better if we use markdown syntax to render the `systemd`.

docs/INSTALL.rst
```diff
@@ -199,7 +199,50 @@ contains ``haas.cfg``. This pre-allocation is easier to reason about
...
 but it also causes some limitations.  For instance, because of this, headnodes
 can only be connected to networks with allocated VLANs.  The bridges must also
-be pre-allocated again on each boot. For now, the recomended approach is to add::
+be pre-allocated again on each boot. For now, the recommened method is to use 
+systemd.  A systemd service for running the create_bridges script is available 
+in the 'scripts' directory.
+
+Name of the service is: create_bridges.service
+
+Name of the script is: create_bridges
```
shwsun: Also mark the `create_bridges.service` and `create_bridges`, there are couple of same spots below.

docs/INSTALL.rst
```diff
@@ -199,7 +199,50 @@ contains ``haas.cfg``. This pre-allocation is easier to reason about
...
+Place the file create_bridges.service under:
+``/usr/lib/systemd/system/``
+
+Ubuntu:
+^^^^^^^
+Systemd is available from Ubuntu 15.04 onwards and LTS version 16.04 will ship with systemd by default.
+
+Edit the create_bridges.service file and change the ExecStart
+to 
+`` /usr/local/bin/create_bridges``
```
shwsun: incorrect blank space

shwsun:
@naved001 I am happy with this PR. Note that in order to pass the travis CI test, you need to have @knikolla 's workaround.

scripts/create_bridges.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Service to create bridges
+Before=libvirtd.service
+After=network.target
```
shwsun: I don't think we should use `network.target` here, as far as I am aware, `network.target` is vaguely defined and only used to make sure network services are properly killed before computer shutdown.  Here it should be `network-online.target` . See [here](https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/)

@henn  @SahilTikale Any corrections?

naved001: Okay. Do you think we should have a ``Wants=network-online.target`` dependency too? 
And while we are at it, should we modify [haas_network.service](https://github.com/CCI-MOC/hil/blob/master/scripts/haas_network.service) as well?

naved001: A ``Wants=`` or ``Requires=`` doesn't cause create_bridges to fail even if ``network.target``/``network-onine.target`` is disabled.

If we really want  ``create_bridges.service`` to fail if network is not up, then we could do a ``BindsTo=network-online.target``. 

I don't know if we need a strict dependency or if there are any side effects due to it. Services like ``libvirtd.service`` have a simple ``After=network.target`` and it seems to work just fine. 

@henn @SahilTikale @shwsun Thoughts?

shwsun:
@henn Can you take some time to take a look, is your concerns addressed?

scripts/create_bridges.service
shwsun: My concern is not fully addressed, but I am ok with the solution right now.

shwsun:
Just tried again, outputs are:
```
(.venv) [shwsun@moc-haas-master: scripts (fix_671)]$ sudo systemctl status create_bridges.service 
● create_bridges.service - Service to create bridges
   Loaded: loaded (/usr/lib/systemd/system/create_bridges.service; disabled; vendor preset: disabled)
   Active: inactive (dead)

Nov 09 11:37:58 moc-haas-master.prb.massopencloud.org create_bridges[9726]: device br-vlan1019 already exists; can't create bridge with the same name
Nov 09 11:37:58 moc-haas-master.prb.massopencloud.org create_bridges[9726]: ERROR: trying to add VLAN #1019 to IF -:eth0:-  error: No such device
Nov 09 11:37:58 moc-haas-master.prb.massopencloud.org create_bridges[9726]: interface eth0.1019 does not exist!
Nov 09 11:37:58 moc-haas-master.prb.massopencloud.org create_bridges[9726]: eth0.1019: ERROR while getting interface flags: No such device
Nov 09 11:37:58 moc-haas-master.prb.massopencloud.org create_bridges[9726]: eth0.1019: ERROR while getting interface flags: No such device
Nov 09 11:37:58 moc-haas-master.prb.massopencloud.org create_bridges[9726]: device br-vlan1020 already exists; can't create bridge with the same name
Nov 09 11:37:58 moc-haas-master.prb.massopencloud.org create_bridges[9726]: ERROR: trying to add VLAN #1020 to IF -:eth0:-  error: No such device
Nov 09 11:37:58 moc-haas-master.prb.massopencloud.org create_bridges[9726]: interface eth0.1020 does not exist!
Nov 09 11:37:58 moc-haas-master.prb.massopencloud.org create_bridges[9726]: eth0.1020: ERROR while getting interface flags: No such device
Nov 09 11:37:58 moc-haas-master.prb.massopencloud.org create_bridges[9726]: eth0.1020: ERROR while getting interface flags: No such device
```

shwsun:
@henn 's concern is already addressed, I will go ahead and hit the button

naved001:
@henn

scripts/create_bridges.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Service to create bridges
+Before=libvirtd.service
+After=network-online.target 
+After=network.target
+
+[Service]
+WorkingDirectory = /etc/
+ExecStart=/usr/bin/create_bridges
+Type=simple
```
henn: I think we want to replace this with `Type=oneshot` here so that we don't have a race condition with libvirt starting VMs before this script finishes.

From the [systemd docs](https://www.freedesktop.org/software/systemd/man/systemd.service.html):
> Behavior of oneshot is similar to simple; however, it is expected that the process has to exit before systemd starts follow-up units.

naved001: done

shwsun: @henn I am not sure about the race condition with `libvirt`. I think the case using `oneshot` is that user will have to do `systemctl start create-bridge.service` every time changes have been made to the `hil.cfg`, since the systemd service will run only once and exit. 
If it is the case, then it makes no difference between using this systemctl script and use bash script at all.

naved001:
> @henn I am not sure about the race condition with libvirt. I think the case using oneshot is that user will have to do systemctl start create-bridge.service every time changes have been made to the hil.cfg, since the systemd service will run only once and exit.
If it is the case, then it makes no difference between using this systemctl script and use bash script at all.

As far as I know, if you make changes to `haas.cfg` you need to restart the service (or reboot the system, or launch the python script) to reflect changes. `Type=simple` or `Type=oneshot` does not affect that. I tested this right now and changing types does not affect it. Moreover, I don't think that the `create_bridge.service` is sensitive to changes in `haas.cfg` anyway. `create_bridge.service` starts the `create_bridges` python script which reads from `haas.cfg`. So if you make any changes to `haas.cfg`, you either need to run the `create_bridges` python script again or restart the `create_bridges.service` (which just starts the python script for you).  

Does that answer your question? @shwsun

henn:
> As far as I know, if you make changes to haas.cfg you need to restart the service (or reboot the system, or launch the python script) to reflect changes. Type=simple or Type=oneshot does not affect that. I tested this right now and changing types does not affect it. Moreover, I don't think that the create_bridge.service is sensitive to changes in haas.cfg anyway. create_bridge.service starts the create_bridges python script which reads from haas.cfg. So if you make any changes to haas.cfg, you either need to run the create_bridges python script again or restart the create_bridges.service (which just starts the python script for you).

That's a good point. In the future, if we want to reload the service, systemd has primitives like `ExecReload=` and `ExecStop=` that would allow us to do that.

naved001:
@SahilTikale

naved001:
@SahilTikale could you re-review this please?

Issue #675, opened 2016-10-31
Travis CI building failure due to upstream bug


naved001:
The travis CI build is failing. It's even failing for a previous commit that successfully passed all the checks. So I think something is up with Travis-CI. 
@knikolla @henn 
Here's a portion of the job log:

> 2016-10-31 19:06:58.614 5399 ERROR keystone   File "/home/travis/build/CCI-MOC/hil/ci/keystone/keystone/.venv/local/lib/python2.7/site-packages/paste/deploy/loadwsgi.py", line 247, in loadapp
> 
> 2016-10-31 19:06:58.614 5399 ERROR keystone     return loadobj(APP, uri, name=name, **kw)
> 
> 2016-10-31 19:06:58.614 5399 ERROR keystone   File "/home/travis/build/CCI-MOC/hil/ci/keystone/keystone/.venv/local/lib/python2.7/site-packages/paste/deploy/loadwsgi.py", line 272, in loadobj
> 
> 2016-10-31 19:06:58.614 5399 ERROR keystone     return context.create()
> 
> 2016-10-31 19:06:58.614 5399 ERROR keystone   File "/home/travis/build/CCI-MOC/hil/ci/keystone/keystone/.venv/local/lib/python2.7/site-packages/paste/deploy/loadwsgi.py", line 710, in create
> 
> 2016-10-31 19:06:58.614 5399 ERROR keystone     return self.object_type.invoke(self)
> 
> 2016-10-31 19:06:58.614 5399 ERROR keystone   File "/home/travis/build/CCI-MOC/hil/ci/keystone/keystone/.venv/local/lib/python2.7/site-packages/paste/deploy/loadwsgi.py", line 144, in invoke
> 
> 2016-10-31 19:06:58.614 5399 ERROR keystone     **context.local_conf)
> 
> 2016-10-31 19:06:58.614 5399 ERROR keystone   File "/home/travis/build/CCI-MOC/hil/ci/keystone/keystone/.venv/local/lib/python2.7/site-packages/paste/deploy/util.py", line 55, in fix_call
> 
> 2016-10-31 19:06:58.614 5399 ERROR keystone     val = callable(*args, **kw)
> 
> 2016-10-31 19:06:58.614 5399 ERROR keystone   File "/home/travis/build/CCI-MOC/hil/ci/keystone/keystone/.venv/local/lib/python2.7/site-packages/paste/urlmap.py", line 31, in urlmap_factory
> 
[...]

PR #676, opened 2016-11-01
Fix Keystone gate breakage

knikolla:
Kombu was updated to 4.0 and oslo.messaging doesn't set an upper
bound to it resulting in package conflict versions for amqp.

Related to: #675

### DIFF
--- a/ci/keystone/keystone.sh
+++ b/ci/keystone/keystone.sh
@@ -46,6 +46,10 @@ case "$1" in
     # Make sure we have the latest:
     pip install --upgrade pip
 
+    # FIXME: Workaround for https://github.com/CCI-MOC/hil/issues/675
+    # remove once https://bugs.launchpad.net/oslo.messaging/+bug/1638263 is fixed
+    pip install "kombu>=3.0.25,<4.0"
+
     pip install -r requirements.txt
     pip install .
     pip install uwsgi # To actually run keystone; no webserver in the deps.
--- a/ci/keystone/install.sh
+++ b/ci/keystone/install.sh
@@ -5,4 +5,4 @@ pip install python-keystoneclient
 # The exact commit we use here is somewhat arbitrary, but we want
 # something that (a) won't change out from under our feet, and (b)
 # works with our existing tests.
-keystone_commit=10.0.0.0b2 ./ci/keystone/keystone.sh setup
+keystone_commit=10.0.0.0rc3 ./ci/keystone/keystone.sh setup


knikolla:
@naved001 @henn @zenhack

shwsun:
Looks good to me.  I guess it means we ought to be a little more careful with package related to openstack cause it may break from time to time?

knikolla:
@shwsun I'm not sure. In the [global upper contraints](https://github.com/openstack/requirements/blob/master/upper-constraints.txt) there is an upper constraint following the kombu 3.0.x releases all the way to a few years back, so I don't know why it still chose to install kombu 4.0 despite that.

zenhack:
The missing bound is an upstream bug. I'd like to handle it as follows:

1. See if it's been reported. If not, report it.
2. Open an issue on our tracker to monitor it
3. Refer to both issues in the comment in the CI script

The right place to fix this is in keystone, and I don't want to leave a workaround in our source longer than we need to; I want to monitor upstream bugs and remove outdated workarounds when we can.

@shwsun, as much as possible I'd like to have the CI do the thing that's *supposed* to work, so we can find out early if upstream has broken something that's going to break us, rather than wait until we update the CI to use a new version. Some things are volatile enogh that we shouldn't expect them to stay the same from one version to the next, but for things that we expect to work, I want to flush out bugs sooner rather than later.

ci/keystone/keystone.sh
```diff
@@ -60,7 +63,7 @@ case "$1" in
     pid=$!
     # Doing this after launching keystone will give it plenty of time to get
     # started without adding any wasteful calls to sleep:
-    pip install python-openstackclient
+    pip install python-openstackclient==3.2.0
```
zenhack: What's the reason for pinning this to `3.2.0`? Is there a more permissive constraint that "should" work? Either way, please document.

knikolla:
@zenhack The issue is known and there's already two patches up related to the issue https://review.openstack.org/#/c/392135/ and https://review.openstack.org/#/c/392245/

ci/keystone/keystone.sh
knikolla: No reasons for that specific version, but since we were freezing keystone to a version while openstackclient would always be the latest, we would probably be bound to maybe hit sooner or later another version incompatibility in dependent libraries. Either way, I removed it from the change since it's not required to fix the gate.

zenhack:
Ok, great. I do want to make sure we track this and remove the workaround once that's upstream; would you (a) link to the relevant issues/patches, and (b) Open an issue to monitor the upstream bug and remove this workaround once it's no longer needed?

knikolla:
@zenhack I linked to #675 in the comment and also to the upstream bug. I'll monitor the state upstream and remove the workaround once fixed.

zenhack:
Looks good to me.

shwsun:
I agree with @zenhack  and @knikolla , let's merge this one and keep the issue open.

PR #679, opened 2016-11-03
Minor cleanups

zenhack:
Independent, but both pretty small. See the commit messages.

//cc @henn, @shwsun, @knikolla


zenhack:
(snuck one more in there).

PR #680, opened 2016-11-07
adding tpm metadata for nodes

kylehogan:
- [x] update docs including but not limited to `rest_api.md`
- [x] include optional metadata field in `node_register`
- [x] testing
- [x] migrations
- [x] add `node_update_metadata`
- [x] update tests to make use of already created nodes


shwsun:
Hmmm, I am a little worried about @SahilTikale 's workload, @kylehogan have you talked to him yet? Or can we subtask those api calls somehow?

shwsun:
Hmmm, I am a little worried about @SahilTikale 's workload, @kylehogan have you talked to him yet? Or can we subtask those api calls somehow?

shwsun:
@kylehogan Regarding the pep8 errors, it is trivial that the generated version file needs to be modified. E.g, those `###` violations should be deleted and the commands in `upgrade` function need to get indented.

haas/api.py
```diff
@@ -423,6 +423,39 @@ def node_detach_network(node, nic, network):
     return '', 202
 
 
+@rest_call('PUT', '/TPM_metadata/<tpm_metadata>', Schema({
```
henn: Could we make the API more generic?
For example, instead of embedding TPM in the URL, make the PUT go to: `/node/<node>/metadata` with a body of
```json
{
"key": key,
"value": value
}
```
In the documentation, we'd then document the "standardized" metadata pieces, like "TPM_EK" and "TPM_BASE_PCRs".

zenhack:
Re: the pep8 errors, @shwsun is correct; the script will need to be modified manually to meet the style guidlines. see also the last paragraph of the commit message for 4bd81ff

Re: migrations, I did some playing around, and it turns out you need to change the `branch_labels` variable in the previous script (the one named by your `down_revision`) to be `None`. So in this case you'd need to change it in the network_acl migration. I'll submit a documentation fix pr.

zenhack:
Also, this touches API calls and doesn't update `rest_api.md`. please fix that. For that matter, I assume others have context on this change that I don't; what is this and why do we want it?

kylehogan:
@zenhack this change is for the secure cloud portion - it's so that show_node will display information about the node's tpm including PCR values and the public endorsement key. Thanks for the migration help!

@shwsun we discussed @SahilTikale being involved in the meeting this week - he will only be handling a subset of the api calls and these will not be included

kylehogan:
@henn regarding making the api more generic - I'm fine with it, but how far do you want to go? should I remove all mention of the tpm and just call it 'metadata'?

haas/api.py
zenhack: Agree. Suggestion: use URIs for the standardized keys, much like we do with switch types; this solves the namespacing mechanism.

zenhack:
> should I remove all mention of the tpm and just call it 'metadata'?

I was about to suggest just this. There's no reason the functionality should be bound to TPMs.

A nice addition also might be to add an optional `"metadata"` field to the node_register call, so admins can set that when registering a node, with a single call. We would keep the calls you've already implemented for modification after the fact.

kylehogan:
@zenhack @henn how does this look?

I did not change `cli.py` for node register, but I added the optional metadata field to the api call.

haas/api.py
```diff
@@ -423,6 +429,39 @@ def node_detach_network(node, nic, network):
     return '', 202
 
 
+@rest_call('PUT', '/node/<node>/metadata', Schema({
+    'node': basestring, 'value': basestring
+}))
+def node_register_metadata(node, label, value):
```
zenhack: You don't have a way to pass in the `label` parameter. Per my other comment, I would add it to the path. You'll also have to put it in the schema.

tests/unit/api/main.py
```diff
@@ -542,6 +561,102 @@ def test_node_register_nic_diff_nodes(self):
...
+        api._assert_absent_n(api._must_find(model.Node,
+                                            'compute-01'),
+                             model.Metadata, 'EK')
+
+    def test_node_delete_metadata_metadata_nexist(self):
+        api.node_register('compute-01', obm={
+                  "type": "http://schema.massopencloud.org/haas/v0/obm/ipmi",
+                  "host": "ipmihost",
+                  "user": "root",
+                  "password": "tapeworm"})
```
zenhack: Could we make use of some of the nodes create in initial_db or additional_db?

zenhack:
Thought: I don't see any reason to restrict the metadata to strings; we could shove arbitrary JSON in there.

docs/rest_api.md
```diff
@@ -493,19 +495,24 @@ The object will have at least the following fields:
...
         * "name", the name/label of the node (string).
         * "project", the name of the project a node belongs to or null if the node does not belong to a project
-        * "nics", a list of nics, each represted by a JSON object having
+        * "nics", a list of nics, each represented by a JSON object having
             at least the following fields:
 
                 - "label", the nic's label.
                 - "macaddr", the nic's mac address.
 		- "networks", a JSON object describing what networks are attached to the nic. The keys are channels and the values are the names of networks attached to those channels.
+	* "metadata", a list of metadata, each represented by a JSON object having the following fields
```
zenhack: You probably want a colon at the end of this.

docs/rest_api.md
```diff
@@ -493,19 +495,24 @@ The object will have at least the following fields:
...
+		- "value", the value of the metadata
 
 Response body:
 
     {"name": "node1",
 	 "project": "project1",
          "nics": [{"label": "nic1", "macaddr": "01:23:45:67:89", "networks": {"vlan/native": "pxe", "vlan/235": "storage"}},
-                       {"label": "nic2", "macaddr": "12:34:56:78:90", "networks":{"vlan/native": "public"}}]
+                       {"label": "nic2", "macaddr": "12:34:56:78:90", "networks":{"vlan/native": "public"}}],
+	"metadata":[{"label": "EK", "value": "pk"}]
```
zenhack: Could we reduce the line length by breaking stuff up across multiple lines? Try to keep it to < 80 chars. I realize it was already violating this before, but let's not make it worse.

haas/api.py
```diff
@@ -237,6 +237,12 @@ def node_register(node, **kwargs):
         raise BadArgumentError('%r is not a valid OBM type.' % obm_type)
     cls.validate(kwargs['obm'])
     node_obj = model.Node(label=node, obm=cls(**kwargs['obm']))
+    if 'metadata' in kwargs:
+        for label, value in kwargs['metadata'].items():
+            label = label
+            value = value
```
zenhack: Why are these two lines here? I assume this is a mistake.

tests/unit/api/main.py
```diff
@@ -412,6 +412,25 @@ def test_node_register(self):
...
+        api._must_find(model.Node, 'node-99')
+
+    def test_node_register_with_multiple_metadata(self):
+        api.node_register('node-99', obm={
+                  "type": "http://schema.massopencloud.org/haas/v0/obm/ipmi",
+                  "host": "ipmihost",
+                  "user": "root",
+                  "password": "tapeworm"}, metadata={
+                  "EK": "pk",
+                  "SHA256": "b5962d8173c14"})
```
kylehogan: the full length was too long to have lines less than 80 characters - couldn't get the formatting to look right

haas/api.py
kylehogan: yep, thanks. copy paste error

tests/unit/api/main.py
zenhack: Two things:

1. The set of metadata you choose to use in the tests doesn't really matter from a functional standpoint, so I'd pick something that is at least not incorrectly self-describing. Perhaps somthing that is more obviously arbitrary just test data, e.g. `{"foo": "bar", "baz": "quux"}`.
2. Python will concatinate string adjacent string literals (just like C does), so in cases like this you can do e.g.:

```python
{
    "EK": "pk",
    "SHA256": "e3b0c44298fc1c149afbf4c8996fb924"
              "27ae41e4649b934ca495991b7852b855",
}
```

docs/rest_api.md
kylehogan: done

docs/rest_api.md
kylehogan: done

docs/rest_api.md
```diff
@@ -353,6 +353,8 @@ see the documentation of the OBM driver in question (read `docs/obm-drivers.md`)
 Request Body:
 	{"obm": { "type": <obm-subtype>,
 		<additional sub-type specific values>}
+	 "metadata": {"label_1": "value_1",
+	 	     "label_2": "value_2"}
```
kylehogan: done

tests/unit/api/main.py
kylehogan: done

zenhack: Did you push? I don't see any new commits.

tests/unit/api/main.py
kylehogan: okay, I have the full hash there now

kylehogan:
> Thought: I don't see any reason to restrict the metadata to strings; we could shove arbitrary JSON in there.

I think you should be able to set the 'value' field to arbitrary JSON if you want. There aren't restrictions on what can be metadata  as long as it is labeled with a string.
Haven't tested it, but I see no reason why this wouldn't work. I can run some tests and update the docs to be more explicit for it if you want.

henn:
@zenhack - any chance you could re-take a look at this since @kylehogan pushed her newest changes?

haas/migrations/versions/57f4c30b0ad4_added_metadata.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = '57f4c30b0ad4'
+down_revision = '89630e3872ec'
+branch_labels = ('haas',)
+
+
+def upgrade():
+    # commands auto generated by Alembic - please adjust! #
```
zenhack: Please remove these comments.

haas/migrations/versions/57f4c30b0ad4_added_metadata.py
```diff
@@ -0,0 +1,35 @@
...
+    op.create_table('metadata',
+                    sa.Column('id', sa.Integer(), nullable=False),
+                    sa.Column('label', sa.String(), nullable=False),
+                    sa.Column('value', sa.String(), nullable=True),
```
zenhack: Why is this nullable?

tests/unit/api/main.py
```diff
@@ -412,6 +412,26 @@ def test_node_register(self):
                   "password": "tapeworm"})
         api._must_find(model.Node, 'node-99')
 
+    def test_node_register_with_metadata(self):
+        api.node_register('node-99', obm={
+                  "type": "http://schema.massopencloud.org/haas/v0/obm/ipmi",
+                  "host": "ipmihost",
+                  "user": "root",
+                  "password": "tapeworm"}, metadata={
+                  "EK": "pk"})
```
zenhack: Similar point to the thing I mentioned in rest_api.md; in general, data structures that are sub-components should be indented further.

zenhack:
@henn, thanks for the reminder.

haas/migrations/versions/57f4c30b0ad4_added_metadata.py
kylehogan: for if the value hadn't been set yet. My guess is that there will be some constant keys whose values might change between owners of the node. I didn't want to just leave the old values to be updated or force the entire piece of metadata to be deleted then recreated so it I thought it made sense to allow the value to be set to Null

haas/api.py
```diff
@@ -428,49 +428,36 @@ def node_detach_network(node, nic, network):
...
-    """
-    get_auth_backend().require_admin()
-    node = _must_find(model.Node, node)
-    metadata = _must_find_n(node, model.Metadata, label)
-    metadata.value = value
-
+    obj_inner = _namespaced_query(node, model.Metadata, label)
+    if obj_inner is not None:        
+        metadata = _must_find_n(node, model.Metadata, label)
+        metadata.value = value
```
zenhack: You'll need to convert `value` to a string to store it in the db (and do the reverse when you pull it out). Just a call to `json.dumps` will do.

tests/unit/api/main.py
```diff
@@ -427,7 +427,7 @@ def test_node_register_JSON_metadata(self):
                   "host": "ipmihost",
                   "user": "root",
                   "password": "tapeworm"}, metadata={
-                      "EK": {"val1": 1, "val2": 2}})
+                      "EK": json.dumps({"val1": 1, "val2": 2})})
```
zenhack: This needs to be done inside the API call, since the HTTP server is going to pass it the parsed JSON.

haas/migrations/versions/57f4c30b0ad4_added_metadata.py
zenhack: Given that we no longer have separate register/update calls I don't think this makes sense. The API doesn't provide a way to clear it without deleteing the object anyway, so if this is ever null it's a bug.

tests/unit/api/main.py
```diff
@@ -1822,12 +1836,12 @@ def test_show_node(self):
             'metadata': [
                 {
                     'label': 'EK',
-                    'value': 'pk',
+                    'value': json.dumps('pk'),
```
zenhack: You shouldn't need `dumps` here (or below); the `loads` should parse the whole structure, so you want `expected` to be the parsed python value, not the serialized JSON.

tests/unit/api/main.py
```diff
@@ -588,7 +602,7 @@ def test_node_update_metadata(self):
         metadata = api._must_find_n(api._must_find(model.Node,
                                                    'runway_node_0'),
                                     model.Metadata, 'EK')
-        assert metadata.value == 'new_pk'
+        assert metadata.value == json.dumps('new_pk')
```
zenhack: I'd do this the other way around: `assert json.loads(metadata.value) == 'new_pk'`. This is a slightly weaker constraint which isn't sensitive to semantically irrelevant details like whitespace. It should work fine anyway as long as we're using the same marshalling library, but I think it's a bit cleaner the other way.

tests/unit/api/main.py
```diff
@@ -639,7 +639,7 @@ def test_node_set_metadata_diff_nodes(self):
 
     def test_node_set_metadata_non_string(self):
         api.node_set_metadata('free_node_0', 'JSON',
-                              json.dumps({"val1": 1, "val2": 2}))
+                              '{"val1": 1, "val2": 2}')
```
zenhack: You shouldn't  quote this -- the function accepts *parsed* JSON, so it'll be a dictonary by the time it gets this far in a running server.

docs/rest_api.md
```diff
@@ -353,6 +353,8 @@ see the documentation of the OBM driver in question (read `docs/obm-drivers.md`)
 Request Body:
 	{"obm": { "type": <obm-subtype>,
 		<additional sub-type specific values>}
+	 "metadata": {"label_1": "value_1",
```
henn: Is there a reason to use `label` instead of `key`?

Not a big deal; just that I usually think of key/value.

docs/rest_api.md
```diff
@@ -493,19 +495,28 @@ The object will have at least the following fields:
...
 	 "project": "project1",
-         "nics": [{"label": "nic1", "macaddr": "01:23:45:67:89", "networks": {"vlan/native": "pxe", "vlan/235": "storage"}},
-                       {"label": "nic2", "macaddr": "12:34:56:78:90", "networks":{"vlan/native": "public"}}]
+         "nics": [{"label": "nic1", 
+	 	   "macaddr": "01:23:45:67:89", 
+		   "networks": {"vlan/native": "pxe", "vlan/235": "storage"}},
+                  {"label": "nic2", 
+		   "macaddr": "12:34:56:78:90", 
+		   "networks":{"vlan/native": "public"}}],
+	 "metadata":[{"label": "EK", "value": "pk"}]
```
henn: Is there a reason to not have all of metadata as a big dictionary? (this is a genuine question; I'm not sure what the appropriate REST/JSON convention is)

For example:
```json
"metadata": {
    "EK" : "pk",
    "PCRs": { "0":"0xabcd", "1":"0xdef"}
}
```

haas/api.py
```diff
@@ -237,6 +237,10 @@ def node_register(node, **kwargs):
         raise BadArgumentError('%r is not a valid OBM type.' % obm_type)
     cls.validate(kwargs['obm'])
     node_obj = model.Node(label=node, obm=cls(**kwargs['obm']))
+    if 'metadata' in kwargs:
```
henn: I think you'll need to add `metadata` as an optional parameter to the schema above.

docs/rest_api.md
kylehogan: i was using label since this was originally for cryptographic keys and I didn't want to overload the term 'key'

docs/rest_api.md
kylehogan: I'm not sure about the convention either - I structured the metadata to be handled the way nics are handled since they're both owned by nodes. Since metadata has only the key:value pair I suppose Icould change it to be a dictionary instead

henn: @zenhack - how do you feel? You're usually the most passionate about these types of things.

haas/api.py
```diff
@@ -423,6 +428,42 @@ def node_detach_network(node, nic, network):
...
+    """Register metadata on a node.
+
+    If the label already exists, the value will be updated.
+    """
+    get_auth_backend().require_admin()
+    node = _must_find(model.Node, node)
+    obj_inner = _namespaced_query(node, model.Metadata, label)
+    if obj_inner is not None:
+        metadata = _must_find_n(node, model.Metadata, label)
+        metadata.value = json.dumps(value)
```
henn: Is there a reason we aren't accepting value as a `basestring` in the first place?

Is json.dumps() here as a means of escaping the input?

zenhack: I suggested the value could be arbitrary JSON. We can't just store a python dictionary in the db though; converting it to a string is just so we can put it in a database column.

docs/rest_api.md
kylehogan: will do

kylehogan:
@zenhack @henn how is this looking?

docs/rest_api.md
```diff
@@ -608,6 +616,32 @@ Authorization requirements:
...
 
+#### node_set_metadata
+
+`PUT /node/<node>/metadata/<label>`
+
+Request Body: 
+
+	{
+		"value": <value>
+	}
```
zenhack: Hm, think it might be cleaner to just have the entire body of the request be the metadata. @henn, thoughts?

haas/api.py
```diff
@@ -1073,6 +1089,7 @@ def show_node(nodename):
                                      attachment.network.label)
                                     for attachment in n.attachments]),
                   } for n in node.nics],
+        'metadata': {m.label: m.value for m in node.metadata}
```
zenhack: This syntax was introduced in python 2.7; this is going to break on 2.6, which we *might* care about if we want to support CentOS 6. I'll leave it to the rest of the team whether or not we care. I don't think we're really testing with it, so maybe moot. But if we want to be 2.6 compatible you can do:

    dict([(m.label, m.value) for m in node.metadata])

zenhack:
I made a couple more comments. I don't much care about the 2.6/2.7 thing, but wanted to point it out and be intentional about what we do there. @henn, you have more visibility into where we are operationally than I do, do we care about CentOS 6 at all anymore?

Other than that I'm happy.

henn:
Ya - let's require 2.7. I just made #700 to update the docs.

haas/api.py
henn: The README says that HIL requires python 2.7.

I'd rather we put effort into supporting python 3 over 2.6 :-)

docs/rest_api.md
henn: Ya - I like the arbitrary-json-blob approach.

henn: One potential issue - I don't think sure arg verification will work with arbitrary body args, since it wants to verify each key. This is a bummer, because otherwise the programmer will have to make an extra step to get rid of the `value:stuff` for the values.

One way to address this would be to add an optional parameter to `@rest_call` like `literal_body`, where the body doesn't get interpreted and gets passed using some fixed name.

Given the trickiness of that code, my thought is to keep this the way it is now, and open an issue to enable arbitrary HTTP bodies to be passed through in the future.

@zenhack - what do you think?

zenhack:
Quoting Jason Hennessey (2017-01-23 21:58:25)

>    One potential issue - I don't think sure arg verification will work
>    with arbitrary body args, since it wants to verify each key. This is a
>    bummer, because otherwise the programmer will have to make an extra
>    step to get rid of the value:stuff for the values.
>
>    One way to address this would be to add an optional parameter to
>    @rest_call like literal_body, where the body doesn't get interpreted
>    and gets passed using some fixed name.
>
>    Given the trickiness of that code, my thought is to keep this the way
>    it is now, and open an issue to enable arbitrary HTTP bodies to be
>    passed through in the future.

The schema can just be:

```python

Schema({'value': object})

```

Which will accept any data type for the value. Since it's going through
a JSON parser first, that does exactly what we want.

Schema is one of my favorite python libraries, and there's a reason for
this, It's super flexible, and if you think it can't do something you're
probably wrong. In general, dig through the docs a bit before you assume
we can't specify something. (worst comes to worst, you can give it an
arbitrary object with a validate() method).

I actually wonder if we want to even have the `{"value": ...}` bit
there; why not just take the whole body (parsed as JSON) and use that as
the value?

zenhack:
Ack, discussion had been swapped out, so forgot I'd already suggested getting rid of `{"value": ...}`.
And actuall, I'm looking at `haas.rest._do_validation`, and @henn is right, it's going to insist on it being a dictionary. Otherwise we could just do `Schema(object)`. Let's leave it with the `{"value": ...}` wrapper for now, but I want an issue filed on a milestone that's a blocker for stabalizing the API; we should fix this before we're stuck with it.

henn:
@zenhack - are you happy with this PR then? If so, could you approve it? Thx

zenhack:
Right, forgot about that. Merging.

zenhack:
Crap, merging this with master resulted in a broken build:

https://travis-ci.org/CCI-MOC/hil/builds/195358637

I'd like to revert this and then ask @kylehogan to merge master back in and fix the issue before merging it again. @henn?

zenhack:
@shwsun, my suspicion is that the problem has to do with interactions between this and something else that had been merged into master but not pulled back into this branch. Without digging into it, the extension pollution patch seems like a good guess.

Perhaps we ought to require that things being merged into master are fast forwards, so this can't happen?

PR #683, opened 2016-11-08
Revert port

zenhack:
WIP; the commit messages give a good summary of what's done so far. Feel free to dig in to what's there; modulo comments I don't have plans to change the design dramatically.

(Edit: added a checklist)

* [x] Modify api server -> networking daemon interface to accommodate this
* [x] Make changes needed in the networking daemon.
* [x] Database migration script for the change to NetworkingAction
* [x] Expose the API call via HTTP
* [x] Add CLI wrapper for the call
* [x] Make the necessary changes to other switch drivers
* [x] Write new tests; the above passes existing tests, but these don't
      exercise the revert_port paths.
  * [x] Write unit tests that work agains the mock switch driver.
  * [x] Integrate into the deployment tests for real switches.

### DIFF
--- a/haas/ext/switches/_console.py
+++ b/haas/ext/switches/_console.py
@@ -84,11 +84,14 @@ class Session(object):
         if channel == 'vlan/native':
             old_native = NetworkAttachment.query.filter_by(
                 channel='vlan/native',
-                nic_id=port.nic.id).one().network.network_id
-            if network_id is None:
-                self.disable_native(old_native)
-            else:
+                nic_id=port.nic.id).first()
+            if old_native is not None:
+                old_native = old_native.network.network_id
+
+            if network_id is not None:
                 self.set_native(old_native, network_id)
+            elif old_native is not None:
+                self.disable_native(old_native)
         else:
             match = re.match(_CHANNEL_RE, channel)
             # TODO: I'd be more okay with this assertion if it weren't possible
--- a/haas/ext/switches/_console.py
+++ b/haas/ext/switches/_console.py
@@ -14,6 +14,7 @@
 """Common functionality for switches with a cisco-like console."""
 
 from abc import ABCMeta, abstractmethod
+from haas.model import Port, NetworkAttachment
 import re
 
 
@@ -65,26 +66,29 @@ class Session(object):
         """
 
     @abstractmethod
+    def disable_port(self):
+        """Disable all vlans on the current port."""
+
+    @abstractmethod
     def disconnect(self):
         """End the session. Must be at the main prompt."""
 
-    def apply_networking(self, action):
-        interface = action.nic.port.label
-        channel = action.channel
+    def modify_port(self, port, channel, network_id):
+        interface = port
+        port = Port.query.filter_by(label=port,
+                                    owner_id=self.switch.id).one()
 
         self.enter_if_prompt(interface)
         self.console.expect(self.if_prompt)
 
         if channel == 'vlan/native':
-            old_native = None
-            old_attachments = filter(lambda a: a.channel == 'vlan/native',
-                                     action.nic.attachments)
-            if len(old_attachments) != 0:
-                old_native = old_attachments[0].network.network_id
-            if action.new_network is None:
+            old_native = NetworkAttachment.query.filter_by(
+                channel='vlan/native',
+                nic_id=port.nic.id).one().network.network_id
+            if network_id is None:
                 self.disable_native(old_native)
             else:
-                self.set_native(old_native, action.new_network.network_id)
+                self.set_native(old_native, network_id)
         else:
             match = re.match(_CHANNEL_RE, channel)
             # TODO: I'd be more okay with this assertion if it weren't possible
@@ -94,11 +98,20 @@ class Session(object):
             assert match is not None, "HaaS passed an invalid channel to the" \
                 "switch!"
             vlan_id = match.groups()[0]
-            if action.new_network is None:
+            if network_id is None:
                 self.disable_vlan(vlan_id)
             else:
-                assert action.new_network.network_id == vlan_id
+                assert network_id == vlan_id
                 self.enable_vlan(vlan_id)
+
+        self.exit_if_prompt()
+        self.console.expect(self.config_prompt)
+
+    def revert_port(self, port):
+        self.enter_if_prompt(port)
+        self.console.expect(self.if_prompt)
+
+        self.disable_port()
 
         self.exit_if_prompt()
         self.console.expect(self.config_prompt)
--- a/haas/ext/switches/brocade.py
+++ b/haas/ext/switches/brocade.py
@@ -92,7 +92,7 @@ class Brocade(Switch):
                 self._add_vlan_to_trunk(interface, vlan_id)
 
     def revert_port(self, port):
-        self._remove_all_vlans_from_trunk(port, vlan)
+        self._remove_all_vlans_from_trunk(port)
         self._remove_native_vlan(port)
 
     def get_port_networks(self, ports):
@@ -223,9 +223,9 @@ class Brocade(Switch):
         payload = '<vlan><remove>%s</remove></vlan>' % vlan
         self._make_request('PUT', url, data=payload)
 
-    def _remove_all_vlans_from_trunk(self, interface, vlan):
+    def _remove_all_vlans_from_trunk(self, interface):
         url = self._construct_url(interface, suffix='trunk/allowed/vlan')
-        payload = '<vlan><none>true</none></vlan>' % vlan
+        payload = '<vlan><none>true</none></vlan>'
         requests.put(url, data=payload, auth=self._auth)
 
     def _set_native_vlan(self, interface, vlan):
--- a/haas/ext/switches/brocade.py
+++ b/haas/ext/switches/brocade.py
@@ -61,11 +61,6 @@ class Brocade(Switch):
         pass
 
     def modify_port(self, port, channel, network_id):
-        """ Apply a NetworkingAction to the switch.
-
-        Args:
-            action: NetworkingAction to apply to the switch.
-        """
         # XXX: We ought to be able to do a Port.query ... one() here, but
         # there's somthing I(zenhack)  don't understand going on with when
         # things are committed in the tests for this driver, and we don't
@@ -224,6 +219,11 @@ class Brocade(Switch):
         self._make_request('PUT', url, data=payload)
 
     def _remove_all_vlans_from_trunk(self, interface):
+        """ Remove all vlan from a trunk port.
+
+        Args:
+            interface: interface to remove the vlan from
+        """
         url = self._construct_url(interface, suffix='trunk/allowed/vlan')
         payload = '<vlan><none>true</none></vlan>'
         requests.put(url, data=payload, auth=self._auth)
--- a/haas/ext/switches/brocade.py
+++ b/haas/ext/switches/brocade.py
@@ -95,10 +95,10 @@ class Brocade(Switch):
         # TODO: there is likely a more efficient way to do this; we may want
         # to inspect the brocade documentation to see if there's a way to
         # clear the port without doing every vlan individually.
-        vlans = self._get_vlans()
+        vlans = self._get_vlans(port)
         for _, vlan in vlans:
-            self._remove_vlan_from_trunk()
-        self._remove_native_vlan()
+            self._remove_vlan_from_trunk(port, vlan)
+        self._remove_native_vlan(port)
 
     def get_port_networks(self, ports):
         """Get port configurations of the switch.
--- a/haas/ext/switches/brocade.py
+++ b/haas/ext/switches/brocade.py
@@ -25,7 +25,7 @@ import requests
 import schema
 
 from haas.migrations import paths
-from haas.model import db, Switch
+from haas.model import db, Switch, Port
 
 paths[__name__] = join(dirname(__file__), 'migrations', 'brocade')
 
@@ -60,32 +60,45 @@ class Brocade(Switch):
     def disconnect(self):
         pass
 
-    def apply_networking(self, action):
+    def modify_port(self, port, channel, network_id):
         """ Apply a NetworkingAction to the switch.
 
         Args:
             action: NetworkingAction to apply to the switch.
         """
-        interface = action.nic.port.label
-        channel = action.channel
+        # XXX: We ought to be able to do a Port.query ... one() here, but
+        # there's somthing I(zenhack)  don't understand going on with when
+        # things are committed in the tests for this driver, and we don't
+        # get any results that way. We should figure out what's going on with
+        # that test and change this.
+        (port,) = filter(lambda p: p.label == port, self.ports)
+        interface = port.label
 
         if channel == 'vlan/native':
-            if action.new_network is None:
+            if network_id is None:
                 self._remove_native_vlan(interface)
             else:
-                self._set_native_vlan(interface,
-                                      action.new_network.network_id)
+                self._set_native_vlan(interface, network_id)
         else:
             match = re.match(re.compile(r'vlan/(\d+)'), channel)
             assert match is not None, "HaaS passed an invalid channel to the" \
                 " switch!"
             vlan_id = match.groups()[0]
 
-            if action.new_network is None:
+            if network_id is None:
                 self._remove_vlan_from_trunk(interface, vlan_id)
             else:
-                assert action.new_network.network_id == vlan_id
+                assert network_id == vlan_id
                 self._add_vlan_to_trunk(interface, vlan_id)
+
+    def port_revert(self, port):
+        # TODO: there is likely a more efficient way to do this; we may want
+        # to inspect the brocade documentation to see if there's a way to
+        # clear the port without doing every vlan individually.
+        vlans = self._get_vlans()
+        for _, vlan in vlans:
+            self._remove_vlan_from_trunk()
+        self._remove_native_vlan()
 
     def get_port_networks(self, ports):
         """Get port configurations of the switch.
--- a/haas/ext/switches/brocade.py
+++ b/haas/ext/switches/brocade.py
@@ -91,7 +91,7 @@ class Brocade(Switch):
                 assert network_id == vlan_id
                 self._add_vlan_to_trunk(interface, vlan_id)
 
-    def port_revert(self, port):
+    def revert_port(self, port):
         # TODO: there is likely a more efficient way to do this; we may want
         # to inspect the brocade documentation to see if there's a way to
         # clear the port without doing every vlan individually.
--- a/haas/ext/switches/brocade.py
+++ b/haas/ext/switches/brocade.py
@@ -88,7 +88,8 @@ class Brocade(Switch):
 
     def revert_port(self, port):
         self._remove_all_vlans_from_trunk(port)
-        self._remove_native_vlan(port)
+        if self._get_native_vlan(port) is not None:
+            self._remove_native_vlan(port)
 
     def get_port_networks(self, ports):
         """Get port configurations of the switch.
--- a/haas/ext/switches/brocade.py
+++ b/haas/ext/switches/brocade.py
@@ -92,12 +92,7 @@ class Brocade(Switch):
                 self._add_vlan_to_trunk(interface, vlan_id)
 
     def revert_port(self, port):
-        # TODO: there is likely a more efficient way to do this; we may want
-        # to inspect the brocade documentation to see if there's a way to
-        # clear the port without doing every vlan individually.
-        vlans = self._get_vlans(port)
-        for _, vlan in vlans:
-            self._remove_vlan_from_trunk(port, vlan)
+        self._remove_all_vlans_from_trunk(port, vlan)
         self._remove_native_vlan(port)
 
     def get_port_networks(self, ports):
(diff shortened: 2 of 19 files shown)


shwsun:
@zenhack I think the cli entry is missing.

zenhack:
@shwsun, good spot. I started a proper checklist above.

henn:
Hey @zehnack

Do you have everything you need to test this in the lab?

zenhack:
@henn, I seem to recall looking through what was available and concluding it was enough; sorry for the delay actually digging into it. I'll try to actually tackle it soon.

henn:
Hey @zenhack. Anything we can do to help out with this one?

zenhack:
@henn, #698 and #696 need to get merged first (in that order), so if you could work on reviewing those that would be the most useful thing.

Sorry for the general slowness on this one.

henn:
@zenhack - I noticed a conflict and clicked github's magic "Resolve" button. Hope you don't mind :-)

Are you waiting for anything from my side on this?

zenhack:
@henn, I'm going to back out your merge and do it manually, since it's broken the tests. Please let me handle it going forward.

I'm not waiting on you, life has just been hectic. I'm going to start digging into it again now. There's not *much* left to do besides the real hardware drivers, and I may find I want to be physically present to deal with those; if so I'll ping you re: appropriate times to let me into the server room.

zenhack:
@henn, this is *mostly* done now. I'm going to drop by tomorrow or Friday, as I want to be physically present to test this on the real hardware. There are two one-liners that need to be filled in and per the comment in the brocade driver we may want to be smarter about our use of the API, but other than that, testing, and review we're good.

zenhack:
Addendum to the stuff that we might still want to address: There's something really weird going on in `tests/unit/ext/switches/brocade.py`. Stuff doesn't get committed to the db when I think it should, and actually looking at the code I don't see where the db even gets initialized; we're not using that fixture. But I'll stare at it when I'm a bit more fresh.

naved001:
I tested `revert_port` on a brocade switch and got this:

> AttributeError: 'Brocade' object has no attribute 'revert_port'

probably because it's called `port_revert` in the brocade driver?

zenhack:
@naved001, good spot. I pushed a fix, could you give it another pass?

knikolla:
@zenhack In the brocade driver's unit tests nothing is actually committed to the database. I just use the objects as fixtures.

zenhack:
@knikolla, I'm not sure I follow -- why are you mentioning this?

knikolla:
@zenhack in response to a prior comment, and comment on the code about how you weren't sure why the brocade unit tests weren't putting stuff in the db. Thought I'd clarify.

zenhack:
Ah, yeah, I remember that now - had lost context. Thanks.

zenhack:
@naved001, ping.

naved001:
@zenhack I'll be on it tomorrow. For some reason the brocade switch was down the other day.

naved001:
Works fine on the dell switch (removes all vlans). 
running `port_revert` on brocade crashes network server. on restarting it throws this error:
```
Invalid arguements.  Usage:
  serve_networks
      Start the HaaS networking server
```

will test it on the cisco switch too.

zenhack:
Huh. Do you have a stack trace from the "crash"?

My guess is that a TypeError is getting thrown for some reason *other* than invalid arguments, and then we're hitting this:

https://github.com/CCI-MOC/hil/blob/master/haas/cli.py#L143

Easy way to find out is to put a breakpoint in that except and inspect the error.

zenhack:
Would be a pretty simple tweak to have the `cmd` decorator just check the arguments up front, instead of catching an overly-broad `TypeError`. It's always just a list of strings, so the check is pretty trivial.

naved001:
Tested  `port_revert`on the Cisco switch and it works! 

> Huh. Do you have a stack trace from the "crash"?

I'll share it asap.

zenhack:
@naved001, I added the missing parameters, thanks for catching those.

Re: not removing the native vlan, does disconnecting a network that's attached natively work with the code currently on master? My guess is no, since I'm just making a call out to the code that should be doing that already; if I'm correct then this is an already-present bug.

naved001:
>  My guess is no, since I'm just making a call out to the code that should be doing that already; if I'm correct then this is an already-present bug.

You are right. I'll make an issue about it.

naved001:
@zenhack I have manually tested this PR for all the 3 switches successfully, and my comments have been addressed so I am mostly happy with this PR. Other than that, my only concern is the failing deployment tests (mentioned that in the email thread).  

Also, we need 1 more reviewer here.

zenhack:
@shwsun, would you take a second lool?

naved001:
@zenhack Do you think `port_revert` for brocade should also disable the switchport because that seems to be the default state and in #702 we first enable switchport and then set mode. 
@henn @knikolla thoughts?

zenhack:
Sounds sensible enough, though we certainly can't do that until #702 is merged.

The changes I made to get the brocade switch to support port_revert are pretty dumb; there's probably a more efficient way to just shut everything on the port down.

@knikolla, is there a set of docs somewhere that you've been referencing when hacking on the brocade driver?

haas/ext/switches/brocade.py
```diff
@@ -60,33 +60,46 @@ def session(self):
...
+            if network_id is None:
                 self._remove_vlan_from_trunk(interface, vlan_id)
             else:
-                assert action.new_network.network_id == vlan_id
+                assert network_id == vlan_id
                 self._add_vlan_to_trunk(interface, vlan_id)
 
+    def revert_port(self, port):
+        # TODO: there is likely a more efficient way to do this; we may want
+        # to inspect the brocade documentation to see if there's a way to
```
naved001: now that I think about it, if you just disable the switchport. You wouldn't have to detach all the vlans 1 by 1.

zenhack: @naved001, I'd be worried about what happens when the switchport gets turned back on; will all of the old vlans get re-enabled?

shwsun:
@zenhack I will work on reviewing this, probably over this weekend.

In terms of the VDX command reference, see http://www.brocade.com/content/html/en/command-reference-guide/nos-700-commandref/

haas/ext/switches/brocade.py
naved001: @zenhack that's a good concern. In the time that I've been playing around with the brocade, the old vlans never got re-enabled. Although I am fairly confident about this, we should still look into the brocade documentation. I'll update this comment if I find it in the docs.

zenhack:
Quoting Naved Ansari (2017-02-24 15:47:26)

>    @zenhack that's a good concern. In the time that I've been playing
>    around with the brocade, the old vlans never got re-enabled. Although I
>    am fairly confident about this, we should still look into the brocade
>    documentation. I'll update this comment if I find it in the docs.

My preference would be to keep the explicit removal in there unless we
have a strong argument as to why it's a problem -- defense in-depth and
all that.

I *would* like to find a way to batch the disable calls; I wrote it as
is just because it didn't require digging through (and finding; see my
other comment) switch documentation.

zenhack:
@shwsun, the brocade driver is using some kind of HTTP interface, not the CLI. Know where the docs for that are?

naved001:
@zenhack Here's the REST API guide for brocade. 
http://www.brocade.com/content/dam/common/documents/content-types/api-reference-guide/nos-700-restapiguide.pdf

knikolla:
@zenhack this is a condensed version a brocade engineer sent me https://drive.google.com/file/d/0B0lXTUzW8Bi5cUJ2SnY5cm1Wd1U/view?usp=sharing

I've used that, together with the pdf naved referenced.

zenhack:
Okay, I just patched it to do the same kind of batching we're doing with the other two drivers. Needs testing; could someone run that on the brocade and see if it does the right thing?

naved001:
@zenhack I ran it on brocade; it crashed the network server with the following message:
```
DEBUG:requests.packages.urllib3.connectionpool:Starting new HTTP connection (1): 192.168.3.240
DEBUG:requests.packages.urllib3.connectionpool:http://192.168.3.240:80 "PUT /rest/config/running/interface/TenGigabitEthernet/%221/0/4%22/switchport/mode HTTP/1.1" 204 0
DEBUG:requests.packages.urllib3.connectionpool:Starting new HTTP connection (1): 192.168.3.240
DEBUG:requests.packages.urllib3.connectionpool:http://192.168.3.240:80 "PUT /rest/config/running/interface/TenGigabitEthernet/%221/0/4%22/switchport/trunk/allowed/vlan HTTP/1.1" 204 0
Fatal Error: set_auth_backed() called twice. Make sure you don't have conflicting extensions loaded.

``` 
I have the changes from the master which should print out the exact error message, but I still see `set_auth_backed()`

zenhack:
Is this otherwise the same setup you were (successfully) using before? What's in the extensions section of haas.cfg?

Weren't we getting a  similar error with the nexus not long ago? I can't remember what was going on there, and I'm not finding the relevant comments. @shwsun ?

can you put a breakpoint at the top of `set_auth_backend()` (in haas/auth.py) and see what's calling it (and verify that it is actually being called twice)?

I suspect this si something related to setup, not the commit I just pushed.

zenhack:
(Just rebased on top of master for the CI fixes)

naved001:
> Is this otherwise the same setup you were (successfully) using before? What's in the extensions section of haas.cfg? 

Yes, it's the same setup that I was using before and the same haas.cfg that I used earlier (for testing this PR and for other stuff) 

> Weren't we getting a similar error with the nexus not long ago? I can't remember what was going on there, and I'm not finding the relevant comments. 

Yes. Issues #723 #726 #727  

> can you put a breakpoint at the top of set_auth_backend() (in haas/auth.py) and see what's calling it (and verify that it is actually being called twice)?

Okay, I'll do that and let you know what I find.

henn:
FYI - we narrowed this error down to the brocade code:
```python
Traceback (most recent call last):
  File "/home/naved001/ian/hil/.venv/bin/haas", line 8, in <module>
    execfile(__file__)
  File "/home/naved001/ian/hil/scripts/haas", line 18, in <module>
    cli.main()
  File "/home/naved001/ian/hil/haas/cli.py", line 866, in main
    command_dict[sys.argv[1]](*sys.argv[2:])
  File "/home/naved001/ian/hil/haas/cli.py", line 157, in wrapped
    f(*args, **kwargs)
  File "/home/naved001/ian/hil/haas/cli.py", line 314, in serve_networks
    while deferred.apply_networking():
  File "/home/naved001/ian/hil/haas/deferred.py", line 103, in apply_networking
    session.handle_action(action)
  File "/home/naved001/ian/hil/haas/deferred.py", line 36, in handle_action
    getattr(self, action.type)(action)
  File "/home/naved001/ian/hil/haas/deferred.py", line 61, in revert_port
    session.revert_port(action.nic.port.label)
  File "/home/naved001/ian/hil/haas/ext/switches/brocade.py", line 95, in revert_port
    self._remove_all_vlans_from_trunk(port, vlan)
NameError: global name 'vlan' is not defined
```

This was made more difficult to catch because of the haas script code:
```python
try:
    execfile(__file__)
except NameError, e:
    print(e)
    exec(compile(open(__file__).read(), __file__, 'exec'))
```

zenhack:
Pushed a fix.

naved001:
> Pushed a fix. 

Batch removal works now!

zenhack:
Quoting Naved Ansari (2017-03-10 17:05:08)

> >    Batch removal works now!

Neat. In that case I think all that's left is dealing with comments from
you all.

kylehogan:
I'm going over this now. Some questions: 

1. from reading though comments it seems that that no one has any outstanding issues - is this the case?
2. is #702 (which is stuck on #740) out of scope or will this PR be waiting on deployment test related fixes?
3. based on above two PR seems finished - why the Do Not Merge label?

I'll add any other comments/questions I end up having inline with the code. Should finish reviewing soon.

zenhack:
1. Correct, everything that has been raised has been addressed.
2. @naved001, it seemed from your comments above that you managed to get the tests passing for this pr; I'm a little unclear on where that stands now? #702 is independent of this pr, but probably there will be merge conflicts between the two that will need to be resolved.
3. Because I forgot to remove it; it's gone now.

naved001:
@zenhack 
I have tested it all manually with real switches and port revert worked perfectly. But I couldn't get the deployment tests ( `tests/deployment/*_networks.py` ) to pass on any of the switches. But it may or may not be related to this PR. I have raised #740 to discuss that.

haas/ext/switches/brocade.py
```diff
@@ -60,33 +60,41 @@ def session(self):
     def disconnect(self):
         pass
 
-    def apply_networking(self, action):
+    def modify_port(self, port, channel, network_id):
         """ Apply a NetworkingAction to the switch.
```
shwsun: @zenhack can you update the comment a bit here? I don't mind getting rid of it or else.

haas/ext/switches/brocade.py
```diff
@@ -60,33 +60,41 @@ def session(self):
...
 
         Args:
             action: NetworkingAction to apply to the switch.
         """
-        interface = action.nic.port.label
-        channel = action.channel
+        # XXX: We ought to be able to do a Port.query ... one() here, but
```
shwsun: Replace XXX with FIXME? I suggest open an issue about this since it can be orthogonal

haas/ext/switches/brocade.py
```diff
@@ -60,33 +60,41 @@ def session(self):
...
 
-            if action.new_network is None:
+            if network_id is None:
                 self._remove_vlan_from_trunk(interface, vlan_id)
             else:
-                assert action.new_network.network_id == vlan_id
+                assert network_id == vlan_id
                 self._add_vlan_to_trunk(interface, vlan_id)
 
+    def revert_port(self, port):
```
shwsun: Can you comment on this function a little? It would be good to be consistent to all others

haas/ext/switches/brocade.py
```diff
@@ -207,6 +215,11 @@ def _remove_vlan_from_trunk(self, interface, vlan):
         payload = '<vlan><remove>%s</remove></vlan>' % vlan
         requests.put(url, data=payload, auth=self._auth)
 
+    def _remove_all_vlans_from_trunk(self, interface):
```
shwsun: some comment here too

naved001:
@shwsun I [have tested](https://github.com/CCI-MOC/hil/pull/683#issuecomment-282120815) it on all 3 switches (dell, brocade, cisco). You can test it again if you feel like it.

shwsun:
@naved001 I will be happy with the tests you have run.

Another concern I have is @SahilTikale 's PR, I don't want to preempt his work again and make him do extra work. I guess this PR do request some changes in #617 right? @zenhack @zenhack  If so how can we resolve it?

haas/ext/switches/brocade.py
zenhack: I'm going to just kill the docstring, since it's redundant with the one for th superclass. Thanks for spotting it.

haas/ext/switches/brocade.py
zenhack: I think it makes sense to do the same thing as with modify_port, and just lean on the superclass's docstring.

haas/ext/switches/brocade.py
zenhack: FIXME is typically used for something that's actually *broken.* as far as I know this code is *correct,* It's just not the most elegant, so I think XXX is more appropriate. I will open an issue for this. I want to wait until this is merged though so I can link to section of the code in the merge commit, rather than some potentially transient state on my branch.

zenhack:
@shwsun, addressed or responsed to each of your comments. Re: @SahilTikale's pr, the merge conflicts should be relatively minor, as that's mostly client-side and this is mostly server-side. Also, I will gladly assist him with the conflicts. (Also, if we end up merging his first, I have no problem fixing the conflicts on my end).

shwsun:
@zenhack I am happy with it then. 

@SahilTikale do you feel like ok with the discussion here? It might be the case that `revert_port` will get merged first, and Ian plus me will be assisting you on adopting changes to CLI lib.

naved001:
@zenhack deployment tests are passing for this PR only on the brocade switch with the fix from #742 

For the dell and cisco, we can debug when we meet.

zenhack:
@naved001, cool, see you then.

haas/ext/switches/_console.py
```diff
@@ -65,26 +66,29 @@ def disable_native(self, vlan_id):
...
 
         if channel == 'vlan/native':
-            old_native = None
-            old_attachments = filter(lambda a: a.channel == 'vlan/native',
-                                     action.nic.attachments)
-            if len(old_attachments) != 0:
-                old_native = old_attachments[0].network.network_id
-            if action.new_network is None:
+            old_native = NetworkAttachment.query.filter_by(
+                channel='vlan/native',
```
naved001: just a reminder: you need to merge master back into this to get these changes
https://github.com/CCI-MOC/hil/blob/master/haas/ext/switches/_console.py#L80

without this the deployment tests for cisco and dell fail.

zenhack:
@naved001, I merged it back in, but it didn't make that change. I think this may have actually been something I touched while working on this. In any case, I made a similar change to the one I did on the fly yesterday; you should probably re-run the tests just to be sure.

naved001:
@zenhack The tests pass for cisco and dell. 

But now the tests fail for the brocade switch because we merged #702 and here's why:

when `port_revert` tries to call `_remove_native_vlan` the switch now [raises an error](https://github.com/CCI-MOC/hil/blob/master/haas/ext/switches/brocade.py#L283). This was added in #702. Earlier that didn't happen and the tests would pass. There's a [bug](https://github.com/CCI-MOC/hil/pull/683#issuecomment-281875830) in the brocade driver due to which it doesn't remove the native vlan. I have opened an issue here #728

naved001:
@zenhack I have 2 ideas in mind to get the deployment tests to pass for the brocade. 

1. In `revert_port` we should check if a network is natively attached before calling  `_remove_native_vlan`. 

2. Update the brocade driver to not raise an error, instead just write it to the logs. so instead of `logger.error` we do `logger.debug` https://github.com/CCI-MOC/hil/blob/master/haas/ext/switches/brocade.py#L283 or something alone these lines. 

or any other solution that you like. 

Here's the error log:

```
tests/deployment/vlan_networks.py:91: in create_networks
    deferred.apply_networking()
haas/deferred.py:103: in apply_networking
    session.handle_action(action)
haas/deferred.py:36: in handle_action
    getattr(self, action.type)(action)
haas/deferred.py:48: in modify_port
    network_id)
haas/ext/switches/brocade.py:76: in modify_port
    self._set_native_vlan(interface, network_id)
haas/ext/switches/brocade.py:237: in _set_native_vlan
    self._enable_and_set_mode(interface, 'trunk')
haas/ext/switches/brocade.py:145: in _enable_and_set_mode
    self._make_request('POST', url, data=payload)
haas/ext/switches/brocade.py:296: in _make_request
    logger.error('Bad Request to switch. Response: %s' % r.text)
/usr/lib64/python2.7/logging/__init__.py:1175: in error
    self._log(ERROR, msg, args, **kwargs)
/usr/lib64/python2.7/logging/__init__.py:1268: in _log
    self.handle(record)
/usr/lib64/python2.7/logging/__init__.py:1278: in handle
    self.callHandlers(record)
/usr/lib64/python2.7/logging/__init__.py:1318: in callHandlers
    hdlr.handle(record)
/usr/lib64/python2.7/logging/__init__.py:749: in handle
    self.emit(record)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <haas.test_common._FailOnLogWarnings object at 0x36de550>, record = <logging.LogRecord object at 0x3c4a850>

    def emit(self, record):
        if record.levelno >= logging.WARNING:
>           raise LoggedWarningError(record)
E           LoggedWarningError

```

zenhack:
I like (1) better; I'll do that one.

zenhack:
@naved001, pushed a fix, give it a try.

naved001:
The deployment tests are still failing for brocade due to issues not related to this PR. I have made this issue here #748

tests/unit/ext/switches/brocade.py
```diff
@@ -205,15 +207,17 @@ def test_get_mode(self, switch):
             response = switch._get_mode(INTERFACE1)
             assert response == 'trunk'
 
-    def test_apply_networking(self, switch, nic, network):
+    def test_modify_port(self, switch, nic, network):
         # Create a port on the switch and connect it to the nic
         port = model.Port(label=INTERFACE1, switch=switch)
-        nic.port = port
+        port.nic = nic
```
kylehogan: Why the reversal here? (setting the nic of a port rather than the port of a nic)
is it just because this is the 'modify_port' function?

zenhack: Yeah, probably at the time I decided it scanned better this way (but I honestly don't remember my thought process at this point).

naved001:
It's passing all deployment tests. You might wanna rebase it on top of master.

naved001:
@shwsun do you wanna have another look at it to see if your concerns have been addressed?

tests/unit/ext/switches/brocade.py
shwsun: This is simply because in our defined model (see model.py), class `Port` is defined with a field `nic` but not vice versa. Calling `nic.port` is not a relation we captured in our database.

shwsun:
Ok, I just reviewed the commits after my plus one last time. It all looks good. @henn @kylehogan do you have any more opinions on this PR? If not we will probably merge it in soon.

henn:
Looks like we have 2 +1's. Merging!

PR #684, opened 2016-11-14
Migration tests: parametrize fresh_create.

zenhack:
This will allow us to write migration tests that look at specific scenarios; right now we call initial_db to create the objects, which means the database we have to migrate from is fixed for all tests.

In particular, I want to shove some `NetworkingAction`s in the DB as part of testing #683, without a whole lot else to complicate things.

### DIFF
--- a/tests/unit/migrations.py
+++ b/tests/unit/migrations.py
@@ -144,7 +144,24 @@ def get_db_state():
     }],
 ])
 def test_db_eq(filename, make_objects, extra_config):
-    """Verify that each function in fns creates the same database."""
+    """Migrating from a snapshot should create the same objects as a new db.
+
+    `make_objects` is a function that, when run against the latest version
+        of a schema, will create some set of objects.
+    `filename` is the name of an sql dump of a previous database, whose
+        contents were created with the then-current version of `make_objects`.
+    `extra_config` specifies modifications to the haas.cfg under which the
+        test is run. this is passed to `config_merge`.
+
+    The test does the following:
+
+        * Restore the database snapshot and run the migration scripts to update
+          its contents to the current schema.
+        * Create a fresh database according to the current schema, and execute
+          `make_objects`.
+        * Compare the two resulting databases. The test passes if and only if
+          they are the same.
+    """
 
     config_merge(extra_config)
     load_extensions()
--- a/tests/unit/migrations.py
+++ b/tests/unit/migrations.py
@@ -87,14 +87,14 @@ def load_dump(filename):
     upgrade(revision='heads')
 
 
-def fresh_create():
+def fresh_create(make_objects):
     """Create a fresh database, and populate it with an initial set of objects.
 
-    The objects are created via `initial_db`. These objects should be such that
-    a migrated database dump will have the same contents.
+    The objects are created via `make_objects`. These objects should be such
+    that a migrated database dump will have the same contents.
     """
     create_db()
-    initial_db()
+    make_objects()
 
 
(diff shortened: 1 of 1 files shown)


kylehogan:
can you explain a bit how your 'make_objects' works? It seems like it would be adding something in addition to what is in initial_db but I'm not positive.
I'm going to go review Alembic again, but want a starting place for what to look for.

zenhack:
It would be instead of initial db; each migration test would supply its own set of objects to test with., so we can write tests for different scenarios. This would obviously need to be coordinated with the database dump.

kylehogan:
> each migration test would supply its own set of objects to test with

Shouldn't there still be a standard (minimum) set of objects as well though? If the objects are different for each migration test it doesn't seem like the same things would be tested across multiple migrations.

zenhack:
You're still going to be running each migration test against all scripts from when the database dump was made through the current. What's different from test to test is the set of objects that exist in the database before & after, not what migration scripts are tested.

zenhack:
@kylehogan, ping.

kylehogan:
thanks for the reminder. Could you write an addition to the docs that describes how to use `make_objects`? it seems that this is currently set up to use `initial_db`, is this correct? (if so, how would you use something other than initial_db?)

zenhack:
@kylehogan, I added a proper docstring to that function. Note that this is a parametrized test, similar to those in `tests/unit/api/auth.py`. The logic for the test is parametrized over the three arguments to that function. The second argument to `parametrize` is a list of sets of arguments to pass to the function; each one is a test. Right now there's only one, but looking at `tests/unit/api/auth.py` gives a better picture of what this looks like when you have more than one test.

To use something other than initial_db, you'd define another function to use instead, take a database dump, and then add another element to that list that's being passed to parametrize, which would include the path to the dump, the function you defined, and the configuration you used when creating the dump.

It would probably be easier to see what was going on if I'd actually added another such test in this patch. I'm planning on doing this as part of port_revert, which should make it more clear, but I also wanted to keep that pr as small as I can, so doing this separately seemed like a good idea.

kylehogan:
thanks for adding the longer docstring. makes sense! 

+1

zenhack:
@henn, @shwsun, @knikolla, could one of you give this a second review?

shwsun:
@zenhack I have the sense that this PR means instead of using `init_db` to populate database for each and every test case, now we can populate database with different set of objects. Does this mean that we now will have a stack of `init_db`, `switch_op_db` stuff later on? 

And I would guess the value would be that only facing a PR with switches changes will require `switch_op_db` to populate the database, right?

zenhack:
@shwsun, I assume you mean `initial_db`?

This is just for the migration tests, but the point is to give us more flexibility in terms of what objects are used for what tests.

It might be a good idea to factor out the db object population later on, and we can discuss how that should look. The concrete use case I have in mind is for the work on port_revert, which adds a column `type` to NetworkingAction. In this instance:

* The `type` column needs to be added
* Any existing `NetoworkingAction`s need to have their `type` field set to `'modify_port'`

I'd like to write a test that excercises the migration script in the case where there are NetworkingActions in the db. None of our existing fixtures does this, and while we could keep piling stuff on, it would be nice if this could be self-contained -- the test will be easier to understand with just a few objects in the DB. Furthermore, having NetworkingActions in the db is likely to complicate how other tests need to deal with things.

shwsun:
@zenhack I now understand that it is for migration testing. +1

PR #685, opened 2016-11-21
Deployment test doc fixes

zenhack:
@naved001 is working on getting a test/dev environment set up for me to use when making changes to the brocade driver re: #683. He asked about required number of nodes, and I went to find the documentation which I was sure existed re: what was needed to run the deployment tests, only to find that it wasn't documented (except in the source). This patch fixes that, and a couple of other issues I noticed in the process.

//cc @henn, @knikolla, @shwsun

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -126,6 +126,9 @@ and has three fields:
   `"pass"`, defining the information needed to talk to the IPMI
   controller of the node.
 
+The tests currently require at least four nodes to be specified in
+`site-layout.json`, each of which must have at least one nic connected
+to the switch.
 
 [1]: http://pytest.org/
 [2]: https://pypi.python.org/pypi/pytest-cov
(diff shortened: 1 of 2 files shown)


docs/testing.md
```diff
@@ -125,6 +126,9 @@ and has three fields:
   `"pass"`, defining the information needed to talk to the IPMI
```
shwsun: I think we should use format of `<switches>` to describe those fields right? I see some same inconsistency in `network-driver.md` as well. And also, line 119  should say "all of them are strings".

shwsun:
Looks good, will +1 once these minor things are fixed

docs/testing.md
zenhack: Line 119 is in fact correct English.

The ipmi thing is out of date; when we made the OBM drivers pluggable this changed. So I just updated that to describe the current behavior.

I'm not sure what you mean when you say "use the format of `<switches>`"

zenhack: @shwsun, ping.

shwsun: Never mind, I saw some inconsistency in `docs/network-drivers.md` and was wondering which format is correct. +1

zenhack:
Ok, since @shwsun has added his +1 I'm going to merge.

PR #686, opened 2016-11-26
Run network-related deployment tests in travis

zenhack:
Might as well; we have the mock drivers we need.

//cc @henn, @knikolla, @shwsun

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,4 +10,5 @@ dist
 *.db
 /haas.cfg
 /testsuite.cfg
+/site-layout.json
 /.cache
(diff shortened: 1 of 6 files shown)


shwsun:
@zenhack I like it a lot! I think it is great that we can run tests on mock switches, and in the long run we should do this for all of our mock drivers.

@henn @SahilTikale @knikolla Anyone interested in giving one more review?

zenhack:
@henn, I'm assigning you as a reviewer. This PR isn't terribly complex, but it's open for almost two months. Please have a look.

zenhack:
@shwsun I'm interpreting your comment as being a +1, and now that @knikolla has approved this, we ought to be able to merge. Am I interpreting that correctly?

PR #689, opened 2016-12-05
Fix bogus i18n call.

zenhack:
The `_` function is a common alias used when doing internationalization.
We don't actually do i18n yet, and there's no alias defined/imported in
this file, so if this statement gets executed, it raises a NameError.

This was introduced in commit 16016078453913918dcec208c2c000a7f935a488,
though it's not clear to me why.

@matsuuran, that commit has your name on it; any idea what happened there?

@shwsun, @knikolla, @henn, reviews appreciated.

### DIFF
--- a/haas/network_allocator.py
+++ b/haas/network_allocator.py
@@ -93,8 +93,8 @@ def set_network_allocator(network_allocator):
     """
     global _network_allocator
     if _network_allocator is not None:
-        sys.exit(_("Fatal Error: set_network_allocator() called twice."
-                   "Make sure you don't have conflicting extensions loaded."))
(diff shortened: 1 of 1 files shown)


zenhack:
@shwsun, @henn, could one of you give this a second review?

zenhack:
@shwsun, I just assigned you as a second reviewer. There's almost nothing to this, but it's been waiting on a second signoff for over a month. Would you give it a second review so we can finally get this merged?

shwsun:
@zenhack Sure, I will take a look at it right now. 
Sorry about the out of date - I have been assigned to another project and almost got no time to work on HIL.

zenhack:
Great, I'll go ahead and merge.

PR #696, opened 2016-12-29
Pending networking action snapshot

zenhack:
I'm going to want this in place for the port_revert stuff, and it seemed self-contained enough to split it into its own pr.


shwsun:
@zenhack I assume you need to fix the travis ci failure before we start the review process right?

zenhack:
@shwsun, yes, but see also #697, which is the the cause of the failures. We're going to need to figure that one out before we merge much of anything that adds migration tests.

zenhack:
I rebased this on top of the fix, tests are passing now.

PR #698, opened 2017-01-04
Fix #697

### LINKED ISSUE #697: Tests are polluting each other's sets of extensions
Somehow the tests are being polluted by each other; on #696, the migration test I added is having extensions it doesn't specify loaded, which breaks the test. I've managed to fix this on a debugging branch by running the migration tests in their own travis instance:

https://travis-ci.org/zenhack/hil/builds/187775909

(We may want to split more things out that way, since it would allow things to finish faster. We probably don't need to run the apache setup with sqlite, and it would be nice to speed things up, but that's another issue).

I *suspect* that some of the tests are importing extensions at top level, which are safe for the modules they're in, but that these are getting processed when the tests are collected, before separate processes are forked off to actually run the tests. If this is the case, a solution may be to just move all of these imports in fixtures.

zenhack:
This patch avoids loading things under haas.ext at top level, since that
happens during test collection, and thus pollutes the environment of
tests that don't use those extensions.

It also includes a regression test.

//cc @shwsun, @knikolla, @henn

### DIFF
--- a/tests/unit/ext/auth/database.py
+++ b/tests/unit/ext/auth/database.py
@@ -5,13 +5,23 @@ from haas.flaskapp import app
 from haas.model import db
 from haas.errors import AuthorizationError
 from haas.rest import init_auth, local
-from haas.ext.auth.database import User, user_create, user_delete, \
-    user_add_project, user_remove_project
 import flask
 import pytest
 import unittest
 
 fail_on_log_warnings = pytest.fixture(autouse=True)(fail_on_log_warnings)
+
+
+@pytest.fixture
+def dbauth():
+    from haas.ext.auth import database
+    return database
+
+
+class DBAuthTestCase(unittest.TestCase):
+
+    def setUp(self):
+        self.dbauth = dbauth()
 
 
 @pytest.fixture
@@ -35,15 +45,15 @@ def configure():
 
 
 @pytest.fixture
-def initial_db(request):
+def initial_db(request, dbauth):
     fresh_database(request)
     with app.app_context():
-        alice = User(label='alice',
-                     password='secret',
-                     is_admin=True)
-        bob = User(label='bob',
-                   password='password',
-                   is_admin=False)
+        alice = dbauth.User(label='alice',
+                            password='secret',
+                            is_admin=True)
+        bob = dbauth.User(label='bob',
+                          password='password',
+                          is_admin=False)
 
         db.session.add(alice)
         db.session.add(bob)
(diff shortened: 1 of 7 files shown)


zenhack:
@kylehogan, @knikolla, would the two of you do a review? @henn assigned you 5 days ago, but I haven't heard anything.

tests/unit/ext/auth/database.py
```diff
@@ -15,6 +13,18 @@
 
 
 @pytest.fixture
+def dbauth():
+    from haas.ext.auth import database
```
knikolla: What is the benefit of having this here instead of at the top as a module level import?

knikolla: Nevermind, answered my own question. The scope of this is to have a 'clean' room for extensions.

tests/unit/ext/auth/database.py
```diff
@@ -224,23 +238,25 @@ class TestUserModel(ModelTest):
...
-    (user_delete, ['bob']),
-    (user_add_project, ['bob', 'runway']),
-    (user_remove_project, ['alice', 'runway']),
+    ('user_create', ['charlie', '1337']),
+    ('user_create', ['charlie', '1337', False]),
+    ('user_create', ['charlie', '1337', True]),
+    ('user_delete', ['bob']),
+    ('user_add_project', ['bob', 'runway']),
+    ('user_remove_project', ['alice', 'runway']),
 ]
```
kylehogan: why did the quotes around `user_create` etc. become necessary here?

tests/unit/model.py
```diff
@@ -23,7 +23,6 @@
 
 from haas.model import *
 from haas import config
-from haas.ext.obm.ipmi import Ipmi
```
kylehogan: is there a way we could still import this once and have it be used by the Nic, Node, and NetworkingAction tests?

tests/unit/ext/auth/database.py
zenhack: Because the function isn't in scope at the top level. It can't be without doing a global import, and that's what this patch is removing due to issues of contamination. Instead, the tests now use `getattr` to get the actual function.

tests/unit/model.py
zenhack: I suppose I could write a subclass of ModelTest that does the import and attaches it to self (and then have the tests that use ipmi inherit from it), or rig up a pytest fixture, but neither of these actually save any code (both are longer than just doing the import in each place), and I think they obscure what's going on. Possible implementations:

```python
# fixture version. Also need to add this fixture to each of the tests that use it.
@pytest.fixture()
def Ipmi():
    from haas.ext.obm import Ipmi as ret
    return ret
```

```
# class version. Each test that uses this now needs to inherit from it, instead of ModelTest
class IpmiModelTest(ModelTest):
    def setUp(self):
        from haas.ext.obm.ipmi import Ipmi
        self.Ipmi = Ipmi
```

My inclination is to just do what we're doing now; I can't think of a way to dedup this that I think is actually better than doing the import multiple times. @knikolla, thoughts?

henn:
[PEP8 says](https://www.python.org/dev/peps/pep-0008/#imports) 
> Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.

I'm ok with making an exception to this in our dev docs, but want to be sure we all agree on that.

Interestingly, I ran into this last week when debugging #703. I originally considered moving the import of `haas.server` into serve()/serve_networks(), since we don't need to load flask/sqlachemy/... for most all of the CLI, but decided against it because of PEP8.

zenhack:
@henn, the whole extension mechanism really is different; by their very nature they aren't imported at parse time, since in prod it happens based on the config file. My reading is that that sentence really only applies to "typical" i.e. top level imports; if you're doing post-startup imports you're doing something semantically different, and it really doesn't apply. And anyway, the pep8 tool doesn't complain.

Issue #303, opened 2014-10-20
Interface to set a node's next boot device


henn:
We should have an interface that allows a user to set the next boot device, likely either PXE or local disk. API should leave open the possibility of other boot sources (like a virtual disk drive).

Perhaps we could even allow them to set it persistently, assuming IPMI supports that? If that happened, we should likely reset that on node delete,

zenhack:
What's our current thinking on this? Now that HIL has seen a bit more use, has anyone found themselves actually wanting this? Per my comments back when this was opened, we can't actually change the boot order for more than the next boot, and calling shutdown from within the OS wil bypass our setting here anyway.

zenhack:
@henn, thoughts?

naved001:
It seems like IPMI supports setting the boot device persistently. 

> [naved001@haas-beta ~]$ ipmitool -I lanplus -U $USER -P $PASSWORD -H $HOSTNAME chassis bootdev disk options=help
Legal options settings are:
	help:	print this message
	valid:	Boot flags valid
	**persistent: Changes are persistent for all future boots**
	efiboot:	Extensible Firmware Interface Boot (EFI)
	clear-cmos:	CMOS clear
	lockkbd:	Lock Keyboard
	screenblank:	Screen Blank
	lockoutreset:	Lock out Resetbuttons
	lockout_power:	Lock out (power off/sleep request) via Power Button
	verbose=default:	Request quiet BIOS display
	verbose=no:	Request quiet BIOS display
	verbose=yes:	Request verbose BIOS display
	force_pet:	Force progress event traps
	upw_bypass:	User password bypass
	lockout_sleep:	Log Out Sleep Button
	cons_redirect=default:	Console redirection occurs per BIOS configuration setting
	cons_redirect=skip:	Suppress (skip) console redirection if enabled
	cons_redirect=enable:	Suppress (skip) console redirection if enabled

I should test this to make sure it works as expected.

zenhack:
I vaguely remember testing that and finding that it didn't work, but that may have been with the old sun blades, in which case it might make sense to try it with newer gear.

henn:
#699 fixes this!

PR #699, opened 2017-01-05
Use Ipmitool to set a node's next boot device persistently

naved001:
This PR fixes issue #303. 

I have tested this on Cisco and supermicro servers, and it works as expected. I could force the machines to boot from disk, pxe or just go to BIOS persistently. 

Let me know what you guys think. @henn @zenhack @SahilTikale


docs/rest_api.md
```diff
@@ -416,6 +416,31 @@ Authorization requirements:
 Power cycle the node named `<node>`, and set it's next boot device to
 PXE. If the node is powered off, this turns it on.
 
+#### node_set_bootdev
+
+`POST /node/<node>/dev/<dev>/set_bootdev`
```
zenhack: Suggestion for a more rest-ish path: `/node/<node>/bootdev`. The general principle being that URLs are resources/objects, not procedures.

docs/rest_api.md
```diff
@@ -416,6 +416,31 @@ Authorization requirements:
...
+
+Sets the nodes next boot device persistently
+
+device <dev> can be set to:
+
+  none  : Do not change boot device order
```
zenhack: This doesn't seem like a coherent option; if you were going to us this, why make the API call at all?

docs/rest_api.md
```diff
@@ -416,6 +416,31 @@ Authorization requirements:
 Power cycle the node named `<node>`, and set it's next boot device to
 PXE. If the node is powered off, this turns it on.
 
+#### node_set_bootdev
+
+`POST /node/<node>/dev/<dev>/set_bootdev`
+
+Sets the nodes next boot device persistently
+
+device <dev> can be set to:
```
zenhack: This should probably be a bullet list, e.g:

* "none": Do not change boot device order
* "disk": Force boot from default Hard-drive
* ... and so on

Worth famliarizing yourself with markdown formatting rules, if you haven't already.

docs/rest_api.md
```diff
@@ -416,6 +416,31 @@ Authorization requirements:
...
+
+  none  : Do not change boot device order
+  
+  pxe   : Force PXE boot
```
zenhack: The word "force" really doesn't need to be here; you're just changing a setting, there's not really any kind of "hard override" semantics here. And of course it's redundant with every other item in the list (except the one I suggested removing).

docs/rest_api.md
```diff
@@ -416,6 +416,31 @@ Authorization requirements:
...
+  
+  safe  : Force boot from default Hard-drive, request Safe Mode
+  
+  diag  : Force boot from Diagnostic Partition
+  
+  cdrom : Force boot from CD/DVD
+  
+  bios  : Force boot into BIOS Setup
+  
+  floppy: Force boot from Floppy/primary removable media
```
zenhack: Questions:

1. Are all of these guaranteeded to be supported by an particular version of IPMI? we should be able to succicntly specify what requirements are for hardware to work with the IPMI driver.
2. Do we really want to allow access to all of these? I don't even know off-hand what exactly is meant by "diagnostic partition".
3. Should we actually enumerate the options in the spec? It would be nice for this interface to be portable to non-IPMI obms, and it would be nice not to have the "floppy" option be part of the stable API once we have one -- it's not clear it's relevant *now* and we want to be able to change what we support to match hardware. Maybe it would be better to have a way to fetch the available options, which may vary based on driver or hardware? @henn, @knikolla, @shwsun, curious to your thoughts on that point.

haas/api.py
```diff
@@ -263,6 +263,19 @@ def node_power_off(node):
     node.obm.power_off()
 
 
+@rest_call('POST', '/node/<node>/dev/<dev>/set_bootdev', Schema({
+    'node': basestring, 'dev': basestring,
```
zenhack: *if* we are going to specify a fixed set of supported boot devices as you've documented (which perhaps we shouldn't; see discussion above), then we can tighten up the schema for `dev`; instead of basestring you can just specify `Or('pxe', 'disk', ...)` (after importing Or from schema of course). Regardless, we need to check that the value is legal *somewhere*.

haas/cli.py
```diff
@@ -499,6 +499,18 @@ def node_power_off(node):
 
 
 @cmd
+def node_set_bootdev(node, dev):
+    """
+    Sets <node> to boot from <dev> persistenly
```
zenhack: Blank line between the summary and the rest of the docstring; see PEP257.

haas/ext/obm/ipmi.py
```diff
@@ -87,6 +87,12 @@ def power_off(self):
             raise OBMError('Could not power off node %s', self.label)
 
     @no_dry_run
+    def set_bootdev(self, dev):
+        if self._ipmitool(['chassis', 'bootdev', dev,
+                          'options=persistent']) != 0:
+            raise OBMError('Coult not set boot device')
```
zenhack: s/Coult/Could/

docs/rest_api.md
naved001: Ipmitool overrides the boot order in BIOS. For example, if I set boot device to `bios`, the node will always boot into the BIOS regardless of the boot order, or if I set it to `disk` it will always boot from disk even if you manually set the boot order in BIOS. 

Setting boot device to `none` lets the node follow the boot order in BIOS.

docs/rest_api.md
naved001: I got this from the [ipmitool manpage](https://linux.die.net/man/1/ipmitool). They are using the word force to imply that it overrides the boot order from BIOS (as explained above).

docs/rest_api.md
naved001: I'll take care of this.

docs/rest_api.md
naved001: okay, I'll make it a bullet list once we decide what supported boot devices we want (or if we want this list at all).

docs/rest_api.md
zenhack: I see, that makes more sense.

I still think we ought to make the API less IPMI-centric. In particular, having this set of options be defined here strikes me as a bad idea; they should be obm-specific. If we move this to a section of docs that specifically describes the IPMI backend, we can also just *say* "these have the same meaning as the options for ipmitool."

naved001: Right. I agree, the API should be less IPMI-centric. I'll change this in the next commit.

docs/rest_api.md
henn: I think we should limit it to disk and pxe, and disallow the rest.

naved001:
@zenhack I have made the changes that you requested.

docs/rest_api.md
```diff
@@ -416,6 +416,12 @@ Authorization requirements:
 Power cycle the node named `<node>`, and set it's next boot device to
 PXE. If the node is powered off, this turns it on.
 
+#### node_set_bootdev
+
+`Put /node/<node>/bootdev`
+
+Sets the nodes next boot device persistently
```
henn: `node's`

docs/rest_api.md
```diff
@@ -416,6 +416,12 @@ Authorization requirements:
 Power cycle the node named `<node>`, and set it's next boot device to
 PXE. If the node is powered off, this turns it on.
 
+#### node_set_bootdev
```
henn: We should document what the valid body values (pxe, disk, none) mean. It's ok to say stuff like "For IPMI devices, ..."

zenhack: All of the other driver-specific stuff is documented separately, c.f. channel formats. We ought to do the same with obm-specific things like available boot devices. And yes, each one needs to be described.

haas/api.py
```diff
@@ -264,7 +264,7 @@ def node_power_off(node):
 
 
 @rest_call('PUT', '/node/<node>/bootdev', Schema({
-    'node': basestring, 'bootdev': Or('pxe','disk','none','bios'),
+    'node': basestring, 'bootdev': Or('pxe', 'disk', 'none'),
```
zenhack: If we're going to have this list be obm-specific, it needs to be specified elsewhere. this probably means going back to `basestring` and doing the validation inside the call to `node.obm.set_bootdev`.

docs/rest_api.md
henn: If we had more than one or two OBM drivers (like we do for network drivers), I'd agree with you, though for now I'm comfortable inlining these until there are more.

@zenhack - what do you think?

henn:
@naved001 - looks good. What do you think about writing a deployment test for these?

docs/rest_api.md
zenhack: I suppose it's fine; it is at least in its own section.

zenhack:
This mostly looks good, but as @henn suggested, some tests are needed. Beyond just the deployment tests I would ask that you add a few that test the basic API mechanism, using the mock driver. It might make sense to actually give that a fixed list of legal devices so we can test cases where the driver rejects the value without needing real hardware to do so. Also, you should add authentication tests; most of these are currently in `tests/unit/api/auth.py`. Use of the `@no_dry_run` decorator may make it possible to test much of the ipmi driver even without a proper deployment environment.

**Also** note that we just merged #680, which broke some things; I've opened a PR to revert it, and we'll get a fixed version in soon, but (once the revert is in) I will want you to merge master back into this before approving it -- lesson learned.

haas/ext/obm/ipmi.py
```diff
@@ -87,6 +87,15 @@ def power_off(self):
             raise OBMError('Could not power off node %s', self.label)
 
     @no_dry_run
+    def set_bootdev(self, dev):
+        valid_bootdevices = ['disk', 'pxe', 'none']
+        if dev not in valid_bootdevices:
+            raise OBMError('Invald boot device')
```
knikolla: Create a new exception OBMValidationError, so you can distinguish between HIL failing to set the boot device (500), and the boot device not being valid (400).

zenhack:
Quoting Kristi Nikolla (2017-01-26 16:56:21)

>    @knikolla commented on this pull request.
>      __________________________________________________________________
>
>    In [1]haas/ext/obm/ipmi.py:
> > @@ -87,6 +87,15 @@ def power_off(self):
>              raise OBMError('Could not power off node %s', self.label)
>
>      @no_dry_run
> +    def set_bootdev(self, dev):
> +        valid_bootdevices = ['disk', 'pxe', 'none']
> +        if dev not in valid_bootdevices:
> +            raise OBMError('Invald boot device')
>
>    Create a new exception OBMValidationError, so you can distinguish
>    between HIL failing to set the boot device (500), and the boot device
>    not being valid (400).

Good catch. OBMError is a subclass of ServerError, which indicates a
server-side problem. Anything that the client is responsible for should
raise some error that is a subclass of APIError.

haas/ext/obm/ipmi.py
naved001: How about I throw a `BadArgumentError` (400) which we already have instead of making a new one?

zenhack:
Quoting Naved Ansari (2017-01-27 10:51:41)

>      @no_dry_run
> +    def set_bootdev(self, dev):
> +        valid_bootdevices = ['disk', 'pxe', 'none']
> +        if dev not in valid_bootdevices:
> +            raise OBMError('Invald boot device')
>
>    How about I throw a BadArgumentError (400) which we already have
>    instead of making a new one?

Works for me. This is what we do for illegal channel specifications too.

naved001:
@zenhack 
> add a few that test the basic API mechanism, using the mock driver. It might make sense to actually give that a fixed list of legal devices so we can test cases where the driver rejects the value without needing real hardware to do so.

The API doesn't reject anything, only the driver does. So I am not sure what would be the point of testing the mock driver. But I have written a unit test where the ipmi driver rejects the value and it doesn't need real hardware either because it just rejects it. 

> Also, you should add authentication tests; most of these are currently in tests/unit/api/auth.py

Done. 
> Use of the @no_dry_run decorator may make it possible to test much of the ipmi driver even without a proper deployment environment.

The `@no_dry_run` decorator just bypasses the function in the driver which defeats the purpose of testing the driver. So in the unit test that I wrote, I am not using `config_testsuite()` (because that has `devel` set to `True`, and I couldn't change it). I just added whatever configuration I required. 

Feel free to correct me if I got something wrong. And I'll write the deployment tests asap.

naved001:
@zenhack @henn I have added the required tests. Let me know if you have any questions.

tests/deployment/ipmi.py
```diff
@@ -67,3 +68,10 @@ def test_node_power_off(self):
         nodes = self.collect_nodes()
         for node in nodes:
             api.node_power_off(node.label)
+
+    def test_node_set_bootdev(self):
+        nodes = self.collect_nodes()
+        for node in nodes:
+            api.node_set_bootdev(node.label, 'pxe')
+            api.node_set_bootdev(node.label, 'disk')
+            api.node_set_bootdev(node.label, 'none')
```
henn: It might be good to have at least one or two negative tests, checking that proper exceptions are thrown.

docs/rest_api.md
```diff
@@ -416,6 +416,20 @@ Authorization requirements:
...
+
+`PUT /node/<node>/boot_device`
+
```
henn: Need to document that the body consists of JSON with a `bootdev` argument.

haas/api.py
```diff
@@ -263,6 +263,19 @@ def node_power_off(node):
...
+    'node': basestring, 'bootdev': basestring,
+}))
+def node_set_bootdev(node, bootdev):
+    auth_backend = get_auth_backend()
+    node = _must_find(model.Node, node)
+    if node.project is None:
+        auth_backend.require_admin()
+    else:
+        auth_backend.require_project_access(node.project)
+    node.obm.set_bootdev(bootdev)
```
henn: One idea that Naved had is to validate the valid bootdev here and not in the driver by adding a callout like `validatte_bootdev` or something.

I really like that, as it keeps invalid values out of the lower depths.

Another option I thought of would be to allow extensions to modify bootdev, but that doesn't work because you could have multiple drivers that support different bootdevs, and a valid bootdev on one (like dell_idrac) wouldn't necessarily work on another (like ipmi).

zenhack: I like having the api call do most of the lifting re: validation. What if we added an abstract method to Obm modeled after the network allocators' `is_legal_channel_for`?

naved001: > I really like that, as it keeps invalid values out of the lower depths.

I know I suggested that, but now that I think of it, a function like `validate_bootdev` would still be a part of the driver. It's just splitting the function into 2 parts  (one to do the validation, the other to set the boot device).

tests/unit/ext/obm/ipmi.py
```diff
@@ -0,0 +1,73 @@
...
+import pytest
+import haas
+from haas import model, server, api
+from haas.test_common import config, config_set, fresh_database, \
+    fail_on_log_warnings, with_request_context
+
+
+@pytest.fixture
+def configure():
+    config_set({
```
naved001: Good point. I changed it.

naved001:
The deployment test now runs both, positive and negative, tests (catches `BadArgumentError` and `OBMError`). 
The unit test is only limited to testing the case where the driver rejects the boot device. I am not sure how to have a case where the driver accepts the boot device because then it will try to actually set the boot device which wouldn't work. (Do I need to have this case? the deployment test has this though)

naved001:
@zenhack @henn ping

docs/rest_api.md
```diff
@@ -416,6 +416,29 @@ Authorization requirements:
...
+Sets the node's next boot device persistently
+
+The request body consists of JSON with a `bootdev` argument:
+
+Request body:
+    {
+
+    'bootdev': <boot device>
+
+    }
```
zenhack: * Use double quotes; single quotes aren't valid JSON.
* Indent the fields relative to the braces
* I don't think you need the extra blank lines between the braces and the field.

zenhack:
Sorry for the delay. With the exception of the minor formatting issues I just pointed out, I'm happy. Fix those and I'll approve this.

docs/rest_api.md
naved001: Thanks! I have made these changes, now the tests are failing due to some other reason with our TravisCI.

zenhack:
Once we figure out the CI issues, I'll approve this.

naved001:
@zenhack the tests are passing now because #730 got merged
@henn ping

zenhack:
@naved001, great, +1.

naved001:
@zenhack: @henn and I have made some changes; you might want to re-review this

haas/ext/obm/ipmi.py
```diff
@@ -86,10 +88,13 @@ def power_off(self):
...
+    def require_legal_bootdev(self, dev):
+        if dev not in self.valid_bootdevices:
+            raise BadArgumentError('Invald boot device')
+
     @no_dry_run
     def set_bootdev(self, dev):
-        valid_bootdevices = ['disk', 'pxe', 'none']
-        if dev not in valid_bootdevices:
+        if dev not in self.valid_bootdevices:
             raise BadArgumentError('Invald boot device')
```
zenhack: I might just call require_legal_bootdev here, to make it explicit that this is the same thing. Might otherwise be a bit confusing to the reader whether this is meant to check something else.

zenhack:
I made one minor comment on the new patch but am otherwise happy.

You'll need to pull master back in before the tests will pass, because of the CI fixes.

naved001:
@zenhack made the changes you suggested. 

> You'll need to pull master back in before the tests will pass, because of the CI fixes.

I didn't have to pull master in this to get the tests to pass. Any idea why?

zenhack:
Not sure. I'd actually appreciate you pulling in master anyway; we've once or twice had test breakage on master because even though there weren't merge conflicts per se, there were changes that needed to be reconciled, so even though the branches passed individually, they failed once merged. Would be good to be careful of that.

henn:
Master appeared to merge cleanly. Interestingly, it seems that github no longer shows diffs for commits that are in both branches, as the commit list only lists my "merge" commit and not the individual master commits!!! Yay!

PR #702, opened 2017-01-14
Updated the Brocade driver to enable the switchport when setting mode

knikolla:
This step had to previously be done by the administrator through the
commandline interface by issuing the command ``switchport``. It
is now automated by the HIL driver by issuing an API call to
enable switching on the interface before the driver sets the port.

### DIFF
--- a/tests/unit/ext/switches/brocade.py
+++ b/tests/unit/ext/switches/brocade.py
@@ -274,3 +274,14 @@ class TestBrocade(object):
             assert mock.called
             assert mock.call_count == 1
             assert mock.request_history[0].text == TRUNK_REMOVE_VLAN_PAYLOAD
+
+    def test_construct_url(self, switch):
+        assert switch._construct_url('1/0/4') == (
+            'http://example.com/rest/config/running/interface/'
+            'TenGigabitEthernet/%221/0/4%22'
+        )
+
+        assert switch._construct_url('1/0/4', suffix='mode') == (
+            'http://example.com/rest/config/running/interface/'
+            'TenGigabitEthernet/%221/0/4%22/switchport/mode'
+        )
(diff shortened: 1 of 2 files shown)


kamfonik:
+1

henn:
@knikolla - thanks for your PR! Were you able to test this on a real brocade switch?

knikolla:
@henn no I haven't. @naved001 can you please test this?

naved001:
@knikolla sure.
 @henn is the brocade switch in the PRB cluster available to use?

shwsun:
@naved001 Yes, you can go ahead and test it

naved001:
Need to change `POST` to `PUT`. After changing that it works! Thanks @knikolla!

naved001:
Tested this on a real switch and it works. Can you review this @zenhack 
Also fixes #718

tests/unit/ext/switches/brocade.py
```diff
@@ -219,9 +221,9 @@ def test_apply_networking(self, switch, nic, network):
             switch.apply_networking(action_native)
 
             assert mock.called
-            assert mock.call_count == 3
-            assert mock.request_history[0].text == TRUNK_PAYLOAD
-            assert mock.request_history[2].text == TRUNK_NATIVE_PAYLOAD
+            assert mock.call_count == 4
+            assert mock.request_history[1].text == TRUNK_PAYLOAD
```
zenhack: Should we still be checking `request_history[0]`? I'm fuzzy on exactly how this test works, but I would expect adding a new request we should check it like the others.

knikolla:
@zenhack updated to resolve your comments.

zenhack:
Great. Have we successfully run the deployment tests? Once I get an ACK on that I'll +1.

naved001:
@zenhack The tests pass with the fix from #742

zenhack:
@naved001, great! I left a comment there, once we've got that merged I'll approve this one.

naved001:
@zenhack #742 is merged.

zenhack:
Cool, merging.

PR #704, opened 2017-01-20
Document haas.ext.auth.null extension in example config

henn:


kamfonik:
do you think an example database endpoint would be useful?

henn:
@kamfonik - we currently have examples for sqlite and postgres. Was there a particular example you were thinking of?

kamfonik:
@henn whoops, missed the fact that the examples were above, I was looking near the auth database setting in the diff.

Looks good +1

henn:
Given that this is akin to documentation, I am merging this with one approval.

PR #711, opened 2017-01-26
Revert "adding tpm metadata for nodes"

zenhack:
Reverts CCI-MOC/hil#680

See the discussion there; this broke our CI, so I'd like to back it out and ask @kylehogan to fix the problems before we merge it again.

//cc @henn, @shwsun, @knikolla


kylehogan:
What do I need to do to fix this? My understanding is that this pull request will revert to before mine was merged (if so, +1) and then I will need to merge master into my branch and fix the errors

kylehogan:
actually after looking at it it might just be the import statements for Ipmi in the unit test for model.py. The two failing tests say that Ipmi is not defined and the only thing that seems to have changed is the location of the import statement.

zenhack:
Yeah, those got removed as part of #698; should be pretty trivial to fix.

kylehogan:
So do we want to revert it then make the changes or should I make a new PR with the fix? I'm not sure which way is standard

zenhack:
I don't think we have a standard; I figured it would be good to have master not broken asap, but if you can get a fix out quickly we can just merge that.

henn:
Thanks for posting this!

I'd say that if Kyle can fix this by mid-day tomorrow, then that would be preferable over reverting this and causing merge confusion. If not, then I'll merge this tomorrow.

zenhack:
Quoting Jason Hennessey (2017-01-26 18:18:40)

>    Thanks for posting this!
>
>    I'd say that if Kyle can fix this by mid-day tomorrow, then that would
>    be preferable over reverting this and causing merge confusion. If not,
>    then I'll merge this tomorrow.

Sounds good.

kylehogan:
@henn @zenhack there's a fix PR now! #712 
it was really trivial - one missing import statement.

zenhack:
Commented. I'm going to close this, since it seems like it will be unnecessary.

PR #712, opened 2017-01-27
fixing missing ipmi import introduced by tpm-metadata

kylehogan:


tests/unit/model.py
```diff
@@ -97,7 +97,8 @@ def sample_obj(self):
 class TestMetadata(ModelTest):
 
     def sample_obj(self):
-        node = Node(label='node-99',
+        from haas.ext.obm.ipmi import Ipmi
+        return Node(label='node-99',
```
henn: I think you meant `node = Node(...` instead of `return Node(...`.

Which brings up the question - shouldn't this test be failing?

zenhack: These tests are just a basic "if I create one of these, does it explode?" Skipping the line below shouldn't cause a test failure, but you are effectively skipping the test.

zenhack:
Except for the problem @henn pointed out, this looks good to me. Fix that and I'll approve.

zenhack:
Looks good, +1. (I would hit approve if I could find it on the mobile page...)

henn:
I merged this into master to fix the tests.

@kylehogan - could you look into why the test didn't fail even though it was receiving a strange type back? Are they being run properly?

zenhack:
Quoting Jason Hennessey (2017-01-27 16:44:09)

>    I merged this into master to fix the tests.
>
>    [1]@kylehogan - could you look into why the test didn't fail even
>    though it was receiving a strange type back? Are they being run
>    properly?

@henn, have a look at the implementation of ModelTest in
haas/test_common.py. The tests do two things;

1. Make sure the object's __repr__ doesn't explode
2. Make sure committing the object to the db doesn't explode.

Neither of these things should be affected by returning the wrong
database model, so a false-success here would have been expected. In
this case it just ends up being a duplicate of the Node test.

Issue #716, opened 2017-02-02
 Unique HTTP status code required to reflect different different error states.


SahilTikale:
Currently different errors listed in [errors.py](https://github.com/CCI-MOC/hil/blob/master/haas/errors.py) return the same status code. eg. for errors `DuplicateError` and `BlocedError` return code `409`. We need to have unique status code for each error state or add new error states to reflect new types of errors.

PR #717, opened 2017-02-02
Replaces 409 with unique appropriate codes in errors.py

SahilTikale:
Addresses issue #716

### DIFF
--- a/haas/errors.py
+++ b/haas/errors.py
@@ -80,7 +80,9 @@ class ProjectMismatchError(APIError):
     """An exception indicating that the resources given don't belong to the
     same project.
     """
-    status_code = 409  # Conflict
+    status_code = 412  # was 409. 412 is appropriate here.
+    # For description read,
+    # http://www.restpatterns.org/HTTP_Status_Codes/412_-_Precondition_Failed
 
 
 class AuthorizationError(APIError):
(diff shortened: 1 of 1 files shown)


naved001:
looks good +1

haas/errors.py
```diff
@@ -80,7 +80,9 @@ class ProjectMismatchError(APIError):
     """An exception indicating that the resources given don't belong to the
     same project.
     """
-    status_code = 409  # Conflict
+    status_code = 412  # was 409. 412 is appropriate here.
```
zenhack: In general, don't describe the history in comments. This would be more reasonable thing to mention in a commit message.

Also, this really *isn't* an appropriate use of the 412 status code, which has a very specific use. From the url you supply below:

> The precondition **given in one or more of the request-header fields** evaluated to false when it was tested on the server. This response code allows the client to place preconditions on the current resource metainformation (header field data) and thus prevent the requested method from being applied to a resource other than the one intended.

(emphasis mine). It's part of a mechanism defined in:

https://tools.ietf.org/html/rfc7232

haas/errors.py
```diff
@@ -92,7 +94,10 @@ class BlockedError(APIError):
     some other change.  For example, deletion is blocked until the components
     are deleted, and possibly until the dirty flag is cleared as well.
     """
-    status_code = 409  # Conflict
+    status_code = 423  # was 409,
```
zenhack: Similarly, this is part of WEBDAV, and has some specific constraints on its use:

> The 423 (Locked) status code means the source or destination resource of a method is locked. **This response SHOULD contain an appropriate precondition or postcondition code, such as 'lock-token-submitted' or 'no-conflicting-lock'.**

(Again, emphasis mine).

The way we use BlockedError isn't really the same as a resource being "locked." 409 Conflict is still the best thing I've seen; you can only get so descriptive with HTTP status codes.

henn:
It sounds like the consensus is to use the RFC status codes, but pay more attention to the status text.

This also means that we need to standardize the error text, so that libraries like Sahil's (#617) can interpret this in a standard way.

PR #727, opened 2017-02-22
CLI: switch_register now prints correct message for invalid number of arguments

### LINKED ISSUE #726: Nexus driver: wrong error msg and documentation fix
This bug is found out during a process of using Nexus driver. As mentioned in `network_driver.md`, registering cisco nexus switches required a extra field `dummy_field`. However, doing so w/out the extra field will not throw an error msg about `dummy_field` but something like `Fatal Error: set_auth_backed() called twice. Make sure you don't have conflicting extensions loaded.`

Another related problem is that we should specify how to use the `dummy_vlan`, i.e. in `network_driver.md` we should show that after configure the dummy vlan, user need to do 
`$ haas switch_register cisco-x nexus IP_ADDR USER_NAME PASSWD dummy_vlan`

The previous issue identified the problem is #723

naved001:
This fixes #726 
So now if you miss an argument, `switch_register` will print the correct message for brocade, dell, and nexus. 
@shwsun can you review this?

### DIFF
--- a/haas/cli.py
+++ b/haas/cli.py
@@ -592,9 +592,9 @@ def switch_register(switch, subtype, *args):
                 "dummy_vlan": args[3]}
         else:
             sys.stderr.write('ERROR: subtype ' + subtype +
-                               ' requires exactly 4 arguments\n'
-                               '<hostname> <username> <password>'
-                               '<dummy_vlan_no>\n')
+                             ' requires exactly 4 arguments\n'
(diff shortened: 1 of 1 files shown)


shwsun:
@henn  @zenhack This fixes the issue with our Nexus driver, do you have some time to take a look? (It is a really small PR)

zenhack:
Looks good, merging.

Issue #729, opened 2017-02-27
Travis CI build failure


naved001:
Here are the details of my PR which was previously passing all the tests. https://travis-ci.org/CCI-MOC/hil/jobs/205965436

@henn @zenhack @knikolla

naved001:
@SahilTikale the issue I was talking about.

zenhack:
Look at the code, I'm actually a bit confused as to why it *was* working, or *is* working on master. It's complaining that additional_db is being called outside of a request context (and thus without access to db.session), which I would expect to cause problems. And indeed our list of fixtures has additonal_db before with_request_context.

@shwsun, your name is on the commit that put additional_db in that list; would appreciate an extra set of eyeballs.

naved001:
@zenhack You are right. Changing the order in the list of fixtures fixes the errors for `tests/unit/model.py`, and `tests/unit/api/auth.py` 

```
pytestmark = pytest.mark.usefixtures('configure','with_request_context','fresh_database', 'additional_db', 'server_init')
```
Still looking at `tests/unit/ext/auth/database.py` where 1 test is failing.

shwsun:
So I guess I don't need to dive into this now. Re: fixtures, IIRC, the order of them does matter, `additional_db` will add some more content to the database based on `fresh_db`. Probably it means I need to add some more doc/comment to our test infra

naved001:
* @shwsun I [reordered the fixtures](https://github.com/CCI-MOC/hil/compare/master...naved001:test-fixes?expand=1), so tests are now passing for `sqlite` but tests that rely on `postgres` are failing. Here are the logs: https://travis-ci.org/naved001/hil/builds/207052073 If you could provide some insight that would be helpful. 

* The tests pass if you run in an old environment, so most likely some package has been updated which affects the db operations and/or how the fixtures are loaded (just spitballing here). SQLAlchemy was recently updated, but even then using an old version of SQLAlchemy didn't fix these issues.

naved001:
Got it, the tests pass if we use `Flask-SQLAlchemy-2.1`. It was recently updated to 2.2. So restricting that would fix it. But I think we should still reorder the fixtures the right way instead of using an old version. For now I'll raise a PR to restrict the version of Flask-SQLAlchemy so that other PRs aren't blocked.

zenhack:
Closing, since #730 is merged.

PR #730, opened 2017-03-02
Restricts FlaskSQLAlchemy to 2.1

naved001:
Issue #729 
Flask-SQLAlchemy was recently updated to 2.2 which breaks our tests (and rightly so). 

This PR restricts it to an older version unless we figure out how to reorder the fixtures the right way (I am on it, but might take me a while).  
@zenhack @shwsun @henn

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -80,6 +80,9 @@ setup(name='haas',
       #
       # [1]: http://semver.org
       # [2]: https://packaging.python.org/en/latest/distributing/#choosing-a-versioning-scheme  # noqa
+      #
+      # For Flask-SQLAlchemy, we are using non-standard semver bounds as
+      # release 2.2 is backwards-incompatible.
       install_requires=['Flask-SQLAlchemy>=2.1,<2.2',
                         'Flask-Migrate>=1.8,<2.0',
                         'Flask-Script>=2.0.5,<3.0',
--- a/setup.py
+++ b/setup.py
@@ -80,7 +80,7 @@ setup(name='haas',
       #
       # [1]: http://semver.org
       # [2]: https://packaging.python.org/en/latest/distributing/#choosing-a-versioning-scheme  # noqa
-      install_requires=['Flask-SQLAlchemy>=2.1,<3.0',
+      install_requires=['Flask-SQLAlchemy>=2.1,<2.2',
                         'Flask-Migrate>=1.8,<2.0',
                         'Flask-Script>=2.0.5,<3.0',
                         'Werkzeug>=0.9.4,<0.10',


zenhack:
I'd prefer not to pin versions without understanding why. From what we've pieced together the problem so far has been *our bug*. For the remainder, there are several possibilities:

1. We're relying on details of the SQLAlchemy implementation that aren't part of the public API. This is a recipe for breakage and we need to fix that.
2. The Flask-SQLAlchemy developers introduced a bug upstream.
3. Flask-SQLAlchemy follows a different versioning scheme than is standard for python packages, in which backwards-incompatible changes are allowed on non-major versions. In this case, this patch is likely the correct solution, since the `<3.0` was assuming the standard conventions.
4. There was a transient issue with travis. The timeout seems like an unusual failure mode for one of our bugs, and I've had travis just time-out on me randomly before.

(4) is the easiest of these to investigate; can you just ask it to re-run the job and see if we get the same/different results?

knikolla:
@zenhack It fails consistently on my machine with a fresh virtualenv where packages are updated. It passes consistently on an older one with older packages. So I believe (4) is not the issue.

naved001:
> From what we've pieced together the problem so far has been our bug.

I agree
> There was a transient issue with travis. The timeout seems like an unusual failure mode for one of our bugs, and I've had travis just time-out on me randomly before.

I don't think that's the case here because the tests fail on my local environment too, so not a travis CI issue. (I re-ran the job anyway and it failed with the same results).

> The Flask-SQLAlchemy developers introduced a bug upstream.

I am not familiar with how Flask-SQLAlchemy works, but I think [this commit](https://github.com/mitsuhiko/flask-sqlalchemy/pull/461) from the FlaskSQLAlchemy changelog could be related to what we are experiencing.

zenhack:
Are you getting a timeout on your local machine as well? If not, what's the error? if so, have you tried putting a breakpoint in the test somewhere and poking around?

naved001:
On running the tests locally, it threw an error complaining additional_db being called outside of request context. The same error what TravisCI is showing. I don't know where you are getting a time out :/
> It's complaining that additional_db is being called outside of a request context (and thus without access to db.session) 

And as I said [here](https://github.com/CCI-MOC/hil/issues/729#issuecomment-283479176)  re-ordering the fixtures fixes this issue. But I haven't been able to fix all the tests yet, so I made this PR as a temporary solution.

zenhack:
I was talking about this travis run, which you linked to earlier:

https://travis-ci.org/naved001/hil/jobs/207052080

I ran it locally, and yeah the error messages make it clear it that it's because of the change from request context to app context. Probably what needs to happen is that we need to swap with_request_context out with something that constructs an *application* context.

So this looks like (3) is what's actually going on here -- the upstream devs made a backwards-incompatible release 2.1 -> 2.2.

Would you add a comment to setup.py explaining the reason we're not using the standard semver bounds? Other than that I'm ready to approve.

naved001:
@knikolla could you review this please?

shwsun:
Yes it seems good, i can give a +1. A question is, if we are using a feature no longer supported in Flask-SQLAchemy, do we want to open an issue and fix it later? @knikolla @zenhack

naved001:
@shwsun Thanks for looking at this!

zenhack:
I'll open an issue about the application vs request context thing. This looks good to me; merging.

Issue #731, opened 2017-03-03
When given a relative path with sqlite, haas-admin db create creates splits data across two databases.


zenhack:
The alembic_version table ends up in a db the current working directory, while everything else ends up in $root_of_source_repo/haas

I noticed this when working on #232.

Probably some weird descrepency between how we're configuring alembic vs. Flask-SQLAlchemy.

This ought to be fixed before #232, since although this wouldn't cause problems for production setups, it means developers would have to set absolute paths.

PR #732, opened 2017-03-07
Fix 232 and 731

zenhack:


tests/unit/cli.py
```diff
@@ -79,3 +79,13 @@ def test_serve():
...
+
+
+def test_serve_no_db():
+    with pytest.raises(CalledProcessError):
+        check_call(['haas', 'serve', '5000'])
+
+
+def test_serve_networks_no_db():
+    with pytest.raises(CalledProcessError):
+        check_call(['haas', 'serve_networks'])
```
henn: Same comment as above

haas/cli.py
```diff
@@ -295,7 +295,9 @@ def serve(port):
     # We need to import api here so that the functions within it get registered
     # (via `rest_call`), though we don't use it directly:
     from haas import model, api, rest
-    server.init(stop_consoles=True)
+    server.init()
+    server.check_db_schema()
```
henn: Is there a strong reason to run check_db_schema() here instead of within server.init()?

zenhack: haas-admin db create calls server.init(), but it can't call check_db_schema, since that would cause a circular dependency; it's job is to initialize the db, but that would cause it to bomb out every time.

zenhack:
@henn, responded to your comments.

zenhack:
(also, I apparently forgot to push my fix for #731, so that's new as well).

zenhack:
This has two signoffs, megring.

PR #736, opened 2017-03-07
Fix the version of oslo.policy to prevent CI breakage.

### LINKED ISSUE #734: Keystone is breaking our CI again
https://travis-ci.org/CCI-MOC/hil/jobs/208425804#L1678

Getting that on both of my open prs (which don't share code other than what's on master). This probably won't be hard, just a bit of fiddling to get the versions to play nice together.

*sigh*

henn:
Fixes #734

Collaborative effort with @naved001

### DIFF
--- a/ci/keystone/keystone.sh
+++ b/ci/keystone/keystone.sh
@@ -46,6 +46,10 @@ case "$1" in
     # Make sure we have the latest:
     pip install --upgrade pip
     
+    # The latest oslo.policy requires a weird version of requests that
+    # conflicts with other things as of 2017/03/07. See #734
+    pip install 'oslo.policy<1.19.0'
+
     pip install -r requirements.txt
     pip install .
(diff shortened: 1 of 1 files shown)


henn:
FYI - the reason we are doing this in the keystone install script (which we control) is because the conflict is caused by oslo.policy, a dependency of keystone itself. The issue is that the latest version of oslo.policy (1.9.0) forbids use of the currently latest requests library (2.13.0), which must be a dependency installed before the script gets to oslo.policy.

As it's not our package, we can't change keystone's `requirements.txt` directly to use an older version of oslo.policy or requests. We instead manually install an older version of oslo.policy which doesn't forbid the latest requests library.

Hopefully oslo.policy or requests will release a newer version sometime soon and we can remove this.

zenhack:
Grumble, grumble, +1 merging.

PR #739, opened 2017-03-08
Use upper contraints file for keystone, set version to newton

knikolla:
OpenStack uses an upper contraint file to prevent package version
conflicts among the different projects. This file is available
at [0]. You can request an upper contraints file for a specific
version of OpenStack by appending "?h=stable/version", such as
"?h=stable/ocata", for the latest stable version.

Since this requires using stable branches and doesn't support
tags for the version you require, I switched from using
a specific tag, to using the stable branch netwon.

Not using ocata because ocata defaults to using fernet tokens,
which need to be setup.

0. https://git.openstack.org/cgit/openstack/requirements/plain/upper-constraints.txt

### DIFF
--- a/ci/keystone/install.sh
+++ b/ci/keystone/install.sh
@@ -5,4 +5,4 @@ pip install python-keystoneclient
 # The exact commit we use here is somewhat arbitrary, but we want
 # something that (a) won't change out from under our feet, and (b)
 # works with our existing tests.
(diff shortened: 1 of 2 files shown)


knikolla:
@zenhack @henn @naved001

shwsun:
I am happy with this.

naved001:
looks good! +1

PR #741, opened 2017-03-16
Fix #629

### LINKED ISSUE #629: Erroneous use of `%` operator for string-formatting in calls to logging functions.
All over the codebase you can find statments like:

``` python
logger.error('Foo: %r' % bar)
```

The `%` operator being python's format-string splicing operator. The problem with this is that the logging functions do the formation string splicing themselves, i.e. what you want in this case is:

``` python
logger.error('Foo: %r', bar)
```

This opens up the possibility of format-string injection vulnerabilities. Frankly, this is too easy to do, especially since in other contexts `%` is the correct thing. We ought to (a) make sure all instances of this mistake are fixed, and (b) come up with a way to catch this mistake automatically going forward; perhaps some kind of linter.

zenhack:
This adds a test that lints the source tree for the issue, and fixes the violations that were present.


tests/lint.py
```diff
@@ -0,0 +1,99 @@
+"""Tests performing linter-like checks"""
```
henn: Maybe this should be called `custom_lint.py` or something like that, to differentiate against running some other linter?

ci/run_unit_tests.sh
```diff
@@ -8,4 +8,7 @@
 #     extra_flags='-n auto'
 # fi
 
-py.test $extra_flags tests/unit tests/stress.py
+py.test $extra_flags \
+	tests/unit \
+	tests/lint.py \
```
henn: Since this runs quickly, could you move it to the beginning?

zenhack:
Quoting Jason Hennessey (2017-03-16 18:22:36)

>    Minor feedback. Overall, I like it, though think this functionality
>    might better belong to something like one or more of the [1]python
>    linters.

I thought I'd looked through them and not found the functionality, but
pylint does seem to support this. I agree that I'd rather to it that way
than maintain our own; I'll see about swapping in that implementation.

>    Rather than including everything by default and using a blacklist,
>    could we switch to a whitelist of directories to scan (like haas,
>    scripst, tests, maybe examples)?

I'm worried about missing things.

>
>    Sometimes I name my virtualenv directory something without a dot at the
>    front to remind myself that I've created one in a particular directory,
>    and I don't want to get bombarded

I'm okay adding a few more things to accommodate other developers'
workflows, though of course I don't want the size of the list to spiral
out of control.  I'm sure we can come up with a small-ish number of ways
to name things that works for everyone though.  How about adding
`startswith('_')`?

henn:
>     Rather than including everything by default and using a blacklist,
>     could we switch to a whitelist of directories to scan (like haas,
>     scripst, tests, maybe examples)?
> I'm worried about missing things.

We currently do this in ci/pep8_tests.sh:
```python
  pep8 *.py tests/ haas/
```

zenhack:
Quoting Jason Hennessey (2017-03-16 22:55:49)

>    We currently do this in ci/pep8_tests.sh:
>
>   pep8 *.py tests/ haas/

One of the issues was in examples/, and we ought to be passing it
scripts/ as well. I think this demonstrates my point.

zenhack:
So I looked at each of the linters in linked from that page. Pylint is the only one that checks for this at all, and when I ran it on master it missed 2 of the 6 instances that mine caught. I agree that this is something there ought to be a generic tool for that we can just use, but the available options don't seem to be up to snuff. I'd rather maintain the solution that I wrote than use something that doesn't work properly.

zenhack:
@henn, ping.

naved001:
@knikolla could you review this please? Thanks!

tests/lint.py
```diff
@@ -0,0 +1,99 @@
...
+    _visit_file(source_root)
+
+
+class LogCallVisitor(ast.NodeVisitor):
+    """Ast node visitor used by test_logger_format_strings."""
+
+    def __init__(self, filename):
+        self.filename = filename
+
+    def visit_Call(self, node):
```
knikolla: ++

zenhack:
@henn, ping. I addressed your other two comments, but per my comment above I don't think the existing linters cut it. It's been almost a month since I mentioned that, and I haven't heard back.

zenhack:
@henn, ping. I addressed your other two comments, but per my comment above I don't think the existing linters cut it. It's been almost a month since I mentioned that, and I haven't heard back.

henn:
@zenhack - do you know why CI is complaining?

zenhack:
@henn, odd, the line it's referencing doesn't actually exist. There were merge conflicts now that the port_revert stuff is merged, so I fixed those, and there was an error in the brocade driver. It passes locally (but so did the old version); we'll see what travis does this time.

tests/lint.py
```diff
@@ -0,0 +1,99 @@
...
+    > come up with a way to catch this mistake automatically going
+    > forward; perhaps some kind of linter.
+
+    This is that linter; it scans the source tree looking for places
+    where the logging functions are called with any first argument
+    that isn't a string literal.
+    """
+
+    def _visit_file(filename):
+        if basename(filename).startswith('.'):
```
zenhack: Arguably the "right" way to do this is by looking at `.gitignore`, but I don't know if I think it's worth the complexity.

henn:
>>  We currently do this in ci/pep8_tests.sh:
>>
>>   pep8 *.py tests/ haas/

> One of the issues was in examples/, and we ought to be passing it
> scripts/ as well. I think this demonstrates my point.

Would having a *better* whitelist seem reasonable to you?
For instance, keeping the whitelist in a single file (with `examples/` added), then referencing it from here and from the pep8 test?

I'd prefer to place as few project-specific naming restrictions as we can get away with.

zenhack:
@henn, what do you think of generating the whitelist like this:

```sh
cd "$(git rev-parse --show-toplevel)" # Enter the soruce tree root
# list all files not matched by a .gitignore and ending in .py or .wsgi:
git ls-files $(find -name .gitignore| sed 's/^/--exclude-per-directory /') | \
	grep -E '\.(py|wsgi)$'
```

zenhack:
@henn, done. Not surprisingly, we were missing a couple of files.

henn:
Cool. Would you mind addressing my comments in `tests/lint.py`.

After that, I think we're good to go.

zenhack:
@henn, I think that's everything -- sorry for missing those on the first pass.

PR #742, opened 2017-03-20
Fix Deployment tests for brocade

naved001:
Some methods (used by the deployment tests) were passing the class objects as a part of the URL which the switch doesn't understand. So I have set a filter to convert it to string before passing it to the switch. 

This fixes the deployment tests for the brocade switch. 

@knikolla @zenhack

### DIFF
--- a/haas/ext/switches/brocade.py
+++ b/haas/ext/switches/brocade.py
@@ -252,6 +252,9 @@ class Brocade(Switch):
         # % escapes the % character.
         # Double quotes are necessary in the url because switch ports contain
         # forward slashes (/), ex. 101/0/10 is encoded as "101/0/10".
+        if not isinstance(interface, basestring):
+            interface = interface.label
+
         return '%(hostname)s/rest/config/running/interface/' \
             '%(interface_type)s/%%22%(interface)s%%22/switchport/%(suffix)s' \
             % {
--- a/haas/ext/switches/brocade.py
+++ b/haas/ext/switches/brocade.py
@@ -104,8 +104,9 @@ class Brocade(Switch):
         """
         response = {}
         for port in ports:
-            response[port] = filter(None, [self._get_native_vlan(port)]) \
-                             + self._get_vlans(port)
+            response[port] = filter(None,
+                                    [self._get_native_vlan(port.label)]) \
+                                    + self._get_vlans(port.label)
         return response
 
     def _get_mode(self, interface):
@@ -252,9 +253,6 @@ class Brocade(Switch):
         # % escapes the % character.
         # Double quotes are necessary in the url because switch ports contain
         # forward slashes (/), ex. 101/0/10 is encoded as "101/0/10".
-        if not isinstance(interface, basestring):
-            interface = interface.label
-
         return '%(hostname)s/rest/config/running/interface/' \
             '%(interface_type)s/%%22%(interface)s%%22/switchport/%(suffix)s' \
             % {
--- a/tests/unit/ext/switches/brocade.py
+++ b/tests/unit/ext/switches/brocade.py
@@ -165,25 +165,30 @@ class TestBrocade(object):
 
     def test_get_port_networks(self, switch):
         with requests_mock.mock() as mock:
+
+            PORT1 = model.Port(label=INTERFACE1, switch=switch)
+            PORT2 = model.Port(label=INTERFACE2, switch=switch)
+            PORT3 = model.Port(label=INTERFACE3, switch=switch)
+
             mock.get(switch._construct_url(INTERFACE1, suffix='trunk'),
                      text=TRUNK_NATIVE_VLAN_RESPONSE_WITH_VLANS)
             mock.get(switch._construct_url(INTERFACE2, suffix='trunk'),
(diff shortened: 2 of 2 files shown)


haas/ext/switches/brocade.py
```diff
@@ -252,6 +252,9 @@ def _construct_url(self, interface, suffix=None):
         # % escapes the % character.
         # Double quotes are necessary in the url because switch ports contain
         # forward slashes (/), ex. 101/0/10 is encoded as "101/0/10".
+        if not isinstance(interface, basestring):
```
knikolla: I would rather do, ``if isinstance(interface, Port (or whatever that is))``

naved001: Good Idea!

zenhack:
Where is `_construct_url` being called with a Port instead of the port's label? The problem is we have a type error somewhere -- I'd rather fix that than add special-purpose code to handle whatever vaguely related value the caller passes in. We should also update the docs to be clear about what the type of interface is supposed to be.

naved001:
@zenhack 

`_get_vlans(self, interface):` https://github.com/CCI-MOC/hil/blob/master/haas/ext/switches/brocade.py#L145 and  ` _get_native_vlan(self, interface):`  are the functions called by the deployment tests. The deployment tests pass the `interface` (Port) to these functions.  Then these functions pass `interface` to `_construct_url`. The interface gets passed on as an object (Port), instead of interface.label. The switch expects the name of the interface not the class object. 

If you look at the `apply_networking(self, action)` [function here](https://github.com/CCI-MOC/hil/blob/master/haas/ext/switches/brocade.py#L63) You'll see that in line 69 `interface = action.nic.port.label` i.e set to port name and then that interface name is passed to 4 functions that are inside `apply_networking` which is why those work. So the rest of the functions that are directly passed `interface` wouldn't work. 

I hope I explained it clearly. I am open to other solutions to this issue.

zenhack:
The deployment test doesn't call anything starting with an underscore directly; those are private to the driver. It *does* call get_port_networks, which per the docstring is supposed to accept port objects. The bug is in that function; it should be passing in the label, not the object.

naved001:
oh yes. get_port_networks calls those underscore functions that are private to the driver. And you are right, the bug is in those functions; it should be passing in the label.  

deployment functions -> get_port_networks -> underscore functions -> _construct_url

So somewhere, in the driver, we need to make sure we pass the port name. I decided to do it in the _construct_url function. Or maybe we can have a new function that does this, like, `_get_port_name` which is called by all the functions that use interface?

zenhack:
Quoting Naved Ansari (2017-03-20 21:15:09)

>    oh yes. get_port_networks calls those underscore functions that are
>    private to the driver. And you are right, the bug is in those
>    functions; it should be passing in the label.
>
>    deployment functions -> get_port_networks -> underscore functions ->
>    _construct_url
>
>    So somewhere, in the driver, we need to make sure we pass the port
>    name. I decided to do it in the _construct_url function. Or maybe we
>    can have a new function that does this, like, _get_port_name which is
>    called by all the functions that use interface?


Just add `.label` to anywhere you'd otherwise be passing a port object
to one of those functions.

naved001:
1. The unit tests for the brocade driver call `get_port_networks` with the port label. So adding, `.label` breaks some of the unit tests, in that case we would need to re-write the unit tests. 
Or
2. In `get_port_networks` we can do a check like `if isintance(port, Port): port = port.label` that way we don't break the unit tests. 

What should we do? 1 or 2?

zenhack:
We should do (1). The spec in the docstrings for the `Switch` class in `haas.model` makes it very clear that the function expects port objects; any callers that are passing labels are incorrect. In a language that had a type checker, they wouldn't even compile.

naved001:
@zenhack 
> We should do (1). 

Done. 

@knikolla I have made some changes, could you review this again?

zenhack:
I think I may have miscommunicated. Per the docs, the tests themselves were doing the correct thing in passing in port objects:

https://github.com/CCI-MOC/hil/blob/master/haas/model.py#L248

The error is in the brocade drivers' implementation of get_port_networks, which calls passes the port object to private functions that expect the interface *name*. The correct solution is to change the arguments to `_get_vlans` and `_get_native_vlan` to `port.label`. We should also update the docstrings for those functions to make it more clear that the *name* is what is expected.

naved001:
@zenhack 

> The error is in the brocade drivers' implementation of get_port_networks, which calls passes the port object to private functions that expect the interface name. The correct solution is to change the arguments to _get_vlans and _get_native_vlan to port.label.

That is what I have done. Isn't it?

> Per the docs, the tests themselves were doing the correct thing in passing in port objects:

The tests aren't passing Port Objects, but a string with port name (INTERFACE1, INTERFACE2 are just strings defined on top). I modified the unit tests to pass a list of Port Objects instead of string because that's what `get_port_networks` expects. 

Or am I misunderstanding something here?

zenhack:
Ack, you're right, you totally did. Apparently I've gone blind in the past few hours. :/

This looks good to me then, +1.

PR #745, opened 2017-03-21
Modified the dell and cisco drivers to fix the deployment tests.

### LINKED ISSUE #740: Deployment tests for networks don't pass
I was trying to run deployment tests for #702 #683 but those were failing. So I decided to run the tests with what we currently have in the master and the tests failed for all the switches. The deployment tests only passed with mock drivers. 

I have put the logs in the comments here; the tests are getting stuck at `create_networks()`

I can share my `site-layout.json` and `testsuite.cfg` files if you feel like testing. 

When was the last time we ran the network deployment tests with real hardware?

naved001:
Collaborative effort with @zenhack 

Fixes #740

### DIFF
--- a/haas/ext/switches/dell.py
+++ b/haas/ext/switches/dell.py
@@ -162,20 +162,17 @@ class _Session(_console.Session):
         """
 
         alternatives = [
-            re.escape(r'More: <space>, '
-                      'Quit: q or CTRL+Z, One line: <return> '),
+            r'More: .*',  # Prompt to press a key to continue
             r'Classification rules:\r\n',  # End
             r'[^ \t\r\n][^:]*:[^\n]*\n',   # Key:Value\r\n,
             r' [^\n]*\n',                  # continuation line (from k:v)
         ]
         self._sendline('show int sw %s' % interface)
-        # Find the first Key:Value pair (this is needed to skip past some
-        # possible matches for other patterns prior to this:
-        self.console.expect(alternatives[2])
 
+        # Name is the first field:
+        self.console.expect('Name: .*')
         k, v = self.console.after.split(':', 1)
         result = {k: v}
-
         while True:
             index = self.console.expect(alternatives)
             if index == 0:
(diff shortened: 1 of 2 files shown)


zenhack:
I feel like I shouldn't be a reviewer on this, since as @naved001 says we both hacked on this. I'm going to assign some other folks.

knikolla:
@naved001 can you provide more information on what was wrong and how you fixed it?

henn:
@zenhack  - since you guys wrote this together, I think it's totally appropriate for you to act as a reviewer, since presumably you were reviewing this as you wrote it.

naved001:
Both, the dell and cisco, drivers were timing out when running the tests. 

In the dell driver, pexepct times out when we try to gather information for an interface because it waits for the "user" to press a key. Because the regex was too strict, it wouldn't find a match and the call would timeout. So what we have done is made the regex pattern to be matched more relxed (Line 165).
The other issue with it was how it was building a dictionary containing the interface information. It was a hodgpe podge because of the timing issues, so changing the pattern here fixes this (Line 173). 

In the nexus driver, sometimes the switch can have "custom" interface names and when it doesn't match what we expect it crashes, so we changed it so that continues to run when it encounters such a situation. (Line 159 and 160)
Also, the name of the switch that the port returns is "Ethernet" with a capital 'E', but we were checking for "ethernet" with a small 'e' so we couldn't find a match. (Line 164).

naved001:
@shwsun could you review this? it's a small PR. Thanks!

shwsun:
Thanks for the explanation to give us the context, could you also update the doc a bit @naved001 ? I will run the test on devhaas and +1 soon.

naved001:
@shwsun I don't know what to put about this in the documents because we don't really describe the driver code anywhere. The commit message has a detailed explanation of what this change is about, which I think is sufficient.

SahilTikale:
Checked the code. Ran tests, passing all deployment tests. 
+1

SahilTikale:
merging.

PR #746, opened 2017-03-22
Cli rework in argparse (user only)

meng-sun:
This includes the transition code that @henn mentioned last meeting. Users can now call cli commands and cli_new commands (for user only) on the command line. Once this piece of code is checked, we can add the other pieces to it. Better documentation will be coming over spring break, but right now please use the example guide we sent over email during IAP.

Please squash commits when merging, thanks!


henn:
Some initial comments/questions from myself and @naved001 :

1. I really like the object oriented design that you used and how well you organized the code!
2. Is it possible to place argument parsing closer to the function itself? For example, `CommandListenener` contains the function *user_register_func()*, which contains all of the parsing info for *user_create()*. If possible, it would be nice to keep all of the info about a method, including arg parsing, in the same place, next to the function itself.
4. Similar to above, it might be worth considering moving the creation of subparsers out of the big `CommandListener` class in cli_new.py and next to the functions. That way, later down the line if we decide to break the CLI into submodules, new commands can in a sense register themselves. I would suggest that the only options kept in the big `CommandListener` be global options (like `--user` or `--endpoint`). This is more stylistic, though.
3. Can we reduce the functions to just the user-related ones for this pull request? That way, we can keep the size of this PR to a minimum to focus on getting the new CLI infrastructure the way we want it and not need to spend a lot of time coding/reviewing the other components. We can save that code off and integrate in a future PR.
5. In keeping with the [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), we should have the `argcomplete` code be the same as the regular argparse. Otherwise, it will be really easy for them to get out of sync.

henn:
Also, @naved001 and I addressed at least one reason that the tests were failing in cebd7b2da3ba8fc77bdf404ae0c660bf73226c18

henn:
Looking at the newest failures, I think implementing suggestion 4 above will fix it, as the tests are complaining about the return values from `haas serve`. By not implementing `haas serve` in this PR, we can work on that later!

Yay for procrastination!

zenhack:
We've got merge conflicts and this doesn't look terribly near completetion bitrot aside. Wondering how we want to go about managing this, given the impending freeze/get everything merged sprint.

henn:
I'm going to close this for now, until we can find an owner.

Issue #688, opened 2016-11-30
HIL needs to write flash config


henn:
Currently, at least the dell driver does not save the flash config, such that if a power outage happens the changes persist.

We need to:
a) Check all the drivers for this class of bug, and
b) fix it.

One possible way to address it would be to stick a "write ..." command in the switch driver's disconnect(). By doing it there rather than after each port mod, we might cut down on flash overwrites and potentially save some wear while speeding things up.

henn:
Looking at #673, since there could be a time/flash wearing vs. consistency tradeoff maybe we could make it a tunable as to whether the `write ...` happens in disconnect() or after every change?

zenhack:
One other option: we could recommend having the switch's stored config have all ports managed by HIL be *disconnected from everything*, so in the case of a power outage we fail-safe. This sidesteps the tricky problem of coordinating database commits and flash writes, as well having to worry about wearing out the flash.

henn:
I like that.
Going into our discussion (hopefully later this week), I think we need to keep in mind that we a) don't want anyone to have access to networks they shouldn't, and b) should have a sane way for HIL providers to restore or reinstall their switch.

We've discussed 4 (mostly non-exclusive) ideas so far:

1. Write to flash (mentioned above)

2. (@zenhack's comment above) Have a default config that has all ports shut off. We may still need something like #687, since I noticed that once the switch was reset, using HIL to disconnect then reconnect the port gave me access to all VLANs

3. Have HIL maintain a copy of the switch config every time that is refreshed after every change.

4. Recommend that HIL providers have an external mechanism for backing up switch config.

henn:
We just had some more discussion with @naved001 @knikolla @kylehogan @meng-sun and @gfaline. It sounds like a plausible solution would be to add a reset_nic(), in addition to having reset_port().

reset_nic() would be able to handle situations like needing to replace a dead switch with one or two others that might have different port capacities. After the nic was disconnected from the previous switch and connected to the new one, one could run reset_nic(). It could also handle moving a node between racks.

reset_port() would be helpful in cases where a switch was replaced in-kind, and one could simply run reset_port() on all of the ports to get them back to functional. It would also handle cases like we had with losing power, since we wouldn't know which nodes needed to be reset -- we just want to do them all.

naved001:
closing this because #749 and #753 were merged.

PR #749, opened 2017-03-28
Dell switch: Save running configuration to flash memory

naved001:
This partially addresses #688 

I have made this feature optional. User/Admin can enable this from the config file.  

The running config file is saved to the startup config file before disconnecting from the switch. 
There's also a deployment test. I'll make similar modifications to the nexus switch based on the feedback here. 

@henn @zenhack

### DIFF
--- a/examples/haas.cfg
+++ b/examples/haas.cfg
@@ -114,3 +114,8 @@ haas.ext.auth.database =
 # are required.
 vlans = 100-200
 
+[haas.ext.switches.dell]
+# This section can be used when dell switch is in use.
+# It enables saving the switch configuration to memory.
+# save = True
+
--- a/examples/haas.cfg
+++ b/examples/haas.cfg
@@ -116,5 +116,4 @@ vlans = 100-200
 
 [haas.ext.switches.dell]
 # Make modifications to the dell switches' configuration persistent
-# save = True
-
+save = True
--- a/examples/haas.cfg
+++ b/examples/haas.cfg
@@ -115,7 +115,6 @@ haas.ext.auth.database =
 vlans = 100-200
 
 [haas.ext.switches.dell]
-# This section can be used when dell switch is in use.
-# It enables saving the switch configuration to memory.
+# Make modifications to the dell switches' configuration persistent
 # save = True
 
--- a/examples/haas.cfg
+++ b/examples/haas.cfg
@@ -115,5 +115,7 @@ haas.ext.auth.database =
 vlans = 100-200
 
 [haas.ext.switches.dell]
-# Make modifications to the dell switches' configuration persistent
+# By default, the modifications made to the dell switches' configuration are
+# persistent. Set `save` to False to stop the switch from writing to
+# flash memory.
 save = True
--- a/haas/ext/switches/dell.py
+++ b/haas/ext/switches/dell.py
@@ -114,8 +114,11 @@ class _Session(_console.Session):
         self._sendline('sw trunk native vlan none')
 
     def disconnect(self):
+        save = True
         if (cfg.has_option('haas.ext.switches.dell', 'save') and
-           cfg.getboolean('haas.ext.switches.dell', 'save')):
+           not cfg.getboolean('haas.ext.switches.dell', 'save')):
+            save = False
+        if save:
             self._save_running_config()
         self._sendline('exit')
         self.console.expect(pexpect.EOF)
--- a/haas/ext/switches/dell.py
+++ b/haas/ext/switches/dell.py
@@ -114,9 +114,9 @@ class _Session(_console.Session):
         self._sendline('sw trunk native vlan none')
 
     def disconnect(self):
-        if cfg.has_option('dell', 'save'):
-            if cfg.getboolean('dell', 'save'):
-                self._save_running_config()
+        if (cfg.has_option('haas.ext.switches.dell', 'save') and
+           cfg.getboolean('haas.ext.switches.dell', 'save')):
+            self._save_running_config()
         self._sendline('exit')
         self.console.expect(pexpect.EOF)
         logger.debug('Logged out of switch %r', self.switch)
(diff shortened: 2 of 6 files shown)


haas/test_common.py
```diff
@@ -273,6 +273,15 @@ def collect_nodes(self):
                 len(nodes))
         return nodes
 
+    def get_config(self, config_type):
```
zenhack: This has problems:

1. It leaks the network connection.
2. It calls a private method specific to the dell driver, without any guarantee that the switch uses that driver.

tests/deployment/switch_config.py
```diff
@@ -0,0 +1,97 @@
+# Copyright 2013-2014 Massachusetts Open Cloud Contributors
```
zenhack: Should probably bump the date.

tests/deployment/switch_config.py
```diff
@@ -0,0 +1,97 @@
...
+site_layout = pytest.fixture(site_layout)
+
+pytestmark = pytest.mark.usefixtures('configure',
+                                     'server_init',
+                                     'fresh_database',
+                                     'with_request_context',
+                                     'site_layout')
+
+
+class TestNativeNetwork(NetworkTest):
```
zenhack: Would you change this to a more descriptive name, add add a docstring explaining what this test is for?

Also, this file is dell-specific; if that's the case I'd like it to be called out more prominently, and the file probably out to be separated from the rest of the deployment tests so you can still do py.test tests/deployment if you don't have a dell switch in your setup. The other option is to use `pytest.skip()`

haas/ext/switches/dell.py
```diff
@@ -113,6 +114,9 @@ def disable_native(self, vlan_id):
         self._sendline('sw trunk native vlan none')
 
     def disconnect(self):
+        if cfg.has_option('dell', 'save'):
+            if cfg.getboolean('dell', 'save'):
```
naved001: I was just worried if we don't have that section then cfg.getboolean returns an error. But now that I tested it it doesn't break the flow of the program. So I'll change it to use `and`.

haas/test_common.py
naved001: > It leaks the network connection.

What does that mean?

Also, I am unsure how to handle this. Any suggestions?

tests/deployment/switch_config.py
naved001: oh yes. It's a copypasta from the other tests. I'll clean it up.

naved001:
> Have we actually thought through the consequences of doing this every time we make a change? I think the possibility of wearing out the flash was raised. Adding a feature like this to the driver only makes sense if we consider it a reasonable solution, and it's not 100% clear to me that it is.

I discussed this with @pjd-nu and he said that the possibility of the flash memory going bad anytime soon is very unlikely.

zenhack:
Quoting Naved Ansari (2017-03-29 12:55:45)

>    I was just worried if we don't have that section then cfg.getboolean
>    returns an error. But now that I tested it it doesn't break the flow of
>    the program. So I'll change it to use and.

Yeah, and is short-circuting just like in C, Java, etc.

zenhack:
Quoting Naved Ansari (2017-03-29 13:01:48)

> I discussed this with [1]@pjd-nu and he said that the possibility of
> the flash memory going bad anytime soon is very unlikely.

Ok, works for me.

zenhack:
Quoting Naved Ansari (2017-03-29 12:57:23)

>      It leaks the network connection.
>
>    What does that mean?

When you create a session, it connects to the switch. You need to call
disconnect when you're done.

>    Also, I am unsure how to handle this. Any suggestions?

Re: (1) the obvious solution is to close the session when you're done
with it.

Re: (2), since you're only using it in the one test, and I'm already
asking you to mark off the test as dell-specific, you could just do
this in that test, rather than test_common, in which case calling
private methods isn't really a problem, since it's tightly coupled
anyway.

naved001:
@zenhack I have made the changes you requested.

examples/haas.cfg
```diff
@@ -114,3 +114,8 @@ haas.ext.auth.database =
 # are required.
 vlans = 100-200
 
+[haas.ext.switches.dell]
+# This section can be used when dell switch is in use.
```
zenhack: This line goes without saying, per the general rules for extensions.

examples/haas.cfg
```diff
@@ -114,3 +114,8 @@ haas.ext.auth.database =
...
+# It enables saving the switch configuration to memory.
```
zenhack: This is a little unclear. When someone says "memory" I immediately think "RAM" which is not what we're talking about here. How about `Make modifications to the dell switches' configuration persistent:`?

tests/deployment/switch_config.py
```diff
@@ -54,7 +61,30 @@ def server_init():
                                      'site_layout')
 
 
-class TestNativeNetwork(NetworkTest):
+@pytest.fixture
+def not_dell():
+    """open the site-layout file to see if we don't have a dell switch"""
+
+    layout_data = open('site-layout.json')
```
zenhack: In general, prefer using a `with` statement when working short-term with files:

```python
with open('filename') as f:
    data = json.load(f)
```

This way it's immediately apparent that you're not leaking the file object; you close it correctly below, but the above is more readable and less error prone.

tests/deployment/switch_config.py
```diff
@@ -54,7 +61,30 @@ def server_init():
...
+    layout = json.load(layout_data)
+    layout_data.close()
+    for switch in layout['switches']:
+        return switch['type'] != DELL
```
zenhack: This is only going to check the first switch. If this was your intent, you can dispense with the loop and just check the first element directly. Otherwise, you've got a bug.

zenhack:
Made a few more comments. Getting there.

tests/deployment/switch_config.py
naved001: I wasn't sure if we are ever going to run deployment tests with multiple switches described in the site-layout file. So what I wanted to do here is, assuming we have more than 1 switch, return false if any of the switch isn't a Dell so that we skip that test.

naved001:
@zenhack your comments are really helpful! 

I have made some changes. let me know if there's something else that needs to be addressed.

zenhack:
LGTM

examples/haas.cfg
```diff
@@ -114,3 +114,7 @@ haas.ext.auth.database =
 # are required.
 vlans = 100-200
 
+[haas.ext.switches.dell]
+# Make modifications to the dell switches' configuration persistent
```
henn: Please start the sentence with something like `# Uncomment to make modifications...`

zenhack: @henn, what you're suggesting is inconsistent with what most of the comments in this file do. See e.g. the comments in the database section. I think it's clear as is.

zenhack:
Quoting Jason Hennessey (2017-04-05 14:18:26)

>    The question on my mind is whether this should default to on or off.
>
>    In the interest of doing the least surprising thing, I'm inclined to
>    think we should make it the default.
>    Thoughts?

I think it should behave the same way as the other drivers. I otherwise
agree with you.

naved001:
This is how it works in other drivers. 

* Brocade switch was setup in a way by the admins such that it always saves the config to flash memory. 
* The cisco switch doesn't save it. I am gonna tweak the driver so that it supports saving to flash memory. (@zenhack should I change the cisco driver right here or make it into another PR? )

I would also make it save by default.

zenhack:
@naved001, let's make it a separate pr.

In that case, I'm in favor of having saving to flash by default.

naved001:
@henn @zenhack I have changed it to save by default. It will save even if the extension section in the config file is not present. To stop it from saving, we would need to have `haas.ext.switches.dell` in the config file with the boolean `save` set to False.

zenhack:
Would you also update the comment in tge example config to make it clear that this is the default behavior?

naved001:
@zenhack done
@henn ping

zenhack:
Cool, +1 from me, just waiting on @henn

Issue #748, opened 2017-03-24
Brocade Driver: less strict error logging


naved001:
The deployment tests are failing again; even for the master (my bad; should have looked at #702 more carefully. the tests pass if you run it once, but fail when ran the second time). 

As far as I understand, the problem here is that every time a function is called again later, that according to the switch is a bad request which raises logger.error. 
For example:
When _enable_and_set_mode() is called the the second time (on the same port), it will try to re-enable switchport and the switch would say Bad Request. 

Similarly, this was when I ran the vlan_networks test. I think there was nothing to DELETE, so this was another Bad Request. 

```
haas/ext/switches/brocade.py:241: in _set_native_vlan
    self._disable_native_tag(interface)
haas/ext/switches/brocade.py:263: in _disable_native_tag
    self._make_request('DELETE', url)
haas/ext/switches/brocade.py:299: in _make_request
    logger.error('Bad Request to switch. Response: %s' % r.text)
```

@knikolla @zenhack would appreciate an extra set of eyes on this. 

I think we should change [logger.error](https://github.com/CCI-MOC/hil/blob/master/haas/ext/switches/brocade.py#L283) to logger.debug otherwise we would have to check state in every function before performing any operation.

zenhack:
I'm nervous about toning down the error checking too much here; I'm worried about a scenario where the switch returns some error code because an operation failed, and then we plow ahead e.g. having not actually detached a network properly, giving someone unintended access to someone else's network.

Options:

1. Suck it up and add the extra check everywhere. This has the advantage that failures are loud, instead of silent, but it is a bit more work than is ideal.
2. Do as you suggest and log this stuff at a lower severity.
3. Somehow encode "acceptable" responses. maybe add a parameter to `_make_requests` that specifies a list of acceptable statuses. This way, we don't have to write explicit checks everywhere, but we find out if something fails in a way we're *not* expecting.

naved001:
These error checks were added in #702.  

In a scenario where a switch returns an error, the networking server (`haas serve_networks`) continues to run without stopping and users wouldn't know unless they look at the logs. So raising a logging error doesn't really help as the user is unaware anyway. 

The deployment tests are failing because we are using the fixture `fail_on_log_warnings` . So the test would stop whenever loger.error is thrown, but the networking server wouldn't. 

Option (1) seems like the right way to do things, but then we may have to add more methods in the brocade driver to have extra checks which in turn might introduce some bugs. 

[...]

PR #750, opened 2017-03-30
Brocade driver: disable logging for _disable_native_tag and when renabling the switchport

naved001:
#748 

the switch returns 409 when renabling the switchport and 404 when we call
_disable_native_tag (called by _set_native_vlan). Due to this, the deployment
tests would fail. So this PR disables logging for these 2 cases.

@knikolla

### DIFF
--- a/haas/ext/switches/brocade.py
+++ b/haas/ext/switches/brocade.py
@@ -139,7 +139,7 @@ class Brocade(Switch):
         # Enable switching
         url = self._construct_url(interface)
         payload = '<switchport></switchport>'
-        self._make_request('POST', url, data=payload)
+        self._make_request('POST', url, data=payload, log=False)
 
         # Set the interface mode
         if mode in ['access', 'trunk']:
@@ -244,7 +244,7 @@ class Brocade(Switch):
 
         """
         url = self._construct_url(interface, suffix='trunk/tag/native-vlan')
-        self._make_request('DELETE', url)
+        self._make_request('DELETE', url, log=False)
 
     def _construct_url(self, interface, suffix=''):
         """ Construct the API url for a specific interface appending suffix.
@@ -277,8 +277,8 @@ class Brocade(Switch):
         """ Construct the xml tag by prepending the brocade tag prefix. """
         return '{urn:brocade.com:mgmt:brocade-interface}%s' % name
 
-    def _make_request(self, method, url, data=None):
+    def _make_request(self, method, url, data=None, log=True):
         r = requests.request(method, url, data=data, auth=self._auth)
-        if r.status_code >= 400:
+        if (r.status_code >= 400 and log):
             logger.error('Bad Request to switch. Response: %s' % r.text)
         return r
--- a/haas/ext/switches/brocade.py
+++ b/haas/ext/switches/brocade.py
@@ -139,7 +139,8 @@ class Brocade(Switch):
         # Enable switching
         url = self._construct_url(interface)
         payload = '<switchport></switchport>'
-        self._make_request('POST', url, data=payload, log=False)
+        self._make_request('POST', url, data=payload,
+                           acceptable_error_codes=[409])
 
         # Set the interface mode
         if mode in ['access', 'trunk']:
@@ -244,7 +245,7 @@ class Brocade(Switch):
 
         """
         url = self._construct_url(interface, suffix='trunk/tag/native-vlan')
-        self._make_request('DELETE', url, log=False)
+        self._make_request('DELETE', url, acceptable_error_codes=[404])
 
     def _construct_url(self, interface, suffix=''):
         """ Construct the API url for a specific interface appending suffix.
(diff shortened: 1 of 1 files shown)


haas/ext/switches/brocade.py
```diff
@@ -277,8 +277,8 @@ def _construct_tag(name):
         """ Construct the xml tag by prepending the brocade tag prefix. """
         return '{urn:brocade.com:mgmt:brocade-interface}%s' % name
 
-    def _make_request(self, method, url, data=None):
+    def _make_request(self, method, url, data=None, logging=True):
```
knikolla: please use ``log`` instead of ``logging`` as it shadows the build in ``logging`` module.

naved001: done

haas/ext/switches/brocade.py
```diff
@@ -139,7 +139,7 @@ def _enable_and_set_mode(self, interface, mode):
         # Enable switching
         url = self._construct_url(interface)
         payload = '<switchport></switchport>'
-        self._make_request('POST', url, data=payload)
+        self._make_request('POST', url, data=payload, log=False)
```
knikolla: Can you check for the expected error code in the case of the switchport already being enabled and log if it's a different one?

haas/ext/switches/brocade.py
```diff
@@ -244,7 +244,7 @@ def _disable_native_tag(self, interface):
 
         """
         url = self._construct_url(interface, suffix='trunk/tag/native-vlan')
-        self._make_request('DELETE', url)
+        self._make_request('DELETE', url, log=False)
```
knikolla: Can you check for the excepted error code in the case of the vlan being already disabled and log if it's a different one?

haas/ext/switches/brocade.py
```diff
@@ -277,8 +277,18 @@ def _construct_tag(name):
...
         return '{urn:brocade.com:mgmt:brocade-interface}%s' % name
 
-    def _make_request(self, method, url, data=None):
+    def _make_request(self, method, url, data=None, log=True):
         r = requests.request(method, url, data=data, auth=self._auth)
-        if r.status_code >= 400:
+
+        # error logging
+        # remove/delete something that's not attached/exists
+        if ((method == 'DELETE') and (not log) and (r.status_code == 404)):
```
knikolla: Not here, in the caller method.

naved001: In that case I would have to do the checking and the logging in the caller method. Because we are already doing error logging in `_make_request` I thought it would be redundant to do it in the caller method. 

Any specific reason why you want to do it in the caller method though?

knikolla: Because the caller method understands the request it is making, ``_make_request`` should just be a wrapper around making a generic request. And shouldn't concern itself with that logic.

```python
r = self._make_request('DELETE', url, log=False)
if r.status_code not in [200, 404]:
    logger.error('Bad Request to switch. Response: %s' % r.text)
```

Alternatively, and probably a better approach is to change ``_make_request`` to accept a list of acceptable error codes:

```python
def _make_request(self, method, url, data=None, acceptable_error_codes=None):
    acceptable_errors = acceptable_errors or []
    r = requests.request(method, url, data=data, auth=self._auth)
    if r.status_code >= 400 and r.status_code not in acceptable_error_codes:
        logger.error('Bad Request to switch. Response: %s' % r.text)
    return r
```

and call it from as ``self._make_request('DELETE', url, acceptable_error_codes=[404])``

naved001:
@knikolla makes sense,  I've made the changes you requested.

haas/ext/switches/brocade.py
```diff
@@ -277,8 +278,11 @@ def _construct_tag(name):
         """ Construct the xml tag by prepending the brocade tag prefix. """
         return '{urn:brocade.com:mgmt:brocade-interface}%s' % name
 
-    def _make_request(self, method, url, data=None):
+    def _make_request(self, method, url, data=None,
+                      acceptable_error_codes=None):
         r = requests.request(method, url, data=data, auth=self._auth)
-        if r.status_code >= 400:
+        acceptable_error_codes = acceptable_error_codes or []
+        if ((r.status_code >= 400) and
```
knikolla: no parentheses are necessary.

naved001: it just looked clear in my head. anyway, I changed it.

naved001:
@zenhack could you review this (or assign someone else)?

haas/ext/switches/brocade.py
```diff
@@ -277,8 +278,11 @@ def _construct_tag(name):
         """ Construct the xml tag by prepending the brocade tag prefix. """
         return '{urn:brocade.com:mgmt:brocade-interface}%s' % name
 
-    def _make_request(self, method, url, data=None):
+    def _make_request(self, method, url, data=None,
+                      acceptable_error_codes=None):
         r = requests.request(method, url, data=data, auth=self._auth)
-        if r.status_code >= 400:
+        acceptable_error_codes = acceptable_error_codes or []
```
zenhack: Instead of doing this, in the signature you can just define the default as `()` (empty tuple). Note that you should *not* define the default as `[]`, as lists are mutable and if we ever change this function to modify the argument, those modifications will propogate to future calls.

knikolla: Hadn't really thought about using a tuple. Good one!

naved001:
@zenhack done.

zenhack:
Cool, merging.

PR #753, opened 2017-04-07
Cisco Nexus switch: Save running configuration to flash memory 

naved001:
Addresses #688 

This modifies the nexus driver so that changes made to the switch are persistent. This PR is similar to #749 

@zenhack @SahilTikale @henn

### DIFF
--- a/haas/ext/switches/_console.py
+++ b/haas/ext/switches/_console.py
@@ -16,7 +16,7 @@
 from abc import ABCMeta, abstractmethod
 from haas.model import Port, NetworkAttachment
 import re
-
+from haas.config import cfg
 
 _CHANNEL_RE = re.compile(r'vlan/(\d+)')
 
@@ -118,6 +118,16 @@ class Session(object):
 
         self.exit_if_prompt()
         self.console.expect(self.config_prompt)
+
+    def _save_check(self, switch_type):
+        """checks the config file to see if switch should save or not"""
+
+        save = True
+        switch_ext = 'haas.ext.switches.' + switch_type
+        if (cfg.has_option(switch_ext, 'save') and
+           not cfg.getboolean(switch_ext, 'save')):
+            save = False
+        return save
 
 
 def get_prompts(console):
--- a/haas/ext/switches/_console.py
+++ b/haas/ext/switches/_console.py
@@ -119,15 +119,12 @@ class Session(object):
         self.exit_if_prompt()
         self.console.expect(self.config_prompt)
 
-    def _save_check(self, switch_type):
+    def _should_save(self, switch_type):
         """checks the config file to see if switch should save or not"""
 
-        save = True
         switch_ext = 'haas.ext.switches.' + switch_type
-        if (cfg.has_option(switch_ext, 'save') and
-           not cfg.getboolean(switch_ext, 'save')):
-            save = False
-        return save
+        return (not cfg.has_option(switch_ext, 'save') or
+                cfg.getboolean(switch_ext, 'save'))
 
 
 def get_prompts(console):
--- a/haas/ext/switches/_console.py
+++ b/haas/ext/switches/_console.py
@@ -123,8 +123,10 @@ class Session(object):
         """checks the config file to see if switch should save or not"""
 
         switch_ext = 'haas.ext.switches.' + switch_type
-        return (not cfg.has_option(switch_ext, 'save') or
-                cfg.getboolean(switch_ext, 'save'))
+        if cfg.has_option(switch_ext, 'save'):
+            if not cfg.getboolean(switch_ext, 'save'):
+                return False
+        return True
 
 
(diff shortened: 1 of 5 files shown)


zenhack:
Could we try to factor out some of the common code with the dell driver? Maybe move what we can into `_console.py`. This is mostly copypasta right now.

naved001:
Only the description in the haas.cfg seems like copypasta to me. The methods `_get_config()` and `_save_runnning_config()` are different from the dell driver. I am not sure how I would go about refactoring those.

zenhack:
@naved001, `disconnect` is almost identical; it's just the config section names that need to be different. (the other differences are incidental I think, and needn't be different). You're right though, there's less copypasta that I thought there was; just that and the comment.

naved001:
@zenhack I refactored some of the stuff. Let me know if this looks fine, or if you have any other suggestions.

haas/ext/switches/_console.py
```diff
@@ -119,6 +119,16 @@ def revert_port(self, port):
...
 
+    def _save_check(self, switch_type):
+        """checks the config file to see if switch should save or not"""
+
+        save = True
+        switch_ext = 'haas.ext.switches.' + switch_type
+        if (cfg.has_option(switch_ext, 'save') and
+           not cfg.getboolean(switch_ext, 'save')):
+            save = False
+        return save
```
zenhack: This function could be more concise; you can skip the assignment statements and just write:

```python
return (not cfg.has_option(switch_ext, 'save') or
        cfg.getboolean(switch_ext, 'save'))

File: haas/ext/switches/_console.py [hunk 2]
```diff
@@ -119,6 +119,16 @@ def revert_port(self, port):
         self.exit_if_prompt()
         self.console.expect(self.config_prompt)
 
+    def _save_check(self, switch_type):
```
zenhack: How about a name that makes the function's job a little more clear, e.g. `_should_save`? `_save_check` could be a number of other things, c.f. the check_call, check_output functions from the subprocess module.

haas/ext/switches/_console.py
naved001: done

haas/ext/switches/_console.py
```diff
@@ -119,6 +119,13 @@ def revert_port(self, port):
         self.exit_if_prompt()
         self.console.expect(self.config_prompt)
 
+    def _should_save(self, switch_type):
+        """checks the config file to see if switch should save or not"""
+
+        switch_ext = 'haas.ext.switches.' + switch_type
+        return (not cfg.has_option(switch_ext, 'save') or
```
henn: Could you make this an "if"? This is a little non-intuitive

haas/ext/switches/nexus.py
```diff
@@ -203,6 +205,30 @@ def get_port_networks(self, ports):
...
+        """returns the requested configuration file from the switch"""
+
+        self.console.sendline('terminal length 0')
+        self.console.expect(r'[\r\n]+.+# ')
+        self.console.sendline('show ' + config_type + '-config')
+        self.console.expect(r'[\r\n]+.+# ')
+        config = self.console.after
+        lines_to_remove = 9
+        # the startup config file always has a couple of extra lines that
+        # we need to remove.
```
henn: Could you add examples of the stuff you want to remove here?

haas/ext/switches/_console.py
naved001: If I used `if` I would have to do nested ifs which is [not as elegant](https://github.com/CCI-MOC/hil/pull/749#discussion_r108596058)

henn: I'm still not a fan - even though I agree that it is elegant, readable is more important in order to avoid .
It's too bad we haven't transitioned to python 3 yet -- we could just use the [fallback](https://docs.python.org/3/library/configparser.html#fallback-values) function in the new ConfigParser.

haas/ext/switches/nexus.py
henn: Would it be possible to exclude these based on text rather than always getting rid of a fixed number of lines?

haas/ext/switches/_console.py
zenhack: Yeah, the python2 configparser api is not great. I don't feel strongly on this point.

haas/ext/switches/nexus.py
naved001: @henn Done.

naved001:
Tests are failing because of #759

shwsun:
> Tests are failing because of #759
> 
> Edit: I ran it again and it passed this time.

Then it is the thing I am nervous about ... @zenhack @henn @SahilTikale

naved001:
@shwsun I think they are working on a patch to fix this.

PR #756, opened 2017-04-11
Close #306

### LINKED ISSUE #306: WIP Passwords present in logs
`moc.rest` is logging all API calls, with all parameters---and the `user_create` API call contains a password as a parameter.  So, `moc.rest` logs the password, too.

zenhack:
We might want to move the fail_on_log_warnings over to
pytest-catchlog as well.


tests/unit/rest.py
```diff
@@ -556,3 +563,21 @@ def increment():
...
+def test_dont_log(client, caplog):
+    @rest.rest_call('POST', '/some-path', Schema({
+        'public': basestring,
+        'private': basestring,
+        'stuff': basestring,
+    }), dont_log=('private',))
+    def some_call(public, private, stuff):
+        pass
+
+    with caplog.atLevel(logging.DEBUG):
```
kylehogan: Was looking up what this is and it looks like `atlevel` is [deprecated](https://github.com/eisensheng/pytest-catchlog/blob/84de498d7a80378e36645cd2907226efc24cac65/CHANGES.rst). Seems like it's the same functionality, but method names changed to follow pep8

setup.py
```diff
@@ -94,6 +94,7 @@ def _get_readme():
                         'pexpect>=3.3,<4.0',
                         'requests>=2.4.1,<3.0',
                         'pytest>=2.6.2,<3.0',
+                        'pytest-catchlog>=1.2.2,<2.0',
```
kylehogan: Is this just making sure the version of catchlog is what we're expecting? And if so isn't 1.2.2 the latest version (so why less than 2.0? Is it just in case they break things with updates?)

zenhack: See the comment about semver above. Yes, it's a compatibility thing.

tests/unit/rest.py
zenhack: Good catch, will fix.

PR #762, opened 2017-04-24
URGENT PATCH: Fixes tests of client library

SahilTikale:
Tests were taking unresonably long time in travis. 
Refactored some of the fixture code. Now running under acceptable time frame.
Prevously took over 5 minutes to finish, now takes less than a minute.

### DIFF
--- a/tests/unit/client.py
+++ b/tests/unit/client.py
@@ -280,6 +280,8 @@ def create_setup(request):
     def fin():
         proc1.terminate()
         proc2.terminate()
+        proc1.wait()
+        proc2.wait()
         cleanup(dir_names)
     request.addfinalizer(fin)
 


SahilTikale:
@zenhack can you have a look at the tests. some logger test is failing. I am not sure why.

zenhack:
Title is a little overdramatic.

1. Tests are failing, and it looks like not for the same reasons as before. I've stared at it a bit and I'm not totally sure what's going on; it looks like you're running something from the wrong directory or something, but I don't see any changes that would cause that.
2. I'm not merging something that just puts skip on a whole bunch of tests -- that kinda defeats the point.

Can you articulate why this was so slow?

zenhack:
(Incidentally, the bit that is actually somewhat urgent is the fact that the tests are *broken,* not that they're slow).

SahilTikale:
@zenhack did not intend to be over-dramatic, Just a little quality conscious and feeling responsible that my work will affect everyone else. so feeling a little sorry too. We can change the title if you wish.
I thought the fix was needed urgently.

zenhack:
@SahilTikale, worst case we'll revert the merge until we've actually got a proper fix -- the wonders of source control.

And again, they really urgent thing is the actual breakage -- the performance thing has been mounting for a long time, and it's clearly time to start dealing with it. But it's not urgent enough to cut corners.

SahilTikale:
The tests were never failing, let me assert that. They were slow because I was under a mistaken assumption that when we setup a fixture with scope='session' it re-uses the same database, and runs the tests against the same instance of the web server. After todays observation, I started looking at those tests closely and realized that was not the case. For each class of test, it was recreating the whole setup from making a new `tmpdir` to restarting a new HIL server. That is what was taking time. So I refactored the tests such that it does not do that anymore. The skipped tests are a smaller issue but I am chasing another (my qual exams) deadline and did not want to spend more time fixing consistency issues there, hence I simply skipped them for now. The idea was that you can have a look and let me know if this patch is good enough and I can, at later point, fix the skipped tests. If you feel that is not right. I can remove the tests for now and re-introduce them later when I have some free to work on it. I hope that explains the problem. 

Let me know what you think.

zenhack:
I think the best solution for avoiding port collision is to just *not start a real web server.* The client tests can use the flask test HTTP client. @knikolla complained about this as well, and I recall arguing about it during review.

zenhack:
But yes, proc.wait() is the right way to make sure the process is actually shut down.

SahilTikale:
I ran this build multiple times and it passed every time without any problem. 
I think this is a robust patch for now and is good for getting accepted. 

@zenhack @henn @shwsun Please review.

SahilTikale:
Apparently you cannot run rebuild beyond a few times if it builds successfully each time. I just re-based the PR since most other commits were redundant. Also this gives a chance to re-run the build few more times just to be sure.

henn:
> I think the best solution for avoiding port collision is to just not start a real web server. The client tests can use the flask test HTTP client

@zenhack - the team had a lengthy debate about this a while back, and decided to go with the full webserver version. I don't recall exactly why and can't find it in #617.

@SahilTikale - do you remember? We should document our reasoning.

SahilTikale:
The reasoning as I remember was this:

1. I need to put in dummy data and needed to customize the haas.cfg based on what I wish to test. This meant I wanted to have control on how the database is populated and wanted to keep it isolated from other tests. 
2. Later you had mentioned that you would like to parameterize the fixtures to do auth testing which will run same tests once against `auth=database` and then for `auth=keystone`
3. Also, This framework can be used to develop cli tests when we move forward to re-designing cli.
4. May not be important at this point but in the spirit of capturing what was discussed, having a separate web-server would help me write tests for recursiveHIL (rHaaS) in future.

SahilTikale:
@zenhack need one +1,
I ran the travis build more around 9 times. It build successfully every time. 
This should take care of the intermittent failure.

shwsun:
There are already 2 +1s, I assume you can merge it

henn:
@zenhack - do you have any remaining concerns for this PR?

@shwsun - we should wait a little bit for @zenhack, since he commented earlier.

SahilTikale:
Hi @zenhack can we merge this ?

PR #763, opened 2017-04-25
Fixes a typo in the client library

naved001:
appends `self` for node delete which was missing. @knikolla @SahilTikale


SahilTikale:
good catch. 
+1

Issue #757, opened 2017-04-19
CLI command to dump everything


henn:
Once the new CLI is integrated, we should add a command that dumps an entire HIL config.

Something like `hil show --all`.
This way, we can keep the complexity of recursively showing everything out of the API.

**Bonus points:** do something like `hil dump` and `hil restore`, where we can restore a configuration.

xuhang57:
PR merged. Close now and pls reopen if necessary.

naved001:
No, the cli call to dump everything still doesn't exist. What was merged are just API calls to fill in the gaps needed for a call like this.

PR #767, opened 2017-04-28
API to see what's connected to a port

naved001:
Currently there's no way of knowing what's connected to a port. Also, this API is needed for #757

Should we have a way of doing this the other way too i.e., checking the nic to see what port it is connected to?

### DIFF
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -1038,7 +1038,7 @@ Response body:
         "networks": {"vlan/1511": "mynetwork"}
     }
 
-In case there is no nic attached to port, the response body is just an empty
+If there is no nic attached to a port, the response body is just an empty
 json object:
 
     {}
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -1038,6 +1038,11 @@ Response body:
         "networks": {"vlan/1511": "mynetwork"}
     }
 
+In case there is no nic attached to port, the response body is just an empty
+json object:
+
+    {}
+
 Authorization requirements:
 
 * Administrative access.
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -1033,9 +1033,9 @@ Show the node and nic to which the port is connected.
 Response body:
 
     {
-        u'node': u'mynode',
-        u'nic': u'mynic',
-        u'networks': {u'vlan/1511': u'mynetwork'}
+        "node": "mynode",
+        "nic": "mynic",
+        "networks": {"vlan/1511": "mynetwork"}
     }
 
 Authorization requirements:
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -1030,6 +1030,14 @@ Possible errors:
 
 Show the node and nic to which the port is connected.
 
+Response body:
+
+    {
+        u'node': u'mynode',
+        u'nic': u'mynic',
+        u'networks': {u'vlan/1511': u'mynetwork'}
+    }
+
 Authorization requirements:
 
 * Administrative access.
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -1024,6 +1024,20 @@ Possible errors:
 * 404, if there is no nic attached to `port`
 * 409, if there is already a networking action pending on `port`
 
+#### show_port
+
+`GET /switch/<switch>/port/<port>`
+
+Show the node and nic to which the port is connected.
+
+Authorization requirements:
+
+* Administrative access.
+
+Possible errors:
+
+* 404, if the switch and/or port do not exist.
+
 ## API Extensions
 
 API calls provided by specific extensions. They may not exist in all
--- a/haas/api.py
+++ b/haas/api.py
@@ -967,6 +967,21 @@ def show_switch(switch):
     }, sort_keys=True)
 
 
+@rest_call('GET', '/switch/<switch>/<port>', Schema({
+    'switch': basestring, 'port': basestring}))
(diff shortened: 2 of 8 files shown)


henn:
This is great! Kudos for adding support to the client library!

I think we'll need a few things (mostly tests) for this to be ready for primetime:

- [ ] Include a `list_ports` so that scripts can enumerate the infrastructure
- [ ] For testing the API call directly: make sure we have both positive (check that it succeeds) and negative (test that it fails properly) tests. Currently, we only have positive tests
- [ ]  Verify authorization is working (both positive and negative)
- [ ] Include test for new client library call
- [ ] (optional) include network connectivity info for the NIC, similar to how show_node() includes networks connected to the NIC.

Regarding NIC-centric, we could either have a {list/show}_nic, or include it in the output of show_node. Either way works for me. Let's keep that as a separate PR, though, to keep the PR size down.

haas/api.py
```diff
@@ -967,6 +967,21 @@ def show_switch(switch):
...
+def show_port(switch, port):
+    """show port details on a switch.
+
+    Return a JSON object showing the node and nic to which port is connected.
+    """
+    get_auth_backend().require_admin()
+    switch = _must_find(model.Switch, switch)
+    port = _must_find_n(switch, model.Port, port)
+    nic = port.nic
+    return json.dumps({'node': nic.owner.label if nic else None, 'nic':
```
naved001: I feel that the API shouldn't return a `None`. I am returning a an empty object `{}`.

Also, if I return a None, then whatever is making that call can't blindly do [response.loads()](https://github.com/CCI-MOC/hil/blob/master/haas/client/base.py#L44) because then it says that "No JSON object could be decoded".

naved001:
>  Include a list_ports so that scripts can enumerate the infrastructure

`show_switch` already lists all the ports on a switch. Do we really need a separate `list_ports`?

> For testing the API call directly: make sure we have both positive (check that it succeeds) and negative (test that it fails properly) tests. Currently, we only have positive tests. 
> Test that can verify authorization is working (both positive and negative)

Done

> Include test for new client library call

I have written some tests, but I am not satisfied. I'll add more to it. 

> (optional) include network connectivity info for the NIC, similar to how show_node() includes networks connected to the NIC.

Done

> Similar to list_switches, we should specify what the output looks like.

Done

docs/rest_api.md
```diff
@@ -1024,6 +1024,28 @@ Possible errors:
...
+
+Show the node and nic to which the port is connected.
+
+Response body:
+
+    {
+        u'node': u'mynode',
+        u'nic': u'mynic',
+        u'networks': {u'vlan/1511': u'mynetwork'}
+    }
```
zenhack: This should be displayed as JSON, not a python literal, since this is the HTTP response body. So, get rid of the `u` prefix, and use double quotes.

tests/unit/api/main.py
```diff
@@ -1486,6 +1486,30 @@ def test_show_switch(self, switchinit):
...
+            api.show_port('non-existing-switch', 'some-port')
+        # call show port on a port that's not registered on that switch
+        with pytest.raises(api.NotFoundError):
+            api.show_port('sw0', 'non-existing-port')
+
+        assert json.loads(api.show_port('sw0', '3')) == {}
+        # connect the port to a nic, and see if show port agrees
+        api.port_connect_nic('sw0', '3', 'compute-01', 'eth0')
+        assert json.loads(api.show_port('sw0', '3')) == {
+                        u'node': u'compute-01', u'nic': u'eth0',
```
zenhack: The formatting makes this a bit hard to read, could you move nic/eth0 to it's own line?

tests/unit/client.py
```diff
@@ -505,6 +505,21 @@ def test_port_detach_nic_error(self):
...
+    def test_show_port(self):
+        # do show port on a port that's not registered yet
+        with pytest.raises(FailedAPICallException):
+            C.port.show('mock-01', 'gi1/1/8')
+
+        C.port.register('mock-01', 'gi1/1/8')
+        assert C.port.show('mock-01', 'gi1/1/8') == {}
+        C.port.connect_nic('mock-01', 'gi1/1/8', 'node-09', 'eth0')
+        assert C.port.show('mock-01', 'gi1/1/8') == {
+                u'node': u'node-09', u'nic': u'eth0', u'networks': {}}
```
zenhack: Same thing here, I have a hard time scanning dictionaries where you have multiple key/value pairs on one line.

docs/rest_api.md
```diff
@@ -1024,6 +1024,28 @@ Possible errors:
...
+
+Show the node and nic to which the port is connected.
+
+Response body:
+
+    {
+        "node": "mynode",
+        "nic": "mynic",
+        "networks": {"vlan/1511": "mynetwork"}
+    }
```
henn: Could we also specify what a valid port looks like when no NIC is attached?

henn:
> show_switch already lists all the ports on a switch. Do we really need a separate list_ports?

I didn't realize that show_switch already did this. I agree - we don't need a list_ports.

henn:
FYI - it looks like the build was breaking due to a bug that was fixed 2 hours ago in python requests ( https://github.com/kennethreitz/requests/pull/4007). I restarted it.

docs/rest_api.md
naved001: Done

naved001:
@zenhack @henn I've made the changes you requested.

naved001:
@henn, ping

PR #772, opened 2017-05-11
Driver for Dell N3048

naved001:
Add support for the Dell N 3048 driver. 

Since the driver is similar to the PowerConnect 5500 we support, I decided to restructure the old dell driver to reuse code. There is more stuff that we can reuse (like `get_port_networks`) but I'll save it for a different PR. 

Deployment tests pass on all the switches that I touched in this PR (Dell PowerConnect 5500, Dell N 3048, and the Cisco Nexus). 

This switch wouldn't let me disable the native vlan, so I made it like the nexus switch where we set the native vlan to some dummy vlan but in the case of this switch that vlan must be created otherwise we can't set it as native.


docs/network-drivers.md
```diff
@@ -70,20 +70,21 @@ Doing so is not difficult, and it is critical for security.
...
 
 * ``haas.ext.switches.dell``, which provides a driver for the Dell
-  Powerconnect 5500 series switches.
+  Powerconnect 5500 series switches and the Dell N3000 series.
 * ``haas.ext.switches.nexus``, which provides a driver for some Cisco
   Nexus switches. Only the 3500 and 5500 have been tested, though it is
   possible that other models will work as well.
+* ``haas.ext.switches.brocade``, for the brocade VDX 6740.
 
 Neither driver requires any extension-specific config options. Per the
```
zenhack: s/Neither driver requires/None of the drivers require/

docs/network-drivers.md
```diff
@@ -114,6 +115,25 @@ Port names must be of the same form accepted by the switch's console
...
+
+#### switch_register
+
+To register a Dell N3000 switch, the ``"type"`` field of the
+request body must have a value of::
+
+    http://schema.massopencloud.org/haas/v0/switches/delln3000
+
+It requires the same additional fields as the powerconnect driver, plus an
+additional field "dummy_vlan" like the nexus driver, which should a JSON
```
zenhack: Missing a word here (shoud *be* a JSON...)?

docs/network-drivers.md
```diff
@@ -114,6 +115,25 @@ Port names must be of the same form accepted by the switch's console
...
+number corresponding to an unused VLAN id on the switch. Unlike the nexus
+switch, this dummy vlan must exists otherwise you cannot set it as the native
```
zenhack: s/exists/exist/

haas/ext/switches/_console.py
```diff
@@ -73,6 +73,9 @@ def disable_port(self):
     def disconnect(self):
         """End the session. Must be at the main prompt."""
 
+    def _set_native(self, old_native, network_id, interface):
```
zenhack: You shoud make it more clear that this method is intended to be overridden -- I spend a bit of time puzzling as to why there was an `interface` parameter that wasn't used. Most of the other methods that are overridden also don't actually have implementations (empty bodys).

haas/ext/switches/_dell_base.py
```diff
@@ -0,0 +1,155 @@
...
+import pexpect
+import re
+import logging
+
+from haas.ext.switches import _console
+
+logger = logging.getLogger(__name__)
+
+
+class _base_session(_console.Session):
```
zenhack: Should follow the naming conventions for types, e.g. `_BaseSession`.

haas/ext/switches/dell.py
```diff
@@ -83,60 +113,129 @@ def connect(switch):
...
                 num_str = match.groups()[0]
                 native = int(num_str)
+                if native == int(self.switch.dummy_vlan):
+                    native = None
             else:
                 native = None
             networks = []
-            range_str = v['Trunking VLANs Enabled']
-            for range_str in v['Trunking VLANs Enabled'].split(','):
+            range_str = v['Trunking Mode VLANs Enabled']
```
zenhack: I realize this was already there, but -- `range_str` is reassigned on the next line, so this doesn't do anything.

haas/ext/switches/dell.py
```diff
@@ -54,10 +54,40 @@ def validate(kwargs):
...
 
-class _Session(_console.Session):
+    @staticmethod
+    def validate(kwargs):
+        schema.Schema({
+            'username': basestring,
+            'hostname': basestring,
+            'password': basestring,
+            'dummy_vlan': schema.And(schema.Use(int),
+                                     lambda v: 0 <= v and v <= 4093,
```
zenhack: We probably want to exclude zero?

haas/ext/switches/_console.py
```diff
@@ -74,6 +74,11 @@ def disconnect(self):
         """End the session. Must be at the main prompt."""
 
     def _set_native(self, old_native, network_id, interface):
+        """This method is intended to be over written.
+
+        In some switches, we need to create a vlan before setting that as
+        native. `interface` is needed to get back to the interface prompt
+        to set the native vlan for that interface"""
```
zenhack: This smells funky to me. You're basically shoving the whole implementation in the docstring; there's got to be a cleaner way to define/describe the interface.

naved001: I can't think of a more succinct docstring than this:

        """This method can be overwritten to pass `interface`

        `interface` is used to get to the interface prompt"""

What do you think?

naved001:
@zenhack @henn ping. 

I want to finish this while it's fresh in my head.

haas/ext/switches/_console.py
zenhack: Honestly, I think the problem is that the control flow is somewhat slapped together. The console code I wrote had a particular way of managing the prompts, which this doesn't fit with, and that mismatch is showing up in the function signatures.

Question: can we just put in the setup docs for the switch that you have to create all of the vlans you intend on using for HIL, and cut out this step entirely?

(Sorry for letting this fall through the cracks).

docs/network-drivers.md
```diff
@@ -70,20 +70,21 @@ Doing so is not difficult, and it is critical for security.
 ## Switch drivers
 
 At present, all switch drivers shipped with HaaS require that the VLAN
-pool allocator is in use. There are two switch drivers shipped with
+pool allocator is in use. There are three switch drivers shipped with
 HaaS:
 
 * ``haas.ext.switches.dell``, which provides a driver for the Dell
-  Powerconnect 5500 series switches.
+  Powerconnect 5500 series switches and the Dell N3000 series.
```
zenhack: These are separate modules now (the n3000 is in `haas.ext.switches.n3000`), so this needs updating.

haas/ext/switches/_console.py
naved001: > Question: can we just put in the setup docs for the switch that you have to create all of the vlans you intend on using for HIL, and cut out this step entirely?

Sure. That's actually easier. I didn't know that we are already doing something like this for the other switches. Had I read the documentation before I started working on this driver, I would have cut out this step.

zenhack:
LGTM

docs/network-drivers.md
```diff
@@ -114,6 +117,33 @@ Port names must be of the same form accepted by the switch's console
...
 
+### Dell N3000 driver
+
+#### Switch preperation
+
+1. Just like the Powerconnect 5500, every VLAN that could be used on the switch
+must first be enabled on the switch. To enable all VLANs to work with the switch, run this command:
+
+   configure
+   vlan 2-4093
```
shwsun: These two lines are commands, please format it

shwsun:
And since there are so many unnamed user in the commits, we should squash it

docs/network-drivers.md
naved001: done.

naved001:
> can you explain a bit about the abstracted _BaseSession

Earlier we had a `_console.py` where we had some common methods for console based drivers (at that point, cisco nexus and dell powerconnect). With this PR we wanted to support the Dell N3048 which was quite similar to the powerconnect. And here we move the session object (which has basically all the methods for the dell powerconnect), then import that class into 2 drivers (the powerconnect and the new n3048) and override the functions as needed,. In the process of doing so, I have also modified some methods so that they work with both dell switches. 
```
 [_Console.py]
   +      +-+
   |        |
   v        v
[nexus.py]  [_dell_base.py]
             +       +
             |       |
             v       v
        [n3048.py]   [dell.py]
```

> it would be good to doc it.

I don't think we are documenting code structure anywhere. The files that I have made have a [description](https://github.com/CCI-MOC/hil/pull/772/files#diff-66fd971d942f05993b034b18448fd45fR14) about it's purpose. Additionally, in future, someone can run git blame and find out about this PR to get some context. 

> And since there are so many unnamed user in the commits, we should squash it

The commits are either associated with my name (Naved Ansari) or my username (naved001), I don't see any orphan commits(?). Whoever is gonna merge can go ahead and use the squash and merge button if that's an issue. 

Let me know if you have more questions or want to play with the switch and test the driver in person.

docs/network-drivers.md
```diff
@@ -114,6 +117,35 @@ Port names must be of the same form accepted by the switch's console
...
+
+It requires the same fields as the powerconnect driver, plus an
+additional field "dummy_vlan" like the nexus driver, which should be a JSON
+number corresponding to an unused VLAN id on the switch. Unlike the nexus
+switch, this dummy vlan must exist otherwise you cannot set it as the native
+vlan, but this vlan should not be used for any networks.
+
+#### switch_register_port
+
+Resgister ports just like the powerconnect driver. e.g. ``gi1/0/5``.
```
knikolla: nit: Resgister

naved001:
Let me look into the db migration script for this before we merge.

naved001:
@zenhack @knikolla I have added script to support migrations. If you are fine with the new changes you can remove the DO NOT MERGE YET label.

zenhack:
Migration script looks good to me.

//cc @knikolla

zenhack:
Cool, we've got 2 signoffs, merging.

PR #774, opened 2017-05-11
Revamp and fix client lib tests to prevent intermittent failures; move to py.test 3.0

henn:
The removal of `--boxed` in setup.cfg was of great aid while debugging things. It's necessary in order for the test to capture stdout/stderr output.

py.test 3.0 gives us the "yield" syntax for fixtures, which lets us pair the teardown with the setup without register a finalizer.


zenhack:
We put --boxed in for a reason, I think to avoid having tests pollute each others' environment. Iirc, this was around the time we added extensions.

I'll look at the rest tomorrow.

tests/unit/client.py
```diff
@@ -485,6 +540,7 @@ def test_port_delete_error(self):
 
     def test_port_connect_nic(self):
         C.port.register('mock-01', 'gi1/1/5')
+        C.node.add_nic('node-08', 'eth0', 'aa:bb:cc:dd:ee:ff')
```
zenhack: Not sure I understand why you added this (and tweaked a few other functional parts of the tests)? Perhaps this is related to the travis failures?

henn:
I believe I've addressed your concerns with my latest push, with the
exception of fixing this warning:

```pytest

WC1 None pytest_funcarg__cov: declaring fixtures using
"pytest_funcarg__" prefix is deprecated and scheduled to be removed in
pytest 4.0.  Please remove the prefix and use the @pytest.fixture
decorator instead.

```

Any ideas on that one? I can't find `pytest_funcarg` in any of our sources.


RE: the test_port_connect_nic() issue, it might have been because I
removed `--boxed` which stopped resetting state between tests.


> That said, I'm not strongly opposed to raising the upper bound if it's
otherwise non-intrusive, but if we're doing that I'd like to fix those
warnings at least.

I figured this was a good time to upgrade. If nobody does it like this,
then I suspect HIL's use of the dependencies will begin to bitrot and
leave us stranded on old versions (not unlike the warnings we got going
to the newest pytest).

zenhack:
I'm certainly on board with keeping deps up to date, just need to make sure the upgrade gets handled properly.

I poked around a bit re: the warning, but I think I'm running out of steam for the night; I'll have another look later.

henn:
@zenhack @SahilTikale  - comments on the latest code?

tests/unit/client.py
```diff
@@ -493,7 +547,8 @@ def test_port_connect_nic_error(self):
         C.port.register('mock-01', 'gi1/1/6')
         C.port.connect_nic('mock-01', 'gi1/1/6', 'node-09', 'eth0')
         with pytest.raises(FailedAPICallException):
-            C.port.connect_nic('mock-01', 'gi1/1/6', 'node-08', 'eth0')
+            C.port.connect_nic('mock-01', 'gi1/1/6', 'node-09', 'eth0')
```
naved001: Why change this here? What the original test seems to do is connect a port (gi1/1/6 on mock-01), that's already connected to a nic (eth0 on node-**09**), to another nic on a different node (eth0 on node-**08**). That would still raise the error correctly.

tests/unit/client.py
```diff
@@ -262,28 +268,69 @@ def avoid_network_race_condition():
...
+        sess.get(url + '/projects')
+        return True
+    except:
+        return False
+
+
+def wait_for_service(url, timeout=15):
+    """
+    Waits for a port to become a working http server
+    timeout -- number of seconds to wait (default 60)
```
zenhack: Docstring says default 60, argument says 15. make these consistent.

tests/unit/client.py
```diff
@@ -262,28 +268,69 @@ def avoid_network_race_condition():
...
+    """
+
+    begin = time.time()
+    while True:
+        try:
+            sess = requests.Session()
+            sess.auth = (username, password)
+            sess.get(url + '/projects')
+            break
```
zenhack: Why not just use `port_has_server` here?

tests/unit/client.py
zenhack: I'd wondered about this as well.

setup.py
```diff
@@ -93,9 +93,9 @@ def _get_readme():
                         'passlib>=1.6.2,<2.0',
                         'pexpect>=3.3,<4.0',
                         'requests>=2.4.1,<3.0',
-                        'pytest>=2.6.2,<3.0',
+                        'pytest>=2.6.2,<4.0',
```
SahilTikale: Does this affect the current PR ? Same for the next change in this file.

tests/unit/client.py
```diff
@@ -262,28 +268,69 @@ def avoid_network_race_condition():
     for timeout in range(10):
         que = db.session.query(NetworkingAction).count()
         if (que > 0):
-            time.sleep(1)
+            time.sleep(.5)
```
SahilTikale: I think this is too short a time interval. I did some tests with different time intervals. It usually takes like 6-ish seconds. 
I think 1 sec is a good interval. Just saying,  no need to change code more than necessary.

tests/unit/client.py
```diff
@@ -106,6 +111,7 @@ def make_config():
 
         ])
         f.write(config)
+        f.close()
```
SahilTikale: This call is not required. when doing `with open....` it automatically closes the file when done. 
remove this line please.

tests/unit/client.py
```diff
@@ -262,28 +268,69 @@ def avoid_network_race_condition():
...
     populate_server()
 
-    def fin():
-        proc1.terminate()
-        proc2.terminate()
-        proc1.wait()
-        proc2.wait()
-        cleanup(dir_names)
-    request.addfinalizer(fin)
+    yield
```
SahilTikale: why was this changed ?

setup.cfg
```diff
@@ -1,7 +1,10 @@
 [easy_install]
 
-[pytest]
+[tool:pytest]
```
SahilTikale: what does this do ?

naved001:
@henn, ping.

setup.cfg
henn: pytest 3.0 throws a warning if you don't

setup.py
henn: Please see Ian and I's discussion in the comments. We should be updating these all the time.

tests/unit/client.py
henn: Since this is done a number of times, and the client tests increased the runtime by 300%, I figured reducing sleep time where possible can help.

Increased the range() to meet your concern.

tests/unit/client.py
henn: Done!

tests/unit/client.py
henn: I wanted to catch **only** `ConnectionError` so as not to mask other potential exceptions coming from requests. port_has_server() catches everything.

tests/unit/client.py
henn: Pytest [recommends](https://docs.pytest.org/en/latest/fixture.html) using this syntax over addfinalizer(). I think it's much cleaner myself.

setup.cfg
henn: Search for `tool:pytest` in the [pytest changelog](https://docs.pytest.org/en/latest/changelog.html#id76) for more info.

tests/unit/client.py
zenhack: I tend to agree -- this was one of a number of things on your PR that I might have suggested, but didn't feel strongly enough about to add to the review time. We're using yield-style fixtures almost everywhere else where we're doing cleanup.

tests/unit/client.py
zenhack: I missed that -- it probably *shouldn't* do that, for exactly the same reason. Would prefer to fix the problem there and then use it here.

Note that sess.get isn't going to raise an exception as long as the HTTP request works, regardless of what response it gets back.

henn:
@zenhack @SahilTikale - is there anything you are waiting on me for? I believe I addressed the substantive review comments in e1c14d6

zenhack:
@henn:

https://github.com/CCI-MOC/hil/pull/774#discussion_r118798070

tests/unit/client.py
SahilTikale: Please remove `f.close()`. It is redundant.

zenhack: @SahilTikale, I think you're looking at an old diff; I don't see the statement on @henn's most recent commit.

henn:
@SahilTikale - please take a look at the latest. I removed the `f.close()`

SahilTikale:
I think I was looking at an old diff. I guess I do not have anything more.

henn:
@SahilTikale - thanks!

@zenhack - what do you think of the refactored version?

zenhack:
LGTM, merging.

Issue #447, opened 2015-06-22
No show_port command


lokI8:
Unclear how to attain the running configuration of the switch through haas. Would like a command like "show_port" that returns the current interface mode, network identifier applied, and nics associated.

zenhack:
Quoting Jonathan Bell (2015-06-22 14:49:45)

>    Unclear how to attain the running configuration of the switch through
>    haas. Would like a command like "show_port" that returns the current
>    interface mode, network identifier applied, and nics associated.

This will be a lot easier once my refactoring work is done (and we'll 
probably want a show_switch) or something as well.

zenhack:
Re: show_switch, since these are polymorphic, I can think of a couple different ways of dealing with them:

1. Add a method to Switch that subclasses are required to implement, which formats the data accordingly.
2. Just show information that's common to all switches (name, type, list of ports)
3. Add a method to Switch that does (2), but let subclasses override it if they want to provide more information.

There's also a question of how much information should show_switch reveal. credentials should probably be off-limits, at least.

henn:
Merging this with #472. Comments there included:

Useful info I can think of in `GET /switch/<switch>` would be:
* The switch type
* Every port on the switch, the NIC+Node it is connected to, and the output of `show_nic`. This could be helpful in debugging connectivity problems.

Useful info for `GET /switch` would be a list of every switch HaaS knows about (optionally recursively showing info).

RE: polymorphism, I lean towards option 2 - a generic driver-agnostic version that shows stuff from the DB. Is there important driver-specific stuff?

zenhack:
I can't think of anything that the current drivers might want to  report; if nobody else can I'm happy with option (2) --- it's certainly the simplest. @gsilvis, @lokI8?

henn:
@gsilvis, @loki8: any thoughts on @zenhack's comment?

We are still needing to fish this data out of the DB.

henn:
It might also be beneficial to add port information to `show_node()` for debugging purposes.

It could then automatically be shown when one does a `show_network()` if #496 decides to include it verbatim.

naved001:
Fixed by #767 and #780

PR #780, opened 2017-05-19
Add port details to show_node

naved001:
Now the user can see what port and switch the node is connected to. It's a part of the `show_node` command. 

The [comment](https://github.com/CCI-MOC/hil/pull/767#issuecomment-298133016) and PR #767 that led to this.

### DIFF
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -537,11 +537,31 @@ Response body:
 	 "metadata":{"EK":"pk"}
 	}
 
+
+Response body if run by an admin:
+
+    {"name": "node1",
+	 "project": "project1",
+         "nics": [{"label": "nic1",
+           "port": "gi1/0/1",
+           "switch": "dell-01",
+	 	   "macaddr": "01:23:45:67:89", 
+		   "networks": {"vlan/native": "pxe", "vlan/235": "storage"}},
+                  {"label": "nic2",
+                   "port": None,
+                   "switch": None,
+		   "macaddr": "12:34:56:78:90", 
+		   "networks":{"vlan/native": "public"}}],
+	 "metadata":{"EK":"pk"}
+	}
+
 Authorization requirements:
 
 * If the node is free, no special access is required.
 * Otherwise, access to the project to which `<node>` is assigned is
   required.
+* If a user is an admin, then information about the port and switch to which
+  the nic is connected to is also printed.
 
 ### Projects
 
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -523,9 +523,11 @@ The object will have at least the following fields:
                 - "macaddr", the nic's mac address.
 		- "networks", a JSON object describing what networks are attached to the nic. The keys are channels and the values are the names of networks attached to those channels.
         - "port", the port to which the nic is connected to or null if the nic
-          is not connected to any port.
+          is not connected to any port. This field is only visibile if the
+          caller is an admin.
         - "switch", the switch that has the port to which the nic is connected
-          to or null if the nic is not connected to any port.
+          to or null if the nic is not connected to any port. Just like port,
+          this is only visible if the caller is an admin.
 	* "metadata", a dictionary of metadata objects
 
 Response body when run by a non-admin user:
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -522,46 +522,73 @@ The object will have at least the following fields:
                 - "label", the nic's label.
                 - "macaddr", the nic's mac address.
 		- "networks", a JSON object describing what networks are attached to the nic. The keys are channels and the values are the names of networks attached to those channels.
+        - "port", the port to which the nic is connected to or null if the nic
+          is not connected to any port.
+        - "switch", the switch that has the port to which the nic is connected
+          to or null if the nic is not connected to any port.
 	* "metadata", a dictionary of metadata objects
 
-Response body:
+Response body when run by a non-admin user:
 
-    {"name": "node1",
-	 "project": "project1",
-         "nics": [{"label": "nic1", 
-	 	   "macaddr": "01:23:45:67:89", 
-		   "networks": {"vlan/native": "pxe", "vlan/235": "storage"}},
-                  {"label": "nic2", 
-		   "macaddr": "12:34:56:78:90", 
-		   "networks":{"vlan/native": "public"}}],
-	 "metadata":{"EK":"pk"}
-	}
+    {
+        "metadata": {
+            "EK": "pk"
+        },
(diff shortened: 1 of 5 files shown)


zenhack:
A concern: up until now non-admin users have been unable to view exact details of switch topology.

We've actually talked about exposing that as something that we may actively want to do (the MOC in general is emphatically positive on letting users understand the platform they're running on), but I still feel we should be deliberate about what we expose and how.

What was the specific motivation for this patch?

naved001:
> What was the specific motivation for this patch?

The folllowing comment by @henn on #767 
> Regarding NIC-centric, we could either have a {list/show}_nic, or include it in the output of show_node. Either way works for me. Let's keep that as a separate PR, though, to keep the PR size down.

And also, because there's no straightforward way to see what port a node is  connected to. An admin would have to go into the database and look at the tables to figure this out. Although, we now have `show_port`,  we would still have to iterate over all the ports and see if our node is connected to one of the ports. 

 > A concern: up until now non-admin users have been unable to view exact details of switch topology

Maybe, we can hide away details based on who's running the command? Or have a separate admin command `list_nic` <node> that prints out all the nics of a node and the ports they are connected to?

What we ultimately want is a CLI command to dump everything #757
cc: @henn

haas/cli.py
```diff
@@ -718,7 +718,7 @@ def show_node(node):
 
     q = check_clientlib_response(lambda: C.node.show(node))
     for item in q.items():
-        sys.stdout.write("%s\t  :  %s\n" % (item[0], item[1]))
+        print item
```
kylehogan: seems like most of  cli is using `sys.stdout.write` why switching only this one to `print`?

zenhack:
Only adding tgis info for admins sounds reasonable.

You need to update rest_api.md.

haas/cli.py
naved001: Not sure why I changed i. Anyway,  changed it back to `sys.stdout.write` for consistency.

naved001:
I don't know if there's a way to have optional key value pairs in a dictionary. So I first add everything and then get rid of port and switch information if the user is not an admin.

naved001:
With this and #767 we can close #447

docs/rest_api.md
```diff
@@ -537,11 +537,31 @@ Response body:
 	 "metadata":{"EK":"pk"}
 	}
 
+
+Response body if run by an admin:
```
zenhack: * I might make these examples more symmetric (maybe by adding "if run by a non-admin" to the other one).
* Mention the new fields in the description of the result, not just in an example.

docs/rest_api.md
```diff
@@ -537,11 +537,31 @@ Response body:
...
+	 "metadata":{"EK":"pk"}
+	}
+
 Authorization requirements:
 
 * If the node is free, no special access is required.
 * Otherwise, access to the project to which `<node>` is assigned is
   required.
+* If a user is an admin, then information about the port and switch to which
+  the nic is connected to is also printed.
```
zenhack: This is organizationally a bit odd; I'd put the extra fields in the description above, and just say "admin access to view port/switch info" or such.

naved001:
@zenhack done

docs/rest_api.md
```diff
@@ -533,7 +533,27 @@ Response body:
...
+	 "metadata":{"EK":"pk"}
+	}
+
+
+Response body when run by an admin:
+
+There are two additional fields, `port` and `switch`.
+
+    {"name": "node1",
+	 "project": "project1",
```
zenhack: Indentation is rather wonky. I could spend time picking it apart, but if you just pipe it through `python -m json.tool` it will output something reasonable. Same goes for the other one.

docs/rest_api.md
```diff
@@ -533,7 +533,27 @@ Response body:
...
+         "nics": [{"label": "nic1",
+           "port": "gi1/0/1",
+           "switch": "dell-01",
+	 	   "macaddr": "01:23:45:67:89", 
+		   "networks": {"vlan/native": "pxe", "vlan/235": "storage"}},
+                  {"label": "nic2",
+                   "port": None,
+                   "switch": None,
```
zenhack: So, this is python notation, it should be JSON. Most of it is similar, but `None` needs to be changed to `null`. Running it through `python -m json.tool` will also serve to validate it.

docs/rest_api.md
```diff
@@ -533,7 +533,27 @@ Response body:
...
 		   "macaddr": "12:34:56:78:90", 
-		   "networks":{"vlan/native": "public"}}],
+		   "networks":{}}],
+	 "metadata":{"EK":"pk"}
+	}
+
+
+Response body when run by an admin:
+
+There are two additional fields, `port` and `switch`.
```
zenhack: I'd meant for this to be up with the rest of the prose describing the response, maybe around here:

https://github.com/CCI-MOC/hil/pull/780/files#diff-0d56cc599e156c786c59f0d2bfafb9ddL523

naved001: yeah, makes more sense to be there. I totally overlooked that earlier.

docs/rest_api.md
naved001: that's a cool tool. thanks for letting me know!

naved001:
@zenhack made the changes you requested. feel free to let me know if there's something that still doesn't look good.

docs/rest_api.md
```diff
@@ -522,46 +522,73 @@ The object will have at least the following fields:
                 - "label", the nic's label.
                 - "macaddr", the nic's mac address.
 		- "networks", a JSON object describing what networks are attached to the nic. The keys are channels and the values are the names of networks attached to those channels.
+        - "port", the port to which the nic is connected to or null if the nic
+          is not connected to any port.
+        - "switch", the switch that has the port to which the nic is connected
+          to or null if the nic is not connected to any port.
```
zenhack: Need to have the bit about "if the caller is an admin..." in there somewhere.

Sorry to keep ping-ponging you...

naved001: > Need to have the bit about "if the caller is an admin..." in there somewhere.

It's in the [authorization requirements](https://github.com/CCI-MOC/hil/pull/780/files#diff-0d56cc599e156c786c59f0d2bfafb9ddR591) Should we have that in the description too?

> Sorry to keep ping-ponging you...

zenhack: Yeah, I think so -- otherwise someone might read this section and then wonder why the fields weren't showing up.

naved001: @zenhack cool. added that bit.

zenhack:
Looks good to me. Merging.

PR #782, opened 2017-05-23
Refuse to run as root

### LINKED ISSUE #387: "haas init_db" should run as haas_user if run as root
If running `haas init_db` as root, the database is created as root and the haas_user cannot access it properly.

init_db should detect if running as root, and if so either su to the haas user (`os.setuid()`) or chown the file as appropriate.

zenhack:
Fixes #387

Note that I didn't bother with checking the db permissions; actually getting that right is a bit finicky, and refusing to run as root will prevent most cases of that in the first place. (As discussed there, given that this only affects sqlite, it's unlikely to come up much in practice, since  folks will mostly be using postgres in prod).

Re: testing, I did manual verification of this (both positive and negitive, with both affected commands), but thought it not worth trying to write unit tests that run the commands as root -- the impact of a bug is relatively low, and it complicates the test infrastructure a bit.

//cc @henn, @naved001, @knikolla

### DIFF
--- a/haas/cli.py
+++ b/haas/cli.py
@@ -15,6 +15,7 @@
 """This module implements the HaaS command line tool."""
 from haas import config, server, migrations
 from haas.config import cfg
+from haas.commands.util import ensure_not_root
 
 import inspect
 import json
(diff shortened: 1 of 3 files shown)


henn:
FYI - I think the test failures are addressed by #774

naved001:
@shwsun this is a small PR. can you review this?

zenhack:
Cool, I'll hit merge then.

PR #783, opened 2017-05-25
New feature specification Documentation

SahilTikale:


zenhack:
What is this, and why?

henn:
> What is this, and why?
We realized that:

a) we could debate major feature designs using the PR process before a
bunch of code was written in order to reduce wasted effort, and

b) it would result in documentation, including the debate itself.

zenhack:
At most, some of this info should be moved to a wiki article discussing what information to put in a proposal for a major feature.

...but I find attempts to discuss major features up front already end up getting too bogged down, and I'm not convinced adding more formal process is a good idea.

My experience is that throwing a wall of text at people mostly just results in them not finding time to read it, and having to ping them repeatedly to actually get anyone to comment. We need to be having *conversations* about development, not making a habit of writing documents we could show to bureaucrats to prove how professional we are. I don't think that was anyone's intent, but I worry that's basically what this amounts to. Detailed design documents have their place, but they should come out of actual need, and if we're needing this level of process with enough frequency to actually get used to it, we really ought to learn to break up our patches.

naved001:
I think this was based off of how openstack has [Specs template](https://specs.openstack.org/openstack/nova-specs/specs/kilo/template.html). @knikolla suggested that we have it for HIL too. Since, we are having a disagreement here, @knikolla, do you want to make an argument on why we should have it?

zenhack:
We are a much much *much* smaller team than openstack, and should accordingly have lighter-weight processes. It feels like it's not addressing a real need. The few times we've actually had a feature that needed a proposal, we've just done an ad-hoc "I'll write something and post to the wiki." I don't think we need more formality than that.

henn:
@zenhack - we've been burned at least a couple times where someone wrote a bunch of code, and then we had an architectural disagreement and a bunch of rewriting had to be done (you might have been one of them?).

Note that specs would only be for **major** features, not minor things, and we wouldn't require extensive stuff -- just enough so we can debate the feature and people can look back on it.

The rest of the team agrees on this, including @okrieg (who was initially skeptical).

zenhack:
So, one concrete piece of feedback then: `docs/features/readme.md` needs to make it clear that this is only for *major* features, as right now it doesn't read that way at all.

I'm willing to be convinced. I'm not sure some version of this is a bad idea in the abstract, but I think it's mis-understanding some of the problems it's trying to address in our particular case. Below is a distillation of my thoughts on the relevant problems in our process

* The point about a paper trail is a good one, and in fact it's something I feel very strongly about, but this doesn't really acknowledge the reason we don't have one: you guys keep discussing these things off-line. Part of the reason I insist on being remote is that it helps with this -- any conversation you guys have with me is going to be recorded, because it's going to happen on Github, or maybe the mailing list or IRC. The formalism of the proposed process won't help if people habitually talk about the proposal in person -- the debate still won't get recorded, just the final result. A simple rule of "keep all architectural conversations on Github" would solve this completely, without any extra formalism.
* Re: wasted effort due to coding first: in principle I agree, but my experience has been that when I write a spec like this and post it to some issue, getting people to pick through it and discuss is like pulling teeth. It's often tempting to just say screw it, implement the feature, and then go on with my life until you guys find time to comment on the result. An extreme example of this is the recent [console support thing][1] I posted, which only @naved001 has responded to so far. Even with the (very real) risk of wasted effort, it's often less draining for me to do it this way. I get more trigger-happy with this kind of thing as time goes on. Making this officially against policy won't solve the issue.

It's also worth noting a common pattern, that's a combination of both above disfunctions, and that this issue itself is guilty of:

* You all have a conversation off-line and come to a consensus.
* A PR shows up with the result, with no prior disucssion on the record.
* You all get to have the same argument with me.

Given a healthy culture of discussing features beforehand and on the record, I don't think having a template handy and keeping things in the repo/using the PR functionality to discuss is necessarily a bad idea. I just think this represents a fundamental misunderstanding of where the problems in our process lie.

[1]: https://github.com/CCI-MOC/hil/issues/417#issuecomment-303564763

pjd-nu:
Note that face to face meetings are an important part of the process - mentoring junior people is much harder without them, as is coming to consensus on certain types of problems. 

Many of the meetings (e.g. the weekly HIL meeting) have remote access; however we can probably be better about this as well as making sure there's a paper trail for decisions made.

zenhack:
Quoting Peter Desnoyers (2017-06-02 02:58:28)

>    Note that face to face meetings are an important part of the process -
>    mentoring junior people is much harder without them, as is coming to
>    consensus on certain types of problems.

Fair, but I'll also point out that learning to use infrastructure like
this ought to be one of the things we're teaching them.

>    Many of the meetings (e.g. the weekly HIL meeting) have remote access;

Do they? I was unaware; are instructions posted somewhere?

At a minimum, we ought to take minutes & post them if we're going to
have these meetings, so we have a paper trail.

pjd-nu:
I'm not a big fan of coming up with a lot of process before the fact, when it's probably going to be wrong when we actually need it. In particular I don't think we need much of a template - any issues that people have can be hashed out in the actual document that gets submitted to kick off a discussion. If you think anything is missing, just complain in a review comment.

zenhack:
Yeah, I'm not convinced we need this at all. I could be talked into turning `docs/features/newFeatureSpecificationTemplate.md` int a short list of bullet points on a wiki page.

docs/features/indiscussion/readme.md
```diff
@@ -0,0 +1,6 @@
+New feature spec documents that are still under review live here.
```
naved001: I don't understand this. If a feature is under discussion, wouldn't that be an open PR where we discuss it? Once we all agree, we can just move it to the `docs/ongoing/` folder.

zenhack:
I'm in favor of closing this. As I said, I could be convinced to let someone throw together a bullet list and post it somewhere, but this reads as something very different, and I'm not sure it's even worth picking apart and refining, vs *if* someone cares enough just creating a wiki page with some bullets. But I don't think it's sensible to spend time on this right now. Unless someone voices an objection before we finish with #800, I'm going to close this so we can just go ahead with the rename.

knikolla:
While I was the one that suggested the spec process first, I agree that discussing in issues is also a good alternative (and things can be tagged appropriately). 

There are two points.
* Proposing major feature changes as PRs gives you the possibility to comment line by line. 

* As the discussion progresses the document is updated with them. This is in contrast with a discussion on a github issue where you basically need to go through the entire thing and piece it together yourself before you get a clearer picture. When I’m constrained on time I’d much rather read an up-to-date document and only if I have concerns about a point read the comments regarding the points I had a concern with.

I don’t have strong opinions about the first point.

The second point can also be emulated in an issue to by continuously updating the first comment of the issue thread with new information.

zenhack:
Ok, sure, might be nice to make use of the pr mechanism, But I'd also say rather than writing up a formal proposal, just try it sometime, see how it goes. Don't thing we need this much verbiage to do that.

zenhack:
I'm going to close this in favor of #803

Issue #781, opened 2017-05-22
Associate NetworkingActions with Ports, not Nics.


zenhack:
Right now, the schema for NetworkignActions has a foreign key referencing the nic table. As a result, We can't do any port operations on ports that don't have an associated nic. This blocks #761.

It shouldn't be a *huge* change to have a port_id instead of a nic_id; we'd need to tweak a few other bits to match, but I don't think it would be that hard.

//cc @henn, @naved001

naved001:
I made some [quick changes](https://github.com/CCI-MOC/hil/compare/master...naved001:port-based-networkingActions) to support this. I have associated NetworkingActions with Ports, in addition to Nics, that way I didn't have to modify all the other functions. so minimal changes. I can call revert port on a port that's not attached to a nic. 

If this approach seems reasonable, then I can finish this up otherwise I am open to other suggestions.

zenhack:
I don't like having the redundancy, and I worry that allowing the nic to be null is going to introduce bugs.

We can probably spoof the old nic property with the `@property` decorator, and maybe even shove a deprecation warning in the getter/setter. This solves the redundancy issue, but not the null concern.

I'd also like to actually look and try to estimate how much work it would be to just update everything, and what the result would look like.

naved001:
> I'd also like to actually look and try to estimate how much work it would be to just update everything, and what the result would look like.

I tried to estimate how much work would it be but ended updating everything. I'll make a PR soon.

naved001:
Should we close this issue if we are going for a more generic mechanism? See this [comment](https://github.com/CCI-MOC/hil/pull/784#issuecomment-308510830)

zenhack:
Yeah, good call. Closing.

PR #784, opened 2017-05-26
Networking actions are now associated with ports, not nics.

naved001:
Fixes issue #781 

I have updated everything to associate networking actions with ports. Updated the tests too. 

Also ran the network deployment tests on the dell switch and those passed.

cc: @henn @SahilTikale

### DIFF
--- a/haas/deferred.py
+++ b/haas/deferred.py
@@ -29,37 +29,34 @@ class DaemonSession(object):
     def handle_action(self, action):
         if action.type not in model.NetworkingAction.legal_types:
             logger.warn('Illegal action type %r from server; ignoring.')
-        elif not action.nic.port:
-            logger.warn('Not modifying NIC %s; NIC is not on a port.',
-                        action.nic.label)
         else:
             getattr(self, action.type)(action)
 
     def modify_port(self, action):
-        session = self.get_session(action.nic.port.owner)
+        session = self.get_session(action.port.owner)
 
         if action.new_network is None:
             network_id = None
         else:
             network_id = action.new_network.network_id
 
-        session.modify_port(action.nic.port.label,
+        session.modify_port(action.port.label,
                             action.channel,
                             network_id)
         if action.new_network is None:
             model.NetworkAttachment.query \
-                .filter_by(nic=action.nic, channel=action.channel)\
+                .filter_by(nic=action.port.nic, channel=action.channel)\
                 .delete()
         else:
             db.session.add(model.NetworkAttachment(
-                nic=action.nic,
+                nic=action.port.nic,
                 network=action.new_network,
                 channel=action.channel))
 
     def revert_port(self, action):
-        session = self.get_session(action.nic.port.owner)
-        session.revert_port(action.nic.port.label)
-        model.NetworkAttachment.query.filter_by(nic=action.nic).delete()
+        session = self.get_session(action.port.owner)
+        session.revert_port(action.port.label)
+        model.NetworkAttachment.query.filter_by(nic=action.port.nic).delete()
 
     def get_session(self, switch):
         if switch.label not in self.switch_sessions:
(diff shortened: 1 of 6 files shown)


naved001:
@zenhack The migration tests are failing because the old table had nic_id and nic, and the new table would have port_id and port. Kinda confused what would be the right way to fix this?

haas/deferred.py
```diff
@@ -29,37 +29,34 @@ def __init__(self):
...
                 network=action.new_network,
                 channel=action.channel))
 
     def revert_port(self, action):
-        session = self.get_session(action.nic.port.owner)
-        session.revert_port(action.nic.port.label)
-        model.NetworkAttachment.query.filter_by(nic=action.nic).delete()
+        session = self.get_session(action.port.owner)
+        session.revert_port(action.port.label)
+        model.NetworkAttachment.query.filter_by(nic=action.port.nic).delete()
```
zenhack: What happens if `action.port.nic is None`? That wasn't possible before (which was kinda the point of this patch).

I don't know whether this query will do the right thing or just barf at us, but it would be correct to wrap it in a check for the above condition, and perhaps clearer.

haas/model.py
```diff
@@ -531,19 +531,18 @@ class NetworkingAction(db.Model):
     # * 'revert_port' detaches the port from all networks.
     type = db.Column(db.String, nullable=False)
 
-    nic_id = db.Column(db.ForeignKey('nic.id'), nullable=False)
     new_network_id = db.Column(db.ForeignKey('network.id'), nullable=True)
 
+    # The port affected by the action. for 'revert_port'
```
zenhack: `for 'revert_port'` isn't really very useful here -- it's still getting used for all of the other operations. Instead, I might mirror the comment you removed, saying something like "the port affected by the action. For operations that conceputally operate on nics, this is the port they are attached to."

The fact that the reason for the change has to do with revert_port is something that makes more sense in a commit message -- the reader of the code isn't necessarily interested in a specific point in history other that the present.

tests/unit/api/port_revert.py
```diff
@@ -43,13 +43,6 @@ def tearDown(self):
         self.request_context.pop()
         releaseDB()
 
-    def test_no_nic(self):
```
zenhack: Might be nice to put something into place that replaces this, i.e. verify that port_revert *does* work when there's no nic.

zenhack:
Just a note, fyi: the `relationship` fields in the python source (nic/port, as opposed to nic_id/port_id) don't have any impact on database the schema -- they just set up nice helpers that you can use to do things like node.nic.port.label, rather than having to explcitly write queries involving nic_id, port_id, etc.

henn:
FYI - it sounds like this could break my proposal in #788, where a force-disconnected NIC (due to a broken switch) could potentially hold onto its networking actions, then resume them when reconnected elsewhere.

@zenhack - I remember that we had a discussion in the past about making the `NetworkingAction` queue more generic/closure-ish. Maybe now would be the right time? I can't seem to find our previous discussion, otherwise I'd link to it.

naved001:
Before I add more stuff to this PR and generate a migration script. I am wondering if we should continue with this PR as this might break #788. 

I just had some thoughts, correct me if I am wrong. The only time we care about a port is when we connect a nic to it. So why can't we do our port initialization when we call `port_connect_nic`?

zenhack:
Ack, I thought I'd posted this comment a few days ago, maybe I forgot to hit the button: I think #788 is the wrong approach. I suggested elsewhere that we have api calls for cancelling networking actions, esp. useful for errored actions. Instead of having a `--force` and having stuff resume on reconnect, we could just require the administrator to clear the queue.

zenhack:
I dug up the conversation @henn was referring to:

https://github.com/CCI-MOC/hil/issues/508#issuecomment-252346268

This might not be a bad time to think about that, esp. given the console daemon stuff that's going on -- should make sure we don't have too many ad-hoc protocols. Though I think the rest-ish interface of the console daemon as spec'd is fairly coherent.

henn:
We decided to first make a more generic RPC mechanism (see @zenhack's last link) first, and then do this.

PR #785, opened 2017-05-26
sleep() in serve_networks() tunable via cfg

Izhmash:
- In response to issue #318 :
- The sleep() time in serve_networks() can now be set in the configuration file under [devel].
- The default sleep() time is 2 seconds if it is not read from haas.cfg
- In the dev branch there is a solution where serve_networks() is passed the sleep() time as a parameter.  However, this may not be necessary if the desired sleep() time will only be set by the config file.

### DIFF
--- a/examples/haas.cfg
+++ b/examples/haas.cfg
@@ -91,7 +91,7 @@ uri = sqlite:////absolute/path/to/haas.db
 
 [network-daemon]
 # The amount of time in seconds to sleep after attempting to empty the journal 
-# when running serve_networks. Range is 0 < sleep_time < 3600. Default is 2:
+# when running serve_networks. If set, must be > 0 and < 3600. Default is 2:
 #sleep_time=
 
 [extensions]
--- a/examples/haas.cfg
+++ b/examples/haas.cfg
@@ -91,7 +91,9 @@ uri = sqlite:////absolute/path/to/haas.db
 
 [network-daemon]
 # The amount of time in seconds to sleep after attempting to empty the journal 
-# when running serve_networks. If set, must be > 0 and < 3600. Default is 2:
+# when running serve_networks. If set, must be > 0 and < 3600 (1 hour). A
+# warning will be logged if sleep_time is greater than 60 (1 minute).
+# Default value if unset is 2:
 #sleep_time=
 
 [extensions]
--- a/examples/haas.cfg
+++ b/examples/haas.cfg
@@ -88,6 +88,10 @@ uri = sqlite:////absolute/path/to/haas.db
 # ``dry_run`` is present (regardless of its value), this functionality is
 # enabled:
 #dry_run=
+#
+# The amount of time in seconds to sleep after attempting to empty the journal 
+# in the serve_networks command.  Default in haas/cli.py is 2 seconds:
+#sleep_time=
 
 [extensions]
 # List of extensions to load. The values should all be empty. See
--- a/examples/haas.cfg
+++ b/examples/haas.cfg
@@ -91,7 +91,7 @@ uri = sqlite:////absolute/path/to/haas.db
 
 [network-daemon]
 # The amount of time in seconds to sleep after attempting to empty the journal 
-# when running serve_networks.  Default time is 2 seconds:
+# when running serve_networks. Range is 0 < sleep_time < 3600. Default is 2:
 #sleep_time=
 
 [extensions]
--- a/examples/haas.cfg
+++ b/examples/haas.cfg
@@ -88,9 +88,10 @@ uri = sqlite:////absolute/path/to/haas.db
 # ``dry_run`` is present (regardless of its value), this functionality is
 # enabled:
 #dry_run=
-#
+
+[network-daemon]
 # The amount of time in seconds to sleep after attempting to empty the journal 
-# in the serve_networks command.  Default in haas/cli.py is 2 seconds:
+# when running serve_networks.  Default time is 2 seconds:
 #sleep_time=
 
 [extensions]
--- a/haas/cli.py
+++ b/haas/cli.py
@@ -257,12 +257,12 @@ def serve_networks():
     # Check if config contains usable sleep_time
     if cfg.has_option('network-daemon', 'sleep_time'):
         try:
-            sleep_time = float(cfg.get('network-daemon', 'sleep_time'))
-            if sleep_time < 0:
-                raise ValueError
+            sleep_time = cfg.getfloat('network-daemon', 'sleep_time')
         except (ValueError):
-            sys.exit("Error: Invalid sleep_time. " \
-                     "Must be a positive decimal value\n")
+            sys.exit("Error: sleep_time set to non-float value\n")
+        if sleep_time <= 0 or sleep_time >= 3600:
+            sys.exit("Error: sleep_time not within bounds " \
+                     "0 < sleep_time < 3600\n")
     else:
         sleep_time = 2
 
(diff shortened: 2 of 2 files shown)


haas/cli.py
```diff
@@ -247,17 +247,24 @@ def serve_networks():
...
     server.register_drivers()
     server.validate_state()
     model.init_db()
     migrations.check_db_schema()
+    # Check if config contains usable sleep_time
+    try:
+        sleep_time = float(cfg.get('devel', 'sleep_time'))
+    except (ConfigParser.NoSectionError, ConfigParser.NoOptionError,
+            ValueError):
+        sleep_time = 2
```
zenhack: I'd rather not just silently use the default value if the user makes a typo in their config file, e.g.

```
[devel]
sleep_time  = e3 # user fat-fingered, hit two keys at once
```

And, re: handling the absent case, it would be clearer to just do:

```python
if cfg.has_option('devel', 'sleep_time'):
    sleep_time = cfg.get('devel', 'sleep_time')
else:
    ...
```

examples/haas.cfg
```diff
@@ -88,6 +88,10 @@ uri = sqlite:////absolute/path/to/haas.db
 # ``dry_run`` is present (regardless of its value), this functionality is
 # enabled:
 #dry_run=
+#
+# The amount of time in seconds to sleep after attempting to empty the journal 
+# in the serve_networks command.  Default in haas/cli.py is 2 seconds:
```
zenhack: We shouldn't (generally) be referencing source-locations in the config file -- remember this is an end-user visible config file -- sysadmins don't need to be expected to read the source.

haas/cli.py
Izhmash: @zenhack Is there a best practice for handling typos in the config file?  I wasn't sure it it's appropriate to print out an error for the user.

naved001: Just like @zenhack said, if there's a typo or something in the config file. let the user see that there's a ValueError instead of silently changing it because that way the user wouldn't know why they couldn't set the value they wanted.

henn: First, we should be using [getfloat()](https://docs.python.org/2/library/configparser.html#ConfigParser.RawConfigParser.getfloat) and not float().

Second, NoOptionError/NoSectionError are separate issues from ValueError. For the first two, the user didn't specify anything, so we should set a default for them.

For ValueError, two options would be to either throw an ERROR and exit in the exception block, or just remove the try/except and throw the exception directly back to the user. I like the first one better, since easy to understand error messages like `sleep_time set to non-integer value` is easier for a sysadmin to understand.

Finally, we should be doing a bound check (like 0 < sleep_time < 3600) on configured sleep values, printing an appropriate error message.

Thanks!

Izhmash: I didn't see these before my latest commit; I'll fix and recommit.  Thanks for the input.

Izhmash: The current commit should now address the latest requests.

haas/cli.py
```diff
@@ -247,17 +247,31 @@ def serve_networks():
     """Start the HaaS networking server"""
     from haas import model, deferred
     from time import sleep
+    import ConfigParser
```
henn: Don't think you need this anymore

haas/cli.py
```diff
@@ -247,17 +247,31 @@ def serve_networks():
...
     server.init()
     server.register_drivers()
     server.validate_state()
     model.init_db()
     migrations.check_db_schema()
+
+    # Check if config contains usable sleep_time
+    if cfg.has_option('network-daemon', 'sleep_time'):
```
henn: Did you check to see if you need a cfg.has_section() first, in case the network-daemon section is missing?

Izhmash: I added a statement checking for the network-daemon section in my last commit.  If either the section or the value is missing I go straight to using the default.

haas/cli.py
naved001: @henn since we are setting a bound, I was wondering if 3600 seconds (an hour) would be too long.

zenhack: I'm actually inclined to just not have an upper bound. The reason is just that it's extra stuff that needs to be documented, and therefore stuff a sysadmin has to sift through, and it's not checking any particularly important invariants -- it's just an arbitrary cap. I'd just check that it's positive and leave it at that.

henn: My thought on this was that we wanted some sort of reasonable upper bound so that wrong things that could look like numbers (like a hex number) would get caught.

My suggestion is that in the example config file we just say "if set, must be > 0 and < 3600" to keep the documentation tidy. What do you guys think?

henn: > if 3600 seconds (an hour) would be too long

My thought on allowing long periods was for debugging, like somebody manually trying to reproduce a problem and insert something into the queue, which could take a lot of time.

haas/cli.py
```diff
@@ -252,12 +252,26 @@ def serve_networks():
...
+    # Check if config contains usable sleep_time
+    if (cfg.has_section('network-daemon') and
+            cfg.has_option('network-daemon', 'sleep_time')):
+        try:
+            sleep_time = cfg.getfloat('network-daemon', 'sleep_time')
+        except (ValueError):
+            sys.exit("Error: sleep_time set to non-float value\n")
+        if sleep_time <= 0 or sleep_time >= 3600:
+            sys.exit("Error: sleep_time not within bounds "
+                     "0 < sleep_time < 3600\n")
```
Izhmash: Ok, I'll fix that.

haas/cli.py
Izhmash: Do we think there could be any case when someone might want more than an hour?  If definitely not then I think the upper bound would be more helpful than hurtful.

haas/cli.py
Izhmash: Should be all set now

haas/cli.py
zenhack: What if we logged a warning if the value was over a certain (much more conservative) value? There's a pretty wide range of unusably long delays between the default (2 seconds) and an hour -- this would give us the ability to be more aggressive with the warning, while not preventing people from debugging. Otherwise, we have to put the upper bound well outside the range of "this might be usable"

I don't have terrbily strong feelings on the matter. If we do put in an upper bound like 3600, I'd like the docs to have a parenthetical `(1 hour)`, just so the reader can see it at a glance. But other than that, I'll go along with whatever you guys decide.

Izhmash: So then in that case I could log something like `Warning: sleep_time greater than 10 minutes` if someone set `sleep_time` to, say, greater than 600 seconds, but the program won't exit.  What do you think @henn ?

henn: I like @zenhack's idea of a warning, though maybe warn if it's > 1 minute, though still keeping some upper bound to prevent garbled input from being interpreted.

zenhack: I think I like 1 minute as the warning bound; that's enough that it really isn't a reasonable value, but with anything less the system won't be *completely* unusable until the sysadmin fixes it.

Izhmash: Ok, I'll add the warning and update the documentation accordingly.

Izhmash: Everything should be all set now.

henn:
@zenhack - Are you happy with the PR now?

Izhmash:
Ok, I updated it so now the logging module is called when sleep_time > 1 min.

Izhmash:
The import and initialization for the logger are now at the top level.

zenhack:
LGTM, merging.

Issue #315, opened 2014-10-27
Integer may be too small to ensure network action ordering.


zenhack:
@henn brought this up. We're using the primary key for our network actions to determine ordering, and on most databases `Integer`is a 32-bit signed int. It's possible that a long-running HaaS installation might overflow eventually; my back-of-the-napkin estimates put this at ~3.5 years if you accumulate ten actions per second.

Using a 64 bit number would be more than sufficient to prevent this.

EDITED BY HENN (6/7/2017) to add checklist we discussed:

- [x] Do first PR with one migration.
- [x] Do the rest

Izhmash:
I've been working on finishing this one up.  Currently reviewing my changes and writing some tests.

Izhmash:
Since I'm changing the primary keys in core `haas`, `haas.ext.switches.mock`, and `haas.ext.obm.mock`, do I need to have three separate migrations? Or do I only need one for all three?

zenhack:
You need separate ones -- the ones for extensions should only get run if those extensions are loaded.

Izhmash:
@henn Would you like me to add the overflow test we discussed before I start the pull request?

henn:
Checklist we discussed:

- [ ] Do first PR with one migration.
- [ ] Do the rest

SahilTikale:
Blocked on #805

Izhmash:
Addressed by #839

PR #789, opened 2017-06-02
DB Migration Integer to BigInteger

Izhmash:
* In response to issue #315:
* I've updated the database creation source so that every primary key is a BIGINT rather than an INTEGER.
* 3 migration scripts: one for haas, one for haas.ext.obm.mock, and one for haas.ext.switches.mock.
* Type changes in the DB are now noticed by Alembic through env.py.
* Note: overflow test implementation is in progress


zenhack:
*sigh* it looks like the client tests are acting up again

It's unfortuante that we don't actually get the errors back when travis just times out...

//cc @henn, @SahilTikale

naved001:
The tests failures here are a bit weird as almost all tests are failing this time. Running it locally, the tests are erroring (E) instead of failing (F) saying `socket.error: [Errno 98] Address already in use
`

I was also wondering why doesn't alembic just generate pep8 formatted commands/scripts

Izhmash:
Since more development will be necessary for this, I'm going to work on getting one table migrated without error first, then make another PR when I'm certain that the rest of the tables can be updated without error.

Izhmash:
After some research on various forums, it's looking like sqlite does not support BIGINT primary keys.  However, the default integer in sqlite can be as large as 8 bytes.  Postgres, on the other hand, does accept a BIGINT primary key.

Someone has an example [here](https://stackoverflow.com/questions/18835740/does-bigint-auto-increment-work-for-sqlalchemy-with-sqlite) about having the code behave differently depending on the database used.  Should I go down that route, or is the extra complexity to the code not worth it?

Izhmash:
I'd argue for the postgres BIGINT migration and having sqlite make up for it some other way since sqlite is not used in production.

zenhack:
We can probably lean on sqlalchemy to paper over the differences; might have to dig through the docs a bit, but I'm willing to bet there is a way to get it to use different types for different dbs.

Izhmash:
Just for an update, I've been working on this on a private branch. I've got a solution that works when upgrading just the NIC table.  Going to push it here as soon as the migration.py tests stop failing.

Izhmash:
I'm going to close this and open a new pull request with my other branch.

PR #791, opened 2017-06-06
Raise pytest version lower bound

Izhmash:
- The client unit test was failing with `pytest.fixture functions cannot use ``yield``. Instead write and return an inner function/generator and let the consumer call and iterate over it.`
- ~After changing from `fixture` to `yield_fixture` the tests ran normally.~

Edit: This is what we are doing now https://github.com/CCI-MOC/hil/pull/791#issuecomment-306569055


zenhack:
I willing to bet this is because you've got an old venv that already had pytest installed, hence was running 2.x. This has been passing on travis, so it works with the latest 3.x.

@henn specifically changed this, and raise tge upper bound on pytest to <4.0. Maybe we shpuld raise the lower bound, instead of reverting his change?

Izhmash:
@zenhack that ended up being my issue, I was running an older version of pytest.

From the [revision history](https://docs.pytest.org/en/latest/changelog.htmll) it looks like `yield_fixture` is no longer necessary.  Should I set the lower bound to version 3.0.0 and change `yield_fixture` back to `fixture`?

zenhack:
Quoting Ian Ballou (2017-06-06 14:04:06)

>    From the revision history it looks like yield_fixture is no longer
>    necessary. Should I set the lower bound to version 3.0.0 and change
>    yield_fixture back to fixture?

Sounds good to me.

Izhmash:
Ok, great. Everything should be all set now.

zenhack:
Pending travis passing (can't imagine it won't...), LGTM. I think we should squash this one.

Issue #713, opened 2017-01-30
API to change a user's admin status


henn:
Currently, it's only possible to either create a user as an admin user. It should also be possible to promote a user.

Izhmash:
I'm going to start working on this one.

naved001:
#792 fixed this!

PR #792, opened 2017-06-07
API: Changing HaaS user privilege

Izhmash:
* In response to issue #713:
* Admin users can now set another user's admin status, either Regular -> Admin or Admin -> Regular.
* Functionality test added to tests/unit/client.py.
* Documentation updated.


tests/unit/client.py
```diff
@@ -637,6 +637,39 @@ def test_user_remove_error(self):
...
+        """Tests error condition while editing a user who doesn't exist. """
+        C.user.create('gary', '12345', 'admin')
+        C.user.delete('gary')
+        with pytest.raises(FailedAPICallException):
+            C.user.set_privilege('gary', 'regular')
+
+    def test_user_set_privilege_promote_error(self):
+        """Tests error condition while editing a user who doesn't exist. """
+        C.user.create('garrey', '12345', 'regular')
+        C.user.delete('garrey')
```
zenhack: spelling.

naved001:
You would need to add tests for the API here `tests/unit/api/main.py` and `/tests/unit/api/auth.py` 
https://github.com/CCI-MOC/hil/tree/master/tests/unit/api

haas/ext/auth/database.py
```diff
@@ -131,6 +131,21 @@ def user_remove_project(user, project):
...
+    'user': basestring,
+    'privilege': bool,
+}))
+def user_set_privilege(user, privilege):
+    get_auth_backend().require_admin()
+    user = api._must_find(User, user)
+    if privilege == user.is_admin:
+        status = 'admin' if privilege else 'regular'
+        raise DuplicateError('No change. User %s privilege status: %s' %
+                             (user.label, status))
```
zenhack: I would just get rid of this if statement entirely; the api call gets much more readable without it (though the readability could probably be improved otherwise).

docs/rest_api.md
```diff
@@ -1138,6 +1138,28 @@ Authorization requirements:
 
 * Administrative access.
 
+#### user_set_privilege
```
zenhack: I'd prefer it if this mirrored the create call a bit better. Suggestion:

* rename it to `user_set_admin`
* change the URL to `/auth/basic/user/<user>/is_admin`
* change the body to accept a boolean instead of a string.
* make the corresponding client-side changes.

An alternative might be to use `PATCH` on `/atuh/basic/user/<user>`.

haas/ext/auth/database.py
```diff
@@ -131,6 +131,21 @@ def user_remove_project(user, project):
     db.session.commit()
 
 
+@rest_call('PUT', '/auth/basic/user/<user>/set_privilege', Schema({
+    'user': basestring,
+    'privilege': bool,
```
zenhack: This is inconsistent with the docs, which say this should be either the string `admin` or `regular`. As outlined above I actually think it should stay a boolean and you should update the docs.

Izhmash:
@zenhack Since there aren't any existing user tests (creating users, deleting users, etc.) in `tests/unit/api/auth.py`, will I need to add those before adding tests for `user_set_admin()`?

I initially assumed my tests for `user_set_admin()` would only go where there are tests for other user-related commands.

tests/unit/client.py
```diff
@@ -637,36 +637,24 @@ def test_user_remove_error(self):
         with pytest.raises(FailedAPICallException):
             C.user.remove('sam03', 'test_proj03')
 
-    def test_user_set_privilege(self):
+    def test_user_set_admin(self):
         """Test changing a user's admin status """
         C.user.create('jimmy', '12345', 'regular')
         C.user.create('jimbo', '678910', 'admin')
         assert C.user.set_privilege('jimmy', 'admin') is None
```
naved001: this should be `C.user.set_admin`
that should fix the client tests error you are having.

zenhack:
Ah, right -- because those calls are defined in an extension, the auth related tests will be elsewhere. In this case tests/unit/ext/auth/database.py.

tests/unit/client.py
Izhmash: Thanks, I totally missed that.

docs/apidesc.md
```diff
@@ -116,6 +116,7 @@ one exception is NICs, where the label is unique only on a per-node basis.
 
     user_create                 <user_label> <password>
     user_delete                 <user_label>
+    user_set_admin              <user_lable> <is_admin>
```
knikolla: nit: lable

docs/rest_api.md
```diff
@@ -1138,6 +1138,26 @@ Authorization requirements:
...
+
+Request Body:
+
+{
+    'is_admin': <boolean>
+}
+
+Set admin status of user '<user>' to true (admin user) or false (regular user)
+
+Authroization requirements:
```
knikolla: nit: Authorization

haas/client/user.py
```diff
@@ -50,3 +50,22 @@ def remove(self, user, project):
         return self.check_response(
                 self.httpClient.request("POST", url, data=payload)
                 )
+
+    def set_admin(self, username, privilege):
```
knikolla: I do not like that we're introducing magic strings here. This should accept a True or False for is_admin and apply that.

haas/ext/auth/database.py
```diff
@@ -131,6 +131,17 @@ def user_remove_project(user, project):
     db.session.commit()
 
 
+@rest_call('PUT', '/auth/basic/user/<user>/set_admin', Schema({
+    'user': basestring,
+    'is_admin': bool,
+}))
+def user_set_admin(user, is_admin):
+    get_auth_backend().require_admin()
```
knikolla: Would like to see some mechanism to prevent someone from removing themselves from admin.

docs/rest_api.md
knikolla: I would have much more preferred the PATCH alternative.

docs/apidesc.md
naved001: this file is now gone #793.

haas/client/user.py
Izhmash: @knikolla I kept the privilege string to mirror the `create()` command (which also accepts a string).  The documentation for `create()` mentions that a user is either 'admin' or 'regular', which maps to true or false. 

I'm not sure if we'd rather go with consistency or with what might make more sense.

Izhmash: I'll update the string to be a boolean, should I also put in an issue for create() using the string still?

zenhack: @lzhmash, yeah, why don't you open one.

docs/rest_api.md
```diff
@@ -1198,6 +1198,26 @@ Authorization requirements:
 
 * Administrative access.
 
+#### user_set_admin
+
+'PUT /auth/basic/user/<user>/set_admin'
```
zenhack: Maybe change the path component to `is_admin`; the general notion with REST Apis is that URLs represent objects, not functions. PUT /my-url` is conceptually the same as an assignment statement, so the URL should denote the "variable" being set.

I'm actually with @knikolla re: changing it to a `PATCH` on the user URL, though I don't feel that strongly. but the request body would make more sense that way.

haas/client/user.py
Izhmash: @zenhack In passing a boolean rather than a string to the client library, the check `if privilege not in('admin', 'regular'): ...` also must change.  Would it be appropriate to move the check for correct spelling to the CLI before I call `set_admin()`?

zenhack: yeah, that should get moved to the CLI.

haas/client/user.py
```diff
@@ -51,21 +51,16 @@ def remove(self, user, project):
                 self.httpClient.request("POST", url, data=payload)
                 )
 
-    def set_admin(self, username, privilege):
-        """Changes the admin status of <user>.
+    def set_admin(self, username, is_admin):
+        """Changes the admin status of <username>.
 
-        <privilege> may by either "admin" or "regular",
+        <is_admin> may be either True or False,
```
zenhack: Could just say it's a boolean.

docs/rest_api.md
```diff
@@ -1200,7 +1200,7 @@ Authorization requirements:
 
 #### user_set_admin
 
-'PUT /auth/basic/user/<user>/set_admin'
+'PATCH /auth/basic/user/<user>/is_admin'
```
zenhack: With the PATCH verb it makes more sense to just do `/auth/basic/user/<user>`; the thing you're patching is the user.

Izhmash:
All of the requests should be addressed now.

zenhack:
LGTM

haas/ext/auth/database.py
```diff
@@ -131,6 +132,19 @@ def user_remove_project(user, project):
...
 
 
+@rest_call('PATCH', '/auth/basic/user/<user>', Schema({
+    'user': basestring,
+    'is_admin': bool,
+}))
+def user_set_admin(user, is_admin):
+    get_auth_backend().require_admin()
+    user = api._must_find(User, user)
+    if user.label == os.environ.get('HAAS_USERNAME'):
```
naved001: I ran into an issue here because I didn't set the env variable where I was running the api server (`haas serve 5000`).

A user would set *his* environment variables. When you do an `os.environ.get` here it would return HAAS_USERNAME of the env where the server is running. So this mechanism would fail.

anybody know what's the right way to get the user here? from the URL?

zenhack: Ack, was this not there when I reviewed or did I just miss it?

One approach might be to allow this at the Api level, but do the sanity check in the cli. There might be a way to fish the current user out of the request, but the design of the server is going to fight you on this sort of thing -- the details of authentication are very purposefully hidden from Api calls. Since this is extension specific, this isn't an issue but: in general there may not be anything you could call a user at all.

Izhmash:
My latest commit adds an easier way for the API to check for user self-promotion and self-demotion.

naved001:
> My latest commit adds an easier way for the API to check for user self-promotion and self-demotion.

Great. This works!

Also, please remove [this file](https://github.com/CCI-MOC/hil/pull/792/files#diff-21232f297a57a5a743894a0e4a801fc3)  I think you committed this by mistake?

haas/cli.py
```diff
@@ -292,6 +292,22 @@ def user_create(username, password, is_admin):
...
+    """Changes the admin status of user <username>.
+
+    <is_admin> may by either "admin" or "regular", and determines whether
+    a user is authorized for administrative privileges.
+    """
+    if is_admin not in ('admin', 'regular'):
+        raise ValueError(
+            "invalid user privilege: must be either 'admin' or 'regular'."
+            )
+    check_clientlib_response(
```
zenhack: This needs to be changed now that #796 has been merged.

haas/ext/auth/database.py
```diff
@@ -11,6 +11,7 @@
 from schema import Schema, Optional
 import flask
 import logging
+import os
```
zenhack: Now that the call to getenv is gone, you no longer need this.

zenhack:
Could we get a test for the check you just added? Other than that and the two comments I made in-line I'm hapy.

Izhmash:
I added a couple of tests and cleaned up.

zenhack:
Ok, I'm happy. @knikolla, how does it look to you? merge button is still blocked on your review.

tests/unit/ext/auth/database.py
```diff
@@ -192,6 +192,71 @@ def test_new_non_admin_cannot_admin(self):
...
+        self.dbauth.user_set_admin('charlie', False)
+        self.dbauth.user_delete('charlie')
+
+    def test_mod_admin_can_admin(self):
+        """Verify that a newly promoted admin can actually do admin stuff."""
+        self.dbauth.user_create('charlie', 'foo', False)
+        self.dbauth.user_set_admin('charlie', True)
+        flask.request = FakeAuthRequest('charlie', 'foo')
+        local.auth = self.dbauth.User.query.filter_by(label='charlie').one()
+        print local.auth
```
knikolla: Remove these prints.

tests/unit/ext/auth/database.py
```diff
@@ -192,6 +192,71 @@ def test_new_non_admin_cannot_admin(self):
...
+        self.dbauth.user_delete('charlie')
+
+    def test_mod_non_admin_cannot_admin(self):
+        """Verify that a newly demoted regular user can't do admin stuff."""
+        self.dbauth.user_create('charlie', 'foo', True)
+        self.dbauth.user_set_admin('charlie', False)
+        flask.request = FakeAuthRequest('charlie', 'foo')
+        local.auth = self.dbauth.User.query.filter_by(label='charlie').one()
+        print local.auth
```
knikolla: Same.

Izhmash:
Cleaned out those print statements.

Issue #543, opened 2016-03-07
API documentation should be in rest_api.md


kylehogan:
API documentation currently duplicated across api.py and rest_api.md. All documentation for the API should be in rest_api.md

Need to move documentation for API calls from api.py to rest_api.md

xuhang57:
@kylehogan  Not sure if this is done. By documentation for API calls from api.py, do you mean the docs string under the name of each metho in api.py?

kylehogan:
Yes, if I am remembering correctly I made this while we were discussing it during group meeting and the consensus was that there should be only short descriptions in api.py while the bulk of the information should be in rest_api.md

I think what was happening was that information was getting duplicated across both locations and that the length of the descriptions within api.py was becoming too much.

knikolla:
Fixed by #793

PR #793, opened 2017-06-07
Remove redundant information from apidesc.md

zenhack:
And update a few bits that were out of date. This has been bugging me
for a while -- rest_api.md is the real API resource, but this has
existed and every time someone submits a patch that adds an API call,
*if* they remember to update apidesc (which is not always, and I never
do...) I always wonder why we have the file at all. Then I remember that
there's some overview stuff at the top.


naved001:
Why don't we put the overview stuff at the top of `rest_api.md`? It's just 60 lines. We would only need the 2 sections `Objects in the HaaS` and `API design philosophy`.

zenhack:
@naved001, good call, done.

naved001:
now that we'll have all the API stuff in one place, this should take care of #543

Issue #768, opened 2017-04-29
Get rid of the lambdas sprinkled throughout CLI


henn:
Since we're re-doing the CLI anyhow in #746 , we should get rid of the extra lambdas that are all over [cli.py](https://github.com/CCI-MOC/hil/blob/master/haas/cli.py). They're extra syntax that complicates the code, and we can get by without them.

My suggestion: stick the call dispatch from main() into a try/except block that catches `FailedAPICallException`, then get rid of check_client_response() and the lambdas.

zenhack:
I like the notion of centralizing the error handling.

Izhmash:
In `check_clientlib_response()` it says `except Exception as e:`.

@henn rather than using `FailedAPICallException`, should I just use `Exception`?  It would still centralize the error handling to `main()`.  

I found for instance when I typed gibberish after `haas user_delete` with `FailedAPICallException` implemented I got a long traceback with a `ValueError`.  With `Exception` instead, the report was simply `Error: No JSON object could be decoded`.

Izhmash:
Extending the last comment, [this comparison](https://github.com/cci-moc/hil/compare/master...Izhmash:replace_lambdas) illustrates my idea to centralize the error checking and to remove the lambdas.  Is it appropriate to fully replace the two original `except` blocks in main()?

zenhack:
I'd actually like to keep the type-specific except blocks -- if we get an error we're not expecting I want it to stand out.

henn:
> I'd actually like to keep the type-specific except blocks --
      if we get an error we're not expecting I want it to stand out.
    I agree. In general, it's better to catch the least amount
      possible.
    Check out https://wiki.python.org/moin/HandlingExceptions

naved001:
Fixed by #796

PR #796, opened 2017-06-09
Replace lambdas

Izhmash:
* In response to issue #768 
* Error-checking lambda functions in haas/cli.py removed.
* `check_clientlib_response()` functionality & error checking implemented in main

### DIFF
--- a/haas/cli.py
+++ b/haas/cli.py
@@ -815,9 +815,9 @@ def main():
         setup_http_client()
         try:
             command_dict[sys.argv[1]](*sys.argv[2:])
+        except FailedAPICallException
+            sys.exit(1)
+        except InvalidAPIArgumentsException:
+            sys.exit(2)
         except Exception as e:
             sys.exit('Error: %s\n' % e.message)
-        #except FailedAPICallException
-        #    sys.exit(1)
-        #except InvalidAPIArgumentsException:
-        #    sys.exit(2)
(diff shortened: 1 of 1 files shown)


zenhack:
Can you explain to me why you re-added the definition of `check_clientlib_response`?

Also, on #768, we talked about descriminating based on error type. I realize now it's not doing that in the first place, but I'd like to distinguish between `FailedAPIException` and other, unexpected errors -- mind adding another `except` case (and maybe change the `Exception` one to print "unexpected error" or something)?

Izhmash:
@zenhack check_clientlib_respsonse is now removed again, re-adding it was a mistake.

For the exceptions, right now there is one for `FailedAPICallException`, `InvalidAPIArgumentsException`, and now unexpected errors.  Was there a fourth `except` that I should add?

zenhack:
No, I don't think there were other cases. One other thing that I noticed though: right now we're just doing a sys.exit(number) for the two specific cases, without printing any specific error info -- it would be better to still print the basic information from the error in these cases as well, as we currently do on master.

Izhmash:
Now it should print error information rather than just the numbers.

zenhack:
Should probably delete the mis-typed call that you're fixing in #799, as well. Other than that this looks good to me.

Izhmash:
I think everything is all set now.  Also closed #799.

zenhack:
@naved001, would you give this a second review?

PR #797, opened 2017-06-09
Use pylint to catch unused imports.

zenhack:
We can add more pylint checks as we go, but this is something that was pretty mechanical to fix, and was all over the place, so it seems like a good place to start. If you enable everything there are a *ton* of errors, so we'll have to take it one step at a time.


haas.wsgi
```diff
@@ -1,7 +1,14 @@
...
 migrations.check_db_schema()
 from haas.rest import app as application  # noqa
+
+# `api` is imported for the side-effect of registering the api functions.
+# silence a pylint error about an unused import:
+api
+
+# ...and the app is imported because this is a wsgi script, and its job
+# is to define `application`. Again, silence the error:
+application
```
naved001: I still don't understand what writing "api" and "application" does here? same thing in the other place.

zenhack: So, if the variables are unused, pylint will issue a warning. Each of these is basically a no-op statement that "uses" the variable, to keep pylint quiet.

Does that clarify? If so I'll add it to the comments.

zenhack:
Re: import *, no, it won't bother us about that, but you can ask it to. That's probably a good next step, but fixing those isn't *quite* as brainless, so I stopped here for now.

haas.wsgi
naved001: Yes, that clears my doubts.

jeremyfreudberg: Why not use `# pylint: disable=unused-import` or `# pylint: disable=W0611` on each offending line to silence the errors? Makes everything a little more self-documenting, and a little less ugly (in my opinion)

zenhack: Good call, that's probably better

zenhack:
@jeremyfreudberg, done; thanks.

naved001:
This one is failing because #772 got merged and it had an unused import (my bad)
```
"************* Module haas.ext.switches.dell

W: 21, 0: Unused import re (unused-import)"
```

zenhack:
@naved001, fixed. Once I'd recommend asking other prs to merge master back in to be sure.

zenhack:
@knikolla, would you do a second review?

PR #799, opened 2017-06-12
node_delete_nic() corrected

Izhmash:
- `check_client_lib()` is an undefined function.
- Prevents `node_delete_nic()` from running correctly.


zenhack:
Hm, given that your other pr just removes this function, do we want to not bother with this and just lean on that for a solution?

zenhack:
(fwiw this looks fine, just seems like since #796 is close to ready we may as well just do that. Probably should have occurred to me on the other issue).

Izhmash:
Yeah, that makes more sense to leave it until it's gone.  I'll close this one.

PR #800, opened 2017-06-14
INTEGER to BIGINT Primary Key DB Migration

Izhmash:
- Addressing issue #315 
- Initially only updating the NIC table
- Sqlite will use its default 64-bit integer
- Postgresql will use the BIGINT datatype


haas/migrations/env.py
```diff
@@ -78,7 +78,8 @@ def process_revision_directives(context, revision, directives):
                                 poolclass=pool.NullPool)
 
     connection = engine.connect()
-    context.configure(connection=connection,
+    context.configure(compare_type=True,
+                      connection=connection,
```
naved001: can you briefly explain what are we doing here?

haas/migrations/versions/c45f6a96dbe7_nic_primary_key_changed_to_bigint.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = 'c45f6a96dbe7'
+down_revision = '3b2dab2e0d7d'
+branch_labels = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
naved001: I think you can remove these comments that say "Autogenerated blah blah" because it's not really helpful.

haas/model.py
```diff
@@ -40,17 +42,31 @@
 
 db = SQLAlchemy(app)
 
+BigIntegerType = BigInteger()
+BigIntegerType = BigIntegerType.with_variant(
+        sqlite.INTEGER(), 'sqlite')
+
```
naved001: Why can't we just do?

`BigIntegerType = BigInteger().with_variant(sqlite.INTEGER(), 'sqlite')`

Same thing in other places too

haas/model.py
```diff
@@ -40,17 +42,31 @@
...
     """
     if uri is None:
         uri = cfg.get('database', 'uri')
     app.config.update(SQLALCHEMY_DATABASE_URI=uri)
 
+    global BigIntegerType
+
+    BigIntegerType = BigInteger()
+    BigIntegerType = BigIntegerType.with_variant(
+            sqlite.INTEGER(), 'sqlite')
```
naved001: Why this again? We already set that globally. Am I missing out on something?

haas/migrations/env.py
Izhmash: The `compare_type=True` allows alembic to detect changes in datatype.

haas/model.py
Izhmash: Sure, I'll give that a go.

haas/model.py
Izhmash: I was having some strange scoping issues but I'll clean that up.

Izhmash:
@naved001 I made some changes cleaning up the code.  Depending on how I call `with_variant()` I get different errors, but the way I'm checking for sqlite vs postgresql after the latest commit looks better and works.

I'm working on the test as well.

haas/migrations/versions/c45f6a96dbe7_nic_primary_key_changed_to_bigint.py
```diff
@@ -0,0 +1,54 @@
...
+        existing_server_default=sa.text(u"nextval('nic_id_seq'::regclass)"))
+    op.alter_column('networking_action', 'nic_id',
+                    existing_type=sa.INTEGER(),
+                    type_=sa.BIGINT(),
+                    existing_nullable=False)
+    op.alter_column('network_attachment', 'nic_id',
+                    existing_type=sa.INTEGER(),
+                    type_=sa.BIGINT(),
+                    existing_nullable=False)
+    # ### end Alembic commands ###
```
zenhack: These comments will need to be removed.

tests/unit/migrations/bigint_nic.sql
```diff
@@ -0,0 +1,1021 @@
+-- Dump taken just after migrating nic to BIGINT primary keyA
```
zenhack: 1. Just putting a file here doesn't do anything -- if you want it to be used in a test, you have to add a test that uses it (though can just be an entry in the list in migrations.py).
2. It's not clear what this is useful for testing? It looks like it's just an empty database.

Izhmash: I'll remove this for now.

Izhmash:
@zenhack I am catching up on the concept of foreign keys in the DB, are foreign keys not updated along with what they reference?  I noticed that there were foreign keys in the NIC table that I upgraded as well.  

As for extensions, I was planning on having a new migration script for each extension to keep them consistent.

zenhack:
Okay, yeah, it does seem to update things where needed. Cool.

zenhack:
(Pending Travis, I'm happy)

haas/model.py
```diff
@@ -40,17 +42,30 @@
...
     config file will be used.
     """
     if uri is None:
         uri = cfg.get('database', 'uri')
     app.config.update(SQLALCHEMY_DATABASE_URI=uri)
 
+    global BigIntegerType
+
+    # Decide if primary keys should be BIGINTs or Integers
+    if 'postgresql' in uri:
```
henn: Why do we need this here, if `BigIntegerType` is properly set above with the sqlite exception?

Izhmash: I have this here so that `BigIntegerType` will instead be a BIGINT if the user has postgresql as a DB back-end. I needed to add this after I learned that sqlite does not willingly accept BIGINTs as primary keys. The sqlite Integer can be 8-bytes max, so this works out.

henn: Why isn't that covered by [line 47](https://github.com/CCI-MOC/hil/pull/800/files/2a959a463171ad2c01a67e6e28823d79ef912cc3#diff-2cdd52227d08d98fe868a52ed69bd91eR47) above?

henn: Given that bigint [should be the same as an int](https://www.sqlite.org/datatype3.html) (since sqlite dynamically sizes the field based on the value), I'm not sure it would hurt anything to use bigint with sqlite and get rid of BigIntegerType completely.

Also, I just tried setting an sqlite PRIMARY key to a bigint, and it seems to work!
```sql
sqlite> create table foo1 ( foo1 int PRIMARY KEY );
sqlite> create table foo2 ( foo1 bigint PRIMARY KEY );
sqlite> 
```

Izhmash: I tried using BIGINT with sqlite earlier, and it seemed to work at first but it consistently failed during testing. I've read that in sqlite you can have a primary key be something other than integer, but for the best case with autoincrement, integer is used. It has to do with the ROWID alias: https://sqlite.org/autoinc.html

henn: Excellent point!

I'm still wondering why the definition of `BigIntegerType` on line 47, with its *with_variant()* exception for sqlite isn't sufficient? Why do we need to create a *with_variant()* version for postgres?

Izhmash: When I used only the sqlite exception, the postgresql DB had integers like before.  After adding the postgresql exception, then I saw a difference in datatypes between the two DBs.

henn: That seems strange.

Maybe try debugging it to see what's happening?

Perhaps there could be a bug in our version of SQLAlchemy? You could try bumping the package version for flask-sqlalchemy to <3.0, since there's a 2.2 out now (though be sure to look for problems, like extra messages in the unit tests).

Izhmash: Ok sure, I'll give that a try 👍

Izhmash: I must've fixed something between now and when I last tested that because it works now without the postgres exception.  Should be all good now!

PR #803, opened 2017-06-15
Create a specs directory/template

henn:
I tried to incorporate some of the feedback from #783 and our meeting.

Issue #764, opened 2017-04-26
Improve client library test performance


henn:
Per discussion with @SahilTikale @naved001 @kylehogan , we'd like to refactor tests to be able to operate in parallel.

Right now this includes:
- [ ] [Client library tests](https://github.com/CCI-MOC/hil/blob/master/tests/unit/client.py) can spawn a different web server for each test, each on its own port. Picking a port could be done either sequentially or randomly, first checking to see if the port is already bound to.
- [ ] Maybe keystone (I've heard @zenhack has ideas on this)

zenhack:
I started some work on this on a branch, but it's blocked by the outright failures we're having. The basic notion I had was to split up our tests into a larger matrix, so e.g. the keystone and apache tests run separately from tests/unit, and thus all of that installation doesn't need to happen in unit, and fewer tests need to be run in those cases.

henn:
Tests running with postgres are currently unable to run in parallel due to sharing a single DB.

To overcome this (allowing us to run unit tests in parallel with `py.test -n auto --boxed`), each worker process could have its own DB named after its worker name. [Pytest-xdist's docs](https://github.com/pytest-dev/pytest-xdist#identifying-the-worker-process-during-a-test) describe how to do this.

zenhack:
I'm going to submit a PR doing the keystone stuff (and some other componentization); I started working on it while we were fussing with the transient failures, but was blocked on the fix.

zenhack:
So, at this point tests/unit/client.py even on sqlite3 is taking longer than the rest of the tests put together -- including the keystone test and apache, and their setup steps. I'm going to do the keystone breakout anyway, since I think it will pay off once we fix that, but this is just dumb -- there's something deeper going on than "we should start parallelizing," those tests should *not* be that slow.

zenhack:
The time is being eaten up by populate_server() -- it takes about 40 seconds to run on my laptop.

zenhack:
...which amounts to about  0.67 requests per second. I'm also getting a whole bunch of this:

```
connectionpool.py          395 DEBUG    "PUT /switch/mock-01/port/gi1/0/3 HTTP/1.1" 200 0
connectionpool.py          248 INFO     Resetting dropped connection: 127.0.0.1
connectionpool.py          395 DEBUG    "POST /switch/mock-01/port/gi1/0/3/connect_nic HTTP/1.1" 200 0
connectionpool.py          248 INFO     Resetting dropped connection: 127.0.0.1
[...]

PR #805, opened 2017-06-16
Rename to hil

zenhack:


zenhack:
Travis's postgres run hung. I've restarted it -- maybe it's just an issue on their end. I feel like this has happened once or twice before though.

README.rst
```diff
@@ -117,7 +117,7 @@ Bibtex:
   keywords = {IaaS, PaaS, bare metal, cloud computing, datacenter management, exokernel},
   }
 
-An early short paper on HIL (then called "Hardware as a Service/HaaS"):
+An early short paper on HIL (then called "Hardware as a Service/HIL"):
```
jeremyfreudberg: incorrect replacement

docs/developer-guidelines.md
```diff
@@ -1,35 +1,35 @@
...
+* Taking a github issue marked with the [easy label](https://github.com/CCI-MOC/hil/labels/easy). (Talk to one of the core team members to have it assigned to you officially!)
+* Taking on a github issue that is part of one of the next [milestones in github](https://github.com/CCI-MOC/hil/milestones?state=open)
 
 ## Communicating
 
 * IRC: The MOC team hangs out on #moc on [freenode](https://www.freenode.net/)
-* IRL (In Real Life): The MOC group office is on the BU campus, at 3 Cummington Mall, Boston, MA room 451. Anyone interested in HaaS is welcome to drop in and work there.
-* Email: HaaS developers or anyone else wishing to stay up to date should subscribe to haas-dev-list@bu.edu by sending a plain text email to majordomo@bu.edu with "subscribe haas-dev-list" in the body.
+* IRL (In Real Life): The MOC group office is on the BU campus, at 3 Cummington Mall, Boston, MA room 451. Anyone interested in HIL is welcome to drop in and work there.
+* Email: HIL developers or anyone else wishing to stay up to date should subscribe to hil-dev-list@bu.edu by sending a plain text email to majordomo@bu.edu with "subscribe hil-dev-list" in the body.
```
jeremyfreudberg: incorrect replacement

setup.py
```diff
@@ -21,11 +21,11 @@ def _get_readme():
...
         return f.read()
 
-setup(name='haas',
+setup(name='hil',
       version='0.2rc2',
-      maintainer='Developers of the HaaS Project at MOC',
-      maintainer_email='haas-dev-list@bu.edu',
-      url='https://github.com/CCI-MOC/haas',
+      maintainer='Developers of the HIL Project at MOC',
+      maintainer_email='hil-dev-list@bu.edu',
```
jeremyfreudberg: incorrect replacement

docs/deployment.md
```diff
@@ -1,11 +1,11 @@
 # Deployment
 
 This is a guide to the best practices for deploying Hardware as a Service
-(HaaS). It documents many of the assumptions that the HaaS development team
-made in terms of how HaaS would be used, how it would be deployed, and the
+(HIL). It documents many of the assumptions that the HIL development team
```
jeremyfreudberg: replacement of abbreviation but not full name

docs/INSTALL-devel.rst
```diff
@@ -69,44 +69,44 @@ Start postgresql service::
...
+
+  sudo -u hil_dev -i
 
 Setup database and role to control it.
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Create database named `haas_dev` owned by user also named as `haas_dev`.
+Create database named `hil_dev` owned by user also named as `hil_dev`.
+
+1. Create a database role named `hil_dev` with priviledges to::
```
jeremyfreudberg: we misspell "privileges" as "priviledges" several times in the code base, and while not the topic of this PR, it does bring everyone's attention to it

zenhack:
@jeremyfreudberg, good catch; I've fixed those (except the spelling errors, which I opened another issue for).

zenhack:
Ok, the stalling postgres build seems to be reproducible; I'll have to look into it.

zenhack:
Still debugging, but one thing I caught is that we have the schema urls like:

    https://schema.masopencloud.org/haas/v0/switches/mock

which I failed to except from the change. Reverting this doesn't fix the issue, but it does need dealing with. If we want to change it we need a migration script. We might not though -- part of the reason for the name change was that we felt it was too generic -- like just naming openstac `IaaS` would have been. But the schema URL denotes a type of switch, so it may make sense for it to keep the generic name.

Thoughts?

zenhack:
So here's where I've gotten with debugging:

* It's hanging in the migration tests, on the first attempt to drop an existing table.
* If I pop open a psql shell, querying some of the tables will hang. I haven't really discovered a pattern, it doesn't seem to be the same each run.
* examining the pg_locks system table reveals that there are a lot of locks (many of them exclusive) being held in the postgres database.

I'm not entirely sure what to make of this. I'm going to put it down for a bit, come back to it later.

shwsun:
I will take a look at it today. 
@naved001 @henn @SahilTikale thoughts?

zenhack:
I noticed another issue; see the latest commit.

Another TODO on this:  Either the values for `api_name` in the various drivers need to stay as they are, or we need to write migration scripts. I think you can make a case for the former -- arguably it's a namespaced identifier for the swtich, so having the generic `haas` in there, as opposed to our "brand" `hil`, is OK. but I don't have a strong opinion.

Re: the hang, I'm still debugging. Here's what I'm seeing locally, letting it run for a while and then hitting Ctrl+C:

```
(.venv) [isd@rook hil]$ py.test tests/unit/migrations.py
================================================================================ test session starts =================================================================================
platform linux2 -- Python 2.7.13, pytest-3.1.2, py-1.4.34, pluggy-0.4.0
rootdir: /home/isd/src/pub/hil, inifile: setup.cfg
plugins: xdist-1.17.1, cov-2.5.1, catchlog-1.2.2
collected 2 items
^Cset option: 'hil.ext.switches.dell'
remove option: 'hil.ext.auth.null'
set option: 'hil.ext.switches.mock'
remove option: 'hil.ext.network_allocators.null'
set option: 'hil.ext.auth.database'
set option: 'hil.ext.obm.mock'
set option: 'hil.ext.network_allocators.vlan_pool'
set option: 'hil.ext.obm.ipmi'
set option: 'hil.ext.switches.nexus'
add section: 'hil.ext.network_allocators.vlan_pool'
set option: 'vlans'
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
INFO  [alembic.runtime.migration] Running upgrade  -> 6a8c19565060, Rename tables to account for Flask-SQLAlchemy's auto-naming.
INFO  [alembic.runtime.migration] Running upgrade 6a8c19565060 -> 89630e3872ec, network ACL
INFO  [alembic.runtime.migration] Running upgrade 89630e3872ec -> 57f4c30b0ad4, added metadata
INFO  [alembic.runtime.migration] Running upgrade 57f4c30b0ad4 -> 3b2dab2e0d7d, Add type field to NetworkingAction
INFO  [alembic.runtime.migration] Running upgrade 3b2dab2e0d7d -> c45f6a96dbe7, NIC primary key changed to BIGINT
INFO  [alembic.runtime.migration] Running upgrade  -> b5b31d19257d, Rename mockswitch table for Flask-SQLAlchemy

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! KeyboardInterrupt !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
to show a full traceback on KeyboardInterrupt use --fulltrace
/home/isd/src/pub/hil/.venv/lib/python2.7/site-packages/py/_process/forkedfunc.py:82: KeyboardInterrupt
=========================================================================== no tests ran in 58.16 seconds ============================================================================
(.venv) [isd@rook hil]$
```

zenhack:
So, it's actually just the *first* migration test that's a problem; if I comment out the first entry in the parametrize list, the other test goes through (though it fails without the URL changes I mentioned, which I just pushed).

Unfortunately, passing `--fulltrace` as suggested in the output doesn't provide much insight. Because we're doing `--boxed`, it just gives me a stack trace that ends in a call to `wait()`, on a process where the actual test is happening. If I get rid of `--boxed` (which should be OK since I'm only running one test), the process just doesn't respond to keyboard interrupts.

zenhack:
I've isolated the problem:

* The module names are used as branch names for our alembic migrations
* This patch changed all of the module names, and the branch names in the source.

Changing the branch names back from hil -> haas solves the problem, though I don't fully understand the details of the failure. I've got a local change that does this in a super hacky way (without actually moving the modules; it just renames stuff in the paths dict). Somehow, alembic is getting confused by the change.

Observation: `flask.sql`, which is used in the hanging test, does not have an alembic_version table, while the script used for the other migration test (which passes without this update) does have this table. I tried deleting it to see if this made the difference, but the migration failed for other reasons (it's trying to re-apply the migrations that are already there).

I'll need to poke at it a bit more to figure out exactly how this is confusing alembic; it's not actually obvious to me, especially since the names aren't in the database anywhere. I think I'm going to call it a night.

zenhack:
For reference, here is the hacky diff that fixes the problem:

```
diff --git a/hil/ext/switches/migrations/brocade/5a6db7a7222d_added_brocade_driver.py b/hil/ext/switches/migrations/brocade/5a6db7a7222d_added_brocade_driver.py
index 0b4e812..bab33eb 100644
--- a/hil/ext/switches/migrations/brocade/5a6db7a7222d_added_brocade_driver.py
+++ b/hil/ext/switches/migrations/brocade/5a6db7a7222d_added_brocade_driver.py
@@ -11,7 +11,7 @@ import sqlalchemy as sa
 # revision identifiers, used by Alembic.
 revision = '5a6db7a7222d'
 down_revision = None
-branch_labels = ('hil.ext.switches.brocade',)
+branch_labels = ('haas.ext.switches.brocade',)
 
 
 def upgrade():
diff --git a/hil/ext/switches/migrations/dell/099b939261c1_rename_dell_switch_table_for_flask_.py b/hil/ext/switches/migrations/dell/099b939261c1_rename_dell_switch_table_for_flask_.py
index a88aac2..737cb94 100644
--- a/hil/ext/switches/migrations/dell/099b939261c1_rename_dell_switch_table_for_flask_.py
+++ b/hil/ext/switches/migrations/dell/099b939261c1_rename_dell_switch_table_for_flask_.py
@@ -15,7 +15,7 @@ from hil.flaskapp import app
 # revision identifiers, used by Alembic.
 revision = '099b939261c1'
 down_revision = None
-branch_labels = ('hil.ext.switches.dell',)
+branch_labels = ('haas.ext.switches.dell',)
 
 
 def upgrade():
diff --git a/hil/ext/switches/migrations/mock/b5b31d19257d_rename_mockswitch_table_for_flask_.py b/hil/ext/switches/migrations/mock/b5b31d19257d_rename_mockswitch_table_for_flask_.py
index 946959e..5744686 100644
--- a/hil/ext/switches/migrations/mock/b5b31d19257d_rename_mockswitch_table_for_flask_.py
+++ b/hil/ext/switches/migrations/mock/b5b31d19257d_rename_mockswitch_table_for_flask_.py
@@ -15,7 +15,7 @@ from hil.flaskapp import app
 # revision identifiers, used by Alembic.
 revision = 'b5b31d19257d'
 down_revision = None
-branch_labels = ('hil.ext.switches.mock',)
+branch_labels = ('haas.ext.switches.mock',)
 
 
 def upgrade():
diff --git a/hil/ext/switches/migrations/n3000/b96d46bbfb12_add_dell_n3048_driver.py b/hil/ext/switches/migrations/n3000/b96d46bbfb12_add_dell_n3048_driver.py
index 7c42a1c..212ed13 100644
--- a/hil/ext/switches/migrations/n3000/b96d46bbfb12_add_dell_n3048_driver.py
+++ b/hil/ext/switches/migrations/n3000/b96d46bbfb12_add_dell_n3048_driver.py
@@ -13,7 +13,7 @@ import sqlalchemy as sa
 # revision identifiers, used by Alembic.
 revision = 'b96d46bbfb12'
 down_revision = None
-branch_labels = ('hil.ext.switches.n3000',)
+branch_labels = ('haas.ext.switches.n3000',)
 
 
 def upgrade():
diff --git a/hil/migrations.py b/hil/migrations.py
index f95aad4..4f9bc87 100644
--- a/hil/migrations.py
+++ b/hil/migrations.py
@@ -15,7 +15,7 @@ from alembic.script import ScriptDirectory
 # Extensions which use this facility must also use their module name as a
 # a branch_label on their migration scripts.
 paths = {
-    'hil': join(dirname(__file__), 'migrations', 'versions'),
+    'haas': join(dirname(__file__), 'migrations', 'versions'),
 }
 
 migrate = Migrate(app, db,
@@ -31,6 +31,10 @@ def _configure_alembic(config):
     """Customize alembic configuration."""
     # Configure the path for version scripts to include all of the directories
     # named in the `paths` dictionary, above:
+    for k, v in list(paths.items()):
+        if 'hil' in k:
+            paths[k.replace('hil', 'haas')] = v
+            del paths[k]
     config.set_main_option('version_locations',
                            ' '.join(paths.values()))
     return config
diff --git a/hil/migrations/versions/3b2dab2e0d7d_add_type_field_to_networkingaction.py b/hil/migrations/versions/3b2dab2e0d7d_add_type_field_to_networkingaction.py
index 4ef5ed1..c720fc4 100644
--- a/hil/migrations/versions/3b2dab2e0d7d_add_type_field_to_networkingaction.py
+++ b/hil/migrations/versions/3b2dab2e0d7d_add_type_field_to_networkingaction.py
@@ -14,7 +14,7 @@ from hil.model import NetworkingAction
 # revision identifiers, used by Alembic.
 revision = '3b2dab2e0d7d'
 down_revision = '57f4c30b0ad4'
-branch_labels = ('hil',)
+branch_labels = ('haas',)
 
 
 def upgrade():
```

Note the for loop that updates the keys in the paths dict.

zenhack:
Solved it -- see the last commit. Holy crap.

Note that this will probably still fail CI for the same reason #803 and #811 are.

zenhack:
I added a thing clarifying the terminology; I'm not sure it makes sense
long term to say we renamed it in there -- nobody really talks about
that thing that used to be called Quantum anymore (Neutron).

Also, thoughts on the URLs used in the type fields? I outlined my
thoughts above.

naved001:
> I added a thing clarifying the terminology; I'm not sure it makes sense long term to say we renamed it in there -- nobody really talks about that thing that used to be called Quantum anymore (Neutron).

Fair point. I am happy with the new commit. 

> Also, thoughts on the URLs used in the type fields? 

I agree with you on this. And also, given how much trouble migration scripts/alembic have caused I don't  want to have another migration script just to change names.

naved001:
@shwsun @henn @SahilTikale can we have a second review on this? Let's get this merged quickly so that we can raise new PRs that are blocked due to this. 

FYI, this PR also fixes (temporarily) the CI failures.

shwsun:
I will take a stab at it this afternoon.

zenhack:
@shwsun, good spot -- fixed.

And yeah, we need to prioritize the test performance. Can you weigh in in the comments on #764?

shwsun:
Ok so I have no input from @henn  or @SahilTikale . I am gonna give +1 and let @zenhack merge it.

zenhack:
Cool, merging.

PR #807, opened 2017-06-20
I guess you mean "Detaches"

xuhang57:
simple docstring fix.


haas/client/project.py
```diff
@@ -47,7 +47,7 @@ def connect(self, project_name, node_name):
                     )
 
         def detach(self, project_name, node_name):
-            """Adds a node to a project. """
+            """Detaches a node to a project. """
```
naved001: Correction: "Detaches a node from a project"

PR #818, opened 2017-07-06
keystoneauth1 lib imported in cli.py

Izhmash:
I needed to add these imports so that the keystone code could run correctly.  Python was reporting previously that `v3` was undefined.  More info here: https://docs.openstack.org/python-keystoneclient/latest/using-sessions.html

### DIFF
--- a/hil/cli.py
+++ b/hil/cli.py
@@ -133,6 +133,8 @@ def setup_http_client():
         return
     # Next try keystone:
     try:
+        from keystoneauth1.identity import v3
+        from keystoneauth1 import session
         os_auth_url = os.getenv('OS_AUTH_URL')
         os_password = os.getenv('OS_PASSWORD')
         os_username = os.getenv('OS_USERNAME')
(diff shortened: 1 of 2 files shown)


jeremyfreudberg:
Related commit (don't ask how I found this): https://github.com/SahilTikale/hil/commit/485ced3d75edb22491d397ef822120dc5d008bdf#diff-f91b561bea80fef80859e07202890ee0

SahilTikale:
makes sense. +1

zenhack:
Hm, can we add a regression test for this? Maybe in tests/integration/keystone we should make sure we invoke the cli at least once.

Otherwise this looks fine.

Izhmash:
@zenhack Would that be satisfied by calling `list_nodes all` and making sure the output matches what's expected?

zenhack:
@Izhmash, yeah that should work.

knikolla:
This looks good to me, agree on the requirement for integration testing.

In a follow-up improvement, we can avoid having to set the hil endpoint env variable in the case of keystone by registering hil in the service catalog and retrieving the endpoint url from the service catalog.

zenhack:
Assuming travis passes, +1

Issue #529, opened 2016-02-16
API call to list extensions active with HaaS


SahilTikale:
We need an API call to list the which extensions are available for a given HaaS installation. 

For eg.
User should be able to run a command and inquire which obm driver is active 
or which switch drivers are available. 

Users will be mostly system administrators managing HaaS.

Izhmash:
Would one API call that lists every extension suffice, or should there be an argument for the type of extension (auth, switches, etc.) to be described?

naved001:
I think an API that lists everything should suffice. Just package the response in a way such that it's easier to get the stuff we need. 

> should there be an argument for the type of extension (auth, switches, etc.) to be described

you can do that in the CLI

Izhmash:
I have an implementation of this finished, going to wait until the name change to present it.

naved001:
#820 fixes this.

PR #820, opened 2017-07-06
API call to list HIL active extensions

Izhmash:
- Addressing #529:
- API call returns all active extensions.
- CLI call accepts a string to narrow the results to a certain extension or group of extensions.
- For example, `hil list_active_extensions mock` will return only the mock extensions currently loaded.
- `hil list_active_extensions all` will return all active extensions.

### DIFF
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -1260,4 +1260,22 @@ Request Body:
 
 {}
 
+Response Body:
+
+[
+    [
+        "hil.ext.switches.mock",
+        ""
+    ],
+    [
+        "hil.ext.network_allocators.null",
+        ""
+    ],
+    ...
+]
+
 List all active extensions.
+
+Authorization requirements:
+
+* Administrative access.
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -1251,3 +1251,13 @@ Remove a user from a project.
 Authorization requirements:
 
 * Administrative access.
+
+#### list_active_extensions
+
+`GET /active_extensions`
+
+Request Body:
+
+{}
+
+List all active extensions.
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -1256,21 +1256,11 @@ Authorization requirements:
 
 `GET /active_extensions`
 
-Request Body:
-
-{}
-
 Response Body:
 
 [
-    [
-        "hil.ext.switches.mock",
-        ""
-    ],
-    [
-        "hil.ext.network_allocators.null",
-        ""
-    ],
+    "hil.ext.switches.mock",
+    "hil.ext.network_allocators.null",
     ...
 ]
 
--- a/hil/api.py
+++ b/hil/api.py
@@ -1227,6 +1227,15 @@ def list_headnode_images():
     return json.dumps(valid_imgs)
 
 
+# Extension code #
+#################
+@rest_call('GET', '/active_extensions', Schema({}))
+def list_active_extensions():
+    """List all active extensions"""
+    extensions = cfg.items('extensions')
+    return json.dumps(extensions)
+
+
 # Console code #
 ################
 @rest_call('GET', '/node/<nodename>/console', Schema({'nodename': basestring}))
--- a/hil/api.py
+++ b/hil/api.py
@@ -1233,8 +1233,7 @@ def list_headnode_images():
 def list_active_extensions():
     """List all active extensions"""
     get_auth_backend().require_admin()
-    extensions_raw = cfg.items('extensions')
-    extensions = sorted([ext[0] for ext in extensions_raw])
+    extensions = [ext[0] for ext in cfg.items('extensions')]
     return json.dumps(extensions)
 
 
--- a/hil/api.py
+++ b/hil/api.py
@@ -1234,7 +1234,7 @@ def list_active_extensions():
     """List all active extensions"""
     get_auth_backend().require_admin()
     extensions_raw = cfg.items('extensions')
-    extensions = [ext[0] for ext in extensions_raw]
+    extensions = sorted([ext[0] for ext in extensions_raw])
     return json.dumps(extensions)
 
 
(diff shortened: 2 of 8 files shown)


hil/cli.py
```diff
@@ -797,6 +797,27 @@ def create_admin_user(username, password):
 
 
 @cmd
+def list_active_extensions(ext_type):
+    """List active extensions by type.
```
naved001: this should be an admin only command imo

docs/rest_api.md
```diff
@@ -1251,3 +1251,13 @@ Remove a user from a project.
...
+
+#### list_active_extensions
+
+`GET /active_extensions`
+
+Request Body:
+
+{}
+
+List all active extensions.
```
naved001: You should add a bit about the authorization requirements and how the response body looks like.

Izhmash:
Addressed @naved001's issues and added some tests.

naved001:
I know it sounds redundant but we should directly test the API too in `tests/unit/api/main/py`.

hil/cli.py
Izhmash: @SahilTikale I fixed that up in my latest commit

docs/rest_api.md
```diff
@@ -1251,3 +1251,31 @@ Remove a user from a project.
...
+    [
+        "hil.ext.switches.mock",
+        ""
+    ],
+    [
+        "hil.ext.network_allocators.null",
+        ""
+    ],
+    ...
+]
```
zenhack: Rather than handing back the nested list we get from configparser, just return a list of the names of the extensions.

hil/cli.py
```diff
@@ -797,6 +797,27 @@ def create_admin_user(username, password):
...
+    """
+    if not config.cfg.has_section('extensions'):
+        sys.exit("There are no active extensions.")
+
+    all_extensions = C.extensions.list_active()
+    if ext_type == 'all':
+        active = [ext[0] for ext in all_extensions]
+        sys.stdout.write('All active extensions: ' + ', '.join(active) + '\n')
+    else:
+        active = [ext[0] for ext in all_extensions if ext_type in ext[0]]
```
zenhack: This is not correct in the general case. There's no reason extension modules have to be at any particular path/name. We have a particular structure to the extensions we ship in-tree, but (a) it's just a convention, and (b) the extension mechanism was designed with third party extensions in mind. An extension could be provided by some other package, and we have no guarantees about the name of the module.

Furthermore, there's not really a way of classifying extensions in general; extensions aren't restricted to simply providing backends for specific mechanisms.

hil/cli.py
```diff
@@ -797,6 +797,27 @@ def create_admin_user(username, password):
...
+
+    Use the 'all' argument to list all active extensions.
+    """
+    if not config.cfg.has_section('extensions'):
+        sys.exit("There are no active extensions.")
+
+    all_extensions = C.extensions.list_active()
+    if ext_type == 'all':
+        active = [ext[0] for ext in all_extensions]
+        sys.stdout.write('All active extensions: ' + ', '.join(active) + '\n')
```
zenhack: I would print each on its own line (and without the description), to make the output usable with e.g. grep.

hil/api.py
```diff
@@ -1227,6 +1227,16 @@ def list_headnode_images():
     return json.dumps(valid_imgs)
 
 
+# Extension code #
+#################
+@rest_call('GET', '/active_extensions', Schema({}))
```
Izhmash: Since I'm not expecting any input, should I just validate that nothing is there?

Izhmash: I was trying to mirror the call just above mine, `list_headnode_images():`

hil/api.py
```diff
@@ -1233,7 +1233,8 @@ def list_headnode_images():
 def list_active_extensions():
     """List all active extensions"""
     get_auth_backend().require_admin()
-    extensions = cfg.items('extensions')
+    extensions_raw = cfg.items('extensions')
+    extensions = [ext[0] for ext in extensions_raw]
```
zenhack: Can combine this into one line:

```python
extensions = [ext[0] for ext in cfg.items('extensions')]
```

hil/cli.py
```diff
@@ -797,24 +797,13 @@ def create_admin_user(username, password):
...
-    """List active extensions by type.
-
-    Default types: auth, obm, switches, network_allocators.
-
-    Use the 'all' argument to list all active extensions.
-    """
+def list_active_extensions():
+    """List active extensions by type. """
     if not config.cfg.has_section('extensions'):
         sys.exit("There are no active extensions.")
```
zenhack: This doesn't make any sense -- the extensions section is only relevant to the server, so the client shouldn't be checking it.

zenhack:
Ack, I put my comments on a single commit diff, not the final thing, so they're already showing up as outdated. The meat of those comments is still relevant to the current version though; please address them.

hil/cli.py
Izhmash: I modified this to check if the list from the API is empty rather than from the config.

Izhmash:
Everything should be addressed now.

hil/api.py
SahilTikale: I missed that you were doing that in CLI. In that case you can ignore this comment.

hil/api.py
```diff
@@ -1233,8 +1233,7 @@ def list_headnode_images():
 def list_active_extensions():
     """List all active extensions"""
     get_auth_backend().require_admin()
-    extensions_raw = cfg.items('extensions')
-    extensions = sorted([ext[0] for ext in extensions_raw])
+    extensions = [ext[0] for ext in cfg.items('extensions')]
```
zenhack: You can keep the call to `sorted`; sorry, my comment had been on the commit before you added that.

zenhack: ...and I am apparently really bad at using github's review feature for some reason. LGTM.

zenhack:
@SahilTikale, I think we're just waiting on you to approve the changes; I think your comments have been addressed.

naved001:
@SahilTikale ping

SahilTikale:
+1 merging.

Issue #804, opened 2017-06-15
Ensure that VLAN ids do not collide


henn:
If one manually creates a network as an admin with an explicit VLAN id, then it is possible that the network ID / VLAN could collide with one in the VLAN pool.

If that were to happen, then someone else could end up with a network from the pool with the same VLAN id as the explicitly created network.

This seems like a bug to me. Can anyone think of a valid use case?
If not, then I suggest we put a check in *network_create()* that ensures a given ID is not a part of the network allocator pool, and also add an interface to the network allocator to check if an ID is in the pool.

If the network needed to be shared across projects, then HIL's network ACL support should be able to meet that need while keeping integrity.

zenhack:
We could debate the semantics of the word "bug" -- there have historically been numerous ways for an admin to break things if not careful, and this is one. We should definitely add a check.

naved001:
I [started working](https://github.com/CCI-MOC/hil/compare/master...naved001:unique-vlanids?expand=1) on this, and there are couple of issues that I ran into.

To make sure that VLAN ids don't collide I must mark them unavailable in the database. So, as long as the Vlan that an admin is using belongs to the allocation pool I can find it in the database and mark it unavailable (because [populate](https://github.com/CCI-MOC/hil/blob/master/hil/ext/network_allocators/vlan_pool.py#L47) adds the vlan to the database)

But if an admin uses a net_id outside of the allocation pool, then I can't mark it unavailable. If I add it to the db then it becomes a part of the regular vlan pool and hil might give out that vlan ID to some regular user. I am not sure how to handle this. (Maybe add an attribute to the Vlan class to identify if it belongs to the pool? )

Thoughts?

zenhack:
So the situation you're concerned about is the one where hil.cfg is updated after-the-fact, to include a vlan that an admin has already manually assigned somewhere?

One thing we could do: in populate() we could check for any pre-existing networks using the same net_id that we're about to add. If we find one, we could set available = False.

henn:
We've discussed two use cases for when an admin manually sets the VLAN for a new network:
1. When a VLAN is in the pool
2. When a VLAN is not in the pool.

The first case is the very worrisome that this issue is concerned with. For this one, we will mark the VLAN as being in use, and release it when the network is free'd. If it is in use, then we'll disallow the network create.

For the second use case, it's more debatable; I'm not convinced it's something we need to worry about anytime soon, but @SahilTikale disagrees.

[...]

PR #823, opened 2017-07-10
Modified vlan_pool network allocator to ensure VLAN IDs dont collide.

naved001:
Issue #804 

* Raises AllocationError if an admin tries to make a network with a net_id
 from the pool that is already in use
* method validate_network_id returns a boolean if the net_id belongs to the
 allocation pool or not.
* it raises a BadArgumentError if the net_id is invalid i.e., greater than 4096
 or less than 1

I did not want to add more methods to the VlanAllocator, so decided to just modify `validate_network_id` because it was introduced only for  admin networks. Let me know if this seems like a reasonable approach to address #804. If yes, then I'll fix the tests that are broken atm and write some more tests. If not, any suggestions?

### DIFF
--- a/hil/api.py
+++ b/hil/api.py
@@ -798,13 +798,9 @@ def network_create(network, owner, access, net_id):
     # Allocate net_id, if requested
     if net_id == "":
         net_id = get_network_allocator().get_new_network_id()
-        if net_id is None:
-            raise AllocationError('No more networks')
         allocated = True
     else:
-        if not get_network_allocator().validate_network_id(net_id):
-            raise BadArgumentError("Invalid net_id")
-        allocated = False
+        allocated = get_network_allocator().validate_network_id(net_id)
 
     network = model.Network(owner, access, allocated, net_id, network)
     db.session.add(network)
--- a/hil/api.py
+++ b/hil/api.py
@@ -800,15 +800,12 @@ def network_create(network, owner, access, net_id):
         net_id = get_network_allocator().get_new_network_id()
         if net_id is None:
             raise AllocationError('No more networks')
-        allocated = True
     else:
         if not get_network_allocator().validate_network_id(net_id):
             raise BadArgumentError("Invalid net_id")
-        if not get_network_allocator().network_id_available(net_id):
-            raise AllocationError("Network ID is in use. Please choose a "
-                                  "different net_id")
-        allocated = get_network_allocator().is_network_id_in_pool(net_id)
+        get_network_allocator().claim_network_id(net_id)
 
+    allocated = get_network_allocator().is_network_id_in_pool(net_id)
     network = model.Network(owner, access, allocated, net_id, network)
     db.session.add(network)
     db.session.commit()
--- a/hil/api.py
+++ b/hil/api.py
@@ -798,9 +798,16 @@ def network_create(network, owner, access, net_id):
     # Allocate net_id, if requested
     if net_id == "":
         net_id = get_network_allocator().get_new_network_id()
+        if net_id is None:
+            raise AllocationError('No more networks')
         allocated = True
     else:
-        allocated = get_network_allocator().validate_network_id(net_id)
+        if not get_network_allocator().validate_network_id(net_id):
+            raise BadArgumentError("Invalid net_id")
+        if not get_network_allocator().network_id_available(net_id):
+            raise AllocationError("Network ID is in use. Please choose a "
+                                  "different net_id")
+        allocated = get_network_allocator().is_network_id_in_pool(net_id)
 
     network = model.Network(owner, access, allocated, net_id, network)
     db.session.add(network)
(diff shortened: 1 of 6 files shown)


naved001:
cc: @henn

zenhack:
I'm ok with modifying the contract for validate_network_id in theory, but the specifics of what this patch does are a little clumsy: before it just returned yes or no, is this valid? now it does a few different checks, using exceptions to report some things and the return value to report others. This is generally a sign that you're trying to do two unrelated things in one function.

I would just go ahead and add another method, `network_id_is_used` or such, that checks the new thing.

naved001:
You are right; it is clumsy. I added a new commit. How does this look?

I have added two new functions; `network_id_available` and `is_network_id_in_pool`. 

I think I can merge `validate_network_id` with `network_id_available` but then if it only returns True or False then we can't differentiate between `net_id` being invalid vs not available (BadArgument vs Allocation Error). If we can move the error handling into the driver then we can raise the appropriate error and not return anything. 

`is_network_id_in_pool` would be required anyway.

naved001:
@zenhack what do you think of the new commit?

hil/ext/network_allocators/vlan_pool.py
```diff
@@ -69,6 +69,20 @@ def validate_network_id(self, net_id):
...
+            vlan.available = False
+            return True
+        elif vlan and not vlan.available:
+            return False
+        else:
+            return True
+
+    def is_network_id_in_pool(self, net_id):
+        vlan = Vlan.query.filter_by(vlan_no=net_id).first()
+        return True if vlan else False
```
zenhack: `vlan is not None`

hil/ext/network_allocators/vlan_pool.py
```diff
@@ -69,6 +69,20 @@ def validate_network_id(self, net_id):
         except ValueError:
             return False
 
+    def network_id_available(self, net_id):
```
zenhack: This functions name suggests it merely does a check, but in fact it has a side effect. Another way to do this would be to have a `claim_network_id`, which raised an AllocationError if the id is unavailable.

hil/api.py
```diff
@@ -800,7 +800,10 @@ def network_create(network, owner, access, net_id):
     else:
         if not get_network_allocator().validate_network_id(net_id):
             raise BadArgumentError("Invalid net_id")
-        allocated = False
+        if not get_network_allocator().network_id_available(net_id):
+            raise AllocationError("Network ID is in use. Please choose a "
+                                  "different net_id")
+        allocated = get_network_allocator().is_network_id_in_pool(net_id)
```
zenhack: At this point in the code, this is still always going to be False. Two sensible options:

* change this back to `allocated = False`
* move it out of the else block, so that it is executed regardless of which branch is taken, and remove `allocated = True` from the first branch.

naved001: why would it always be False? If `net_id` belongs to the vlan pool, then even though it wasn't "allocated" by HIL, we still need to mark it as `True` otherwise we can't free the network in [`network_delete`](https://github.com/CCI-MOC/hil/blob/master/hil/api.py#L838)

hil/ext/network_allocators/vlan_pool.py
naved001: > Another way to do this would be to have a claim_network_id, which raised an AllocationError if the id is unavailable.

I like this, I am gonna do that.

naved001:
@zenhack I have added unit tests for vlan_pool. it's ready for review.

tests/unit/ext/network_allocators/vlan_pool.py
```diff
@@ -76,3 +77,60 @@ def test_vlanid_for_admin_network():
...
+class TestAdminCreatedNetworks():
+    """
+    Test networks created by administrator
+    """
+
+    def test_create_network_with_id_from_pool(self):
+
+        api.project_create('nuggets')
+
+        # create a project owned network and get it's network_id
```
Izhmash: Minor, should be 'its'

naved001: changed!

hil/ext/network_allocators/vlan_pool.py
```diff
@@ -69,6 +70,21 @@ def validate_network_id(self, net_id):
...
+    def claim_network_id(self, net_id):
+        vlan = Vlan.query.filter_by(vlan_no=net_id).first()
+        if vlan and vlan.available:
+            vlan.available = False
+            return
+        elif vlan and not vlan.available:
+            raise BlockedError("Network ID is not available."
+                               " Please choose a different ID.")
+        else:
+            return
```
zenhack: The if-else ladder here is a bit awkward. I'd do something more like:

```python

if vlan is None:
    return
elif vlan.available:
    vlan.available = False
else:
    raise BlockedError(...)
```

PR #827, opened 2017-07-14
Make clientlib tests faster, simpler, and more robust.

zenhack:
Turns out I was a little off re: where the slowdown was coming from; every request made in the populate_server() fixture was triggering a call to sha512_crypt.verify, which is (intentionally) very slow.

This patch mocks that out during the client tests, since we aren't actually testing that bit, but do need that backend loaded for some of the tests.

I also still went ahead with ripping out the HTTP server; some of the other benefits are described in the commit message.

### DIFF
--- a/tests/unit/client.py
+++ b/tests/unit/client.py
@@ -13,140 +13,139 @@
 # governing permissions and limitations under the License.
 
 """Unit tests for client library"""
-from flask.ext.sqlalchemy import SQLAlchemy
 from hil.flaskapp import app
-from hil.model import NetworkingAction
 from hil.client.base import ClientBase, FailedAPICallException
-from hil.client.client import Client, RequestsHTTPClient
+from hil.client.client import Client, HTTPClient, HTTPResponse
+from hil.test_common import config_testsuite, config_merge, \
+    fresh_database, fail_on_log_warnings
+from hil.model import db
+from hil import config, deferred, server
 
 import json
-import os
 import pytest
-import requests
-import sys
-import tempfile
-import time
 
-from subprocess import check_call, Popen
+from urlparse import urlparse
+from base64 import urlsafe_b64encode
+from passlib.hash import sha512_crypt
 
-ep = "http://127.0.0.1:8000" or os.environ.get('HIL_ENDPOINT')
-username = "hil_user" or os.environ.get('HIL_USERNAME')
-password = "hil_pass1234" or os.environ.get('HIL_PASSWORD')
+ep = "http://127.0.0.1:8000"
+username = "hil_user"
+password = "hil_pass1234"
 
-http_client = RequestsHTTPClient()
-http_client.auth = (username, password)
+
+class FlaskHTTPClient(HTTPClient):
+
+    def __init__(self):
+        self._flask_client = app.test_client()
+
+    def request(self, method, url, data=None, params=None):
+
+        # Flask doesn't provide a straightforward way to do basic auth,
+        # but it's not actually that complicated:
+        auth_header = 'Basic ' + urlsafe_b64encode(username + ':' + password)
+
+        resp = self._flask_client.open(
+            method=method,
+            headers={'Authorization': auth_header},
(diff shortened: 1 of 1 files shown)


zenhack:
On my laptop, the client tests took a total of about 2 minutes, vs 50 minutes before the change.

naved001:
Whoa!  Seeing Travis only run for like 10 minutes feels weird now haha.

tests/unit/client.py
```diff
@@ -13,140 +13,139 @@
...
+    Password hashing is just way too slow to do it this many times in
+    the test suite.
     """
 
-    os.remove('hil.cfg')
-    os.remove('hil.db')
-    os.chdir(cwd)
-    os.rmdir(tmpdir)
-
+    @staticmethod
```
Izhmash: Why do we need the dummy function?

tests/unit/client.py
```diff
@@ -13,140 +13,139 @@
...
+        )
+        return HTTPResponse(status_code=resp.status_code,
+                            headers=resp.headers,
+                            content=resp.get_data())
+
+http_client = FlaskHTTPClient()
 C = Client(ep, http_client)  # Initializing client library
 MOCK_SWITCH_TYPE = 'http://schema.massopencloud.org/haas/v0/switches/mock'
 OBM_TYPE_MOCK = 'http://schema.massopencloud.org/haas/v0/obm/mock'
 OBM_TYPE_IPMI = 'http://schema.massopencloud.org/haas/v0/obm/ipmi'
```
naved001: We aren't using these constants anywhere in the tests. Might as well remove it (or use it) while we are cleaning up the tests.

tests/unit/client.py
```diff
@@ -13,140 +13,139 @@
...
-#       4. Populates hil with dummy objects
-#       5. tears down the setup in a clean fashion.
+fail_on_log_warnings = pytest.fixture(fail_on_log_warnings)
+fresh_database = pytest.fixture(fresh_database)
 
-db_dir = None
 
+@pytest.fixture
+def dummy_verify():
+    """replace sha512_crypt.verify with something faster (albeit broken).
```
naved001: I don't really understand how this works. 

1. What was calling it (db-auth?) 
2. How are we over-riding/mocking the original ?
3. Why do we keep `sha512_crypt.verify` in old and then restore it after yield? It still works if I don't do this.

zenhack:
Quoting Ian Ballou (2017-07-17 10:00:52)

>    Why do we need the dummy function?

It's used to replace verify() for the duration of the test -- which is
where all of the time had been going. See the assignment statements
immediately below. (the test gets run where `yield` is executed).

zenhack:
Quoting Naved Ansari (2017-07-17 10:40:23)

>  MOCK_SWITCH_TYPE = 'http://schema.massopencloud.org/haas/v0/switches/mock'
>  OBM_TYPE_MOCK = 'http://schema.massopencloud.org/haas/v0/obm/mock'
>  OBM_TYPE_IPMI = 'http://schema.massopencloud.org/haas/v0/obm/ipmi'
>
>
>    We aren't using these constants anywhere in the tests. Might as well
>    remove it (or use it) while we are cleaning up the tests.

Will do.

> +@pytest.fixture
> +def dummy_verify():
> +    """replace sha512_crypt.verify with something faster (albeit broken).
>
>    I don't really understand how this works.
>     1. What was calling it (db-auth?)
>     2. How are we over-riding/mocking the original ?
>     3. Why do we keep sha512_crypt.verify in old and then restore it after
>        yield? It still works if I don't do this.

It works, but it's *slow*. This is where most of the time was going.
Since both of you had questions, I'm going to take that as a sign that
the docs need to be better; I'll update it.

zenhack:
@Izhmash, @naved001, does the new docstring make things more clear?

Also, @naved001, I got rid of those constants.

naved001:
2 sign offs, so merging.

PR #829, opened 2017-07-18
* Added method `ensure_config_file_exists()` that exits if the config…

### LINKED ISSUE #338: haas doesn't detect missing config file
It just complains when an option it needs is missing.

naved001:
… file is not found.
* Calling this function before the hil admin commands, `serve`,
`serve_networks`, and `create admin user`.

Fix #338

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,9 +36,5 @@ install:
   - sh -e ci/keystone/install.sh
 
 script:
-  - sh -e ci/pep8_tests.sh
-  - sh -e ci/pylint_tests.sh
-  - sh -e ci/run_unit_tests.sh
   - sh -e ci/apache/run_integration_tests.sh
-  - sh -e ci/keystone/run_integration_tests.sh
   - sh -e ci/deployment-mock-networks/run_integration_tests.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,5 +36,9 @@ install:
   - sh -e ci/keystone/install.sh
 
 script:
+  - sh -e ci/pep8_tests.sh
+  - sh -e ci/pylint_tests.sh
+  - sh -e ci/run_unit_tests.sh
   - sh -e ci/apache/run_integration_tests.sh
+  - sh -e ci/keystone/run_integration_tests.sh
   - sh -e ci/deployment-mock-networks/run_integration_tests.sh
--- a/docs/INSTALL-devel.rst
+++ b/docs/INSTALL-devel.rst
@@ -221,7 +221,8 @@ Configuring HIL
 -----------------
 
 Now the ``hil`` executable should be in your path. First, create a
-configuration file ``hil.cfg``. There are two examples for you to work from,
+configuration file ``hil.cfg`` because if it's not found then hil would refuse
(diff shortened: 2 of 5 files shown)


zenhack:
Can we just make use of the return value from ConfigPareser.read? That way there's no race condition.

naved001:
@zenhack Done. But I don't understand the race condition though, can you explain that bit? Thanks!

zenhack:
Yeah, it's still there; I think you misunderstood my suggestion. Here's the scenario:

* we check for the config file, finding that it's there.
* the file is deleted
* we try to load the config file.

My suggestion was, in the existing call ro read that loads the config, we check the return value to determine if loading was successful, and act based on that.

naved001:
I am still not clear what needs to be done. In the existing call to `read` [here](https://github.com/CCI-MOC/hil/blob/master/hil/config.py#L38), the docstring says that the function will load an empty configuration. So is that gonna change now? Like if we don't find the file we `sys.exit` at that point ?

zenhack:
Yeah, that's the idea.

naved001:
Done.

zenhack:
Should document the exiting behavior. Also, now that your other pr has been merged, you should rebase to get the fix.

naved001:
I documented it in [`docs/INSTALLATION.rst`](https://github.com/CCI-MOC/hil/blob/master/docs/INSTALL.rst)
and [`docs/INSTALLATION-devel.rst`](https://github.com/CCI-MOC/hil/blob/master/docs/INSTALL-devel.rst)

zenhack:
Ack, I meant in the docstring to config.load -- I actually think it's already clear enough from the user perspective (though I don't object to haveing it there), but the programmatic API should be more clear.

naved001:
fixed the docstring. still need to figure out the failing integration tests. Will take a look at it tomorrow.

zenhack:
Think I know what's happening with the tests: We currently load the config file from the CLI, even when not launching the server. We probably shouldn't -- it doesn't really make sense to. IIRC you can still set the server endpoint there, but nothing else is used.

We should probably just move the call to load a bit later, after we've determined that we're running a command where it's relevant (e.g. serve or serve_networks)

naved001:
Not sure if I understand the issue here.

> We currently load the config file from the CLI, even when not launching the server. 

But the integration tests are using the CLI, so the file is loaded in it's [`main()`](https://github.com/CCI-MOC/hil/blob/master/hil/cli.py#L835) and the tests are complaining file not found. I don't understand why is that so? since `ci/apache/install.sh` copy and place `hil.cfg` in `/etc/`

> We should probably just move the call to load a bit later, after we've determined that we're running a command where it's relevant (e.g. serve or serve_networks)

How would that help?

zenhack:
The file in /etc is only used by hil.wsgi, so it's not going to affect the cli.

re: moving the call later, my suggestion is that for client operations it doesn't make sense to load the file *at all*, so instead we move the call to config.setup to a code-path that only gets executed by the commands that need it (serve, serve_networks...)

naved001:
@zenhack done.

PR #831, opened 2017-07-19
More documentation on setting up Keystone + HIL

Izhmash:
- Naved and I recently setup HIL in a development environment.
- I've added some documentation that resolved issues we had in setting it up.

### DIFF
--- a/docs/INSTALL.rst
+++ b/docs/INSTALL.rst
@@ -395,6 +395,8 @@ former is a WSGI application, which we recommend running with Apache's
 (The file may already exist, with just the ``LoadModule`` option. If so, it is
 safe to replace it.)
 
+**Note:** if accessing HIL through a public IP address, be sure to change the ``VirtualHost`` and ``ServerName`` IP addresses to match the public one.
+
 **Note:** certain calls to HIL such as *port_register()* may pass arbitrary
 strings that should be escaped (see `issue 361 <https://github.com/CCI-MOC/hil/issues/360>`_). By default, Apache `Doesn't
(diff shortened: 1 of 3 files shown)


docs/INSTALL.rst
```diff
@@ -395,6 +395,8 @@ former is a WSGI application, which we recommend running with Apache's
 (The file may already exist, with just the ``LoadModule`` option. If so, it is
 safe to replace it.)
 
+**Note:** if accessing HIL through a public IP address, be sure to change the ``VirtualHost`` and ``ServerName`` IP addresses to match the public one.
```
henn: I'd suggest making this something more like `be sure to update your VirtualHost entry accordingly`, with a link to the docs at https://httpd.apache.org/docs/current/mod/core.html#virtualhost

docs/USING.rst
```diff
@@ -26,6 +26,9 @@ which HIL instance to use, be sure to do one of:
 If using the basic auth/database auth backend, you must set the environment
 variables ``HIL_USERNAME`` and ``HIL_PASSWORD`` to the correct credentials.
 
+If using the auth/keystone auth backend, first make sure that the keystonemiddleware library is installed by running ``pip install keystonemiddleware``.
+Make sure to also source the OpenStack authentication credentials for the correct user and project.
```
henn: `source` might be ambiguous here. Maybe `Make sure to ensure the appropriate OpenStack credential environmental variables are set.`

zenhack:
@henn has already brought up the couple of issues I would have mentioned. Otherwise this looks ok.

naved001:
@henn said he only made drive-by comments and that we can merge it.

Issue #631, opened 2016-07-28
Optional parameter to do hard power_cycle


henn:
Currently, the [reset API](https://github.com/CCI-MOC/hil/blob/master/docs/rest_api.md#node_power_cycle) does a "soft reset", based on [this code](https://github.com/CCI-MOC/hil/blob/master/haas/ext/obm/ipmi.py#L68):

``` python
    @no_dry_run
    def power_cycle(self):
        self._ipmitool(['chassis', 'bootdev', 'pxe'])
        if self._ipmitool(['chassis', 'power', 'cycle']) == 0:
            return
        if self._ipmitool(['chassis', 'power', 'on']) == 0:
            # power cycle will fail if the machine is not running.
            # To avoid such a situation, just turn it on anyways.
            # Doing this saves power by turning things off without
            # Without breaking the HaaS.
            return
        # If it is still does not work, then it is a real error:
        raise OBMError('Could not power cycle node %s' % self.node.label)
```

We should introduce an optional parameter (like `force`) that does an `ipmitool chassis power reset`. Here's what the ipmitool man page says about the different states:

```
Performs a chassis control  command  to  view  and change the power state.

status

       Show current chassis power status.

on

       Power up chassis.

off

       Power  down  chassis  into  soft off (S4/S5 state).  WARNING: This command does not
       initiate a clean shutdown of the operating system prior to powering down  the  sys‐
       tem.

cycle

       Provides  a  power  off  interval  of at least 1 second.  No action should occur if
       chassis power is in S4/S5 state, but it is recommended to check power  state  first
       and  only issue a power cycle command if the  system  power is on or in lower sleep
       state than S4/S5.

reset

       This command will perform a hard reset.

diag

       Pulse a diagnostic interrupt (NMI) directly to the processor(s).

soft

       Initiate a soft-shutdown of OS via ACPI.  This can be done in  a  number  of  ways,
       commonly by simulating an overtemperture or by simulating a power button press.  It
       is necessary for there to be Operating System support for ACPI  and  some  sort  of
       daemon watching for events for this soft power to work.
```

henn:
@apoorvemohan and @ravisantoshgudimetla are interested in this so they can reset machines faster than the current implementation.

matsuuran:
[...]

PR #835, opened 2017-07-20
Hard power cycle option

Izhmash:
- In response to issue #631:
- Adds the option to force a hard reboot via API and Client libraries.
- Default is still to soft reboot the system.

### DIFF
--- a/hil/ext/obm/ipmi.py
+++ b/hil/ext/obm/ipmi.py
@@ -68,10 +68,14 @@ class Ipmi(Obm):
         return status
 
     @no_dry_run
-    def power_cycle(self):
+    def power_cycle(self, force):
         self._ipmitool(['chassis', 'bootdev', 'pxe'])
-        if self._ipmitool(['chassis', 'power', 'cycle']) == 0:
-            return
+        if force:
+            if self._ipmitool(['chassis', 'power', 'reset']) == 0:
+                return
+        else:
+            if self._ipmitool(['chassis', 'power', 'cycle']) == 0:
+                return
         if self._ipmitool(['chassis', 'power', 'on']) == 0:
             # power cycle will fail if the machine is not running.
             # To avoid such a situation, just turn it on anyways.
--- a/hil/ext/obm/ipmi.py
+++ b/hil/ext/obm/ipmi.py
@@ -68,14 +68,14 @@ class Ipmi(Obm):
         return status
 
     @no_dry_run
-    def power_cycle(self, force=False):
+    def power_cycle(self, force):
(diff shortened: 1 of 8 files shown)


naved001:
Did you test this on a real node?

Izhmash:
@naved001 I haven't yet, is there a server setup that I can try that on?

naved001:
yep, Ill message you.

naved001:
@Izhmash and I tested this on a real machine and it works as expected. yay!

hil/ext/obm/ipmi.py
```diff
@@ -68,10 +68,14 @@ def _ipmitool(self, args):
         return status
 
     @no_dry_run
-    def power_cycle(self):
+    def power_cycle(self, force=False):
```
zenhack: I would *not* have a default for force here (in the API call and client library it makes sense). We're only invoking this from the one place, so having the default just introducess pointless complexity. Also, the mock driver doesn't have the default, so the APIs aren't actually the same, which is a problem.

hil/ext/obm/ipmi.py
```diff
@@ -68,10 +68,14 @@ def _ipmitool(self, args):
...
         self._ipmitool(['chassis', 'bootdev', 'pxe'])
-        if self._ipmitool(['chassis', 'power', 'cycle']) == 0:
-            return
+        if force:
+            if self._ipmitool(['chassis', 'power', 'reset']) == 0:
+                return
+        else:
+            if self._ipmitool(['chassis', 'power', 'cycle']) == 0:
+                return
```
zenhack: Maybe factor out the differences here by using a variable:

```python
if force:
    op = 'reset'
else:
    op = 'cycle'
self._ipmitool(...)
...
```

Makes it easier to see what's going on.

tests/unit/client.py
```diff
@@ -319,6 +319,9 @@ def test_show_node(self):
     def test_power_cycle(self):
         assert C.node.power_cycle('node-07') is None
 
+    def test_power_cycle_force(self):
```
henn: Please add a test for `force=False` and `force=NotABoolean` as well, so we can be complete

Izhmash:
Everything should be set now.

zenhack:
LGTM. @henn?

naved001:
I am happy with this, just one change; Can you update/add-to the deployment [tests here](https://github.com/CCI-MOC/hil/blob/master/tests/deployment/ipmi.py#L62) to test both cases (`cycle` vs `reset`) on real machines?

Izhmash:
Added test for hard reboot as well.

PR #838, opened 2017-07-24
Auth check api

### LINKED ISSUE #692: API for explicit auth checks
As we're doing more and more integrations (BMI, Net Ex, VPN, ...), the need for services to verify that a user has access to things like a node keeps coming up.

The use case we discussed today was having a dedicated, separate console proxy service. It would need to be able to verify that a particular user had access to a node (without checking with HIL, it would have no way of knowing which nodes were in which projects if keystone were used).

Currently the way to check this would be to do something like calling show_node(), which is hacky.

I propose we have an explicit API call for these types of integrations. It could be something like `GET /node/node1/is_authorized` or something else.

naved001:
* Fix #692 
* Added an API to check if a user is authorized to access a node.

### DIFF
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -1269,3 +1269,19 @@ List all active extensions.
 Authorization requirements:
 
 * Administrative access.
+
+#### is_authorized
+
+`GET /node/<nodename>/is_authorized`
+
+Response Body:
+
+{
+    "authorized": <boolean>
+}
+
+Returns whether or not a user is authorized to access <node>
+
+Authorization requirements:
+
+* No special access.
--- a/hil/api.py
+++ b/hil/api.py
@@ -1278,9 +1278,12 @@ def stop_console(nodename):
     'nodename': basestring,}))
 def is_authorized(nodename):
     node = _must_find(model.Node, nodename)
-    if node.project is None:
-        return json.dumps(False)
-    return json.dumps(get_auth_backend().have_project_access(node.project))
+    if get_auth_backend().have_admin():
+        return json.dumps({'authorized': True})
+    elif node.project is None:
+        return json.dumps({'authorized': False})
+    return json.dumps({
+        'authorized': get_auth_backend().have_project_access(node.project)})
 
 # Helper functions #
 ####################
--- a/hil/api.py
+++ b/hil/api.py
@@ -1275,7 +1275,7 @@ def stop_console(nodename):
 
 
 @rest_call('GET', '/node/<nodename>/is_authorized', Schema({
-    'nodename': basestring,}))
+    'nodename': basestring, }))
 def is_authorized(nodename):
     node = _must_find(model.Node, nodename)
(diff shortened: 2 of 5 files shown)


Izhmash:
If I understand correctly, this is checking authorization for the current user.  I think I would be helpful to mention in the docs that it's only for the current user credentials.

zenhack:
Also, I'm not sure this actually solves #692 anyway; as @Izhmash mentions, you can't check anyone else's auth with this. I'm not sure I understand the use cases then? Assuming the user knows what projects they're a part of, they could just do project_list_nodes.

naved001:
Ah, I was under the impression that we needed the API just to check the *caller's* authorisation,  not *any* user.

"Currently the way to check this would be to do something like calling show_node(), which is hacky." -- @henn from the [description here](https://github.com/CCI-MOC/hil/issues/692#issue-194513725)
This is the only thing that this aimed to solve, giving an explicit API rather than doing a show_node or project_list_nodes like you suggested.
 
In any case, if you guys think that this API is not needed and because it has the race condition, I am fine with closing this. (In fact, now that I think of it, we shouldn't make this API available just because of the race condition).

Want to hear your thoughts on this @henn

On a separate note, I read your (@zenhack) proposal and it seems alright to me at this point. Are you planning to finish it and make a PR anytime soon?

henn:
@pgrosu from the @CCI-MOC/bmi-breakers team and I had a long talk about this last week.

My suggestion was that BMI should only call HIL to connect itself to a user's network after the user gave the BMI service permission (via network ACLs).

Users should be able to create arbitrary nodes (which, to BMI, are just MAC addresses) and networks in BMI, since the only time that authorization would matter would be when asking BMI to manage booting on a particular network.

Others may not necessarily agree, though, so we need to have a big BMI design summit where a decision is made whether to pursue a highly-coupled HIL/BMI, or a looser model where users authorize/invite BMI and we primarily depend on keystone for auth checks to networks.

As to where this leaves us: let's talk to the BMI group and see which direction they want to head in.

zenhack:
I am highly antagonistic to adding anything on the HIL side that makes for tight coupling with anything else. I'm also pretty adamant about having conversations around this on the bug tracker. If you guys come together for a summit and then come to me with a pre-packaged answer, you're just going to have to waste a bunch of time hashing it out again with me. Don't do that.

I don't have much visibility into exactly what the BMI team is working on right now, but it sounds like the general problem is that we want projects to be able to delegate access to things to other services. I'd like to see if we can solve that problem in a general way, rather than just putting together a one of solution that works for BMI, another one that works for the console service, etc. Do you guys have something written up that I can read that will get me (somewhat) up to speed on where you're at/what you need from HIL?

henn:
@pgrosu / @CCI-MOC/bmi-breakers : can you guys comment on this?

naved001:
@apoorvemohan you want to comment on this one? Otherwise I might just close this soon.

PR #839, opened 2017-07-24
Bigint PK migration -- HIL core and all extensions

Izhmash:
- In response to #315
- Migrations for core HIL and all extensions
- Some migrations checked for table existence using `sa.MetaData(...)`, which caused hangs in the CI
    - Replaced with `db.inspect(...)`


hil/ext/auth/database.py
```diff
@@ -11,17 +11,26 @@
...
+from sqlalchemy.dialects import sqlite
+from os.path import join, dirname
+from hil.migrations import paths
 
 logger = ContextLogger(logging.getLogger(__name__), {})
 
+paths[__name__] = join(dirname(__file__), 'migrations', 'database')
+
+BigIntegerType = BigInteger().with_variant(
+        sqlite.INTEGER(), 'sqlite')
```
zenhack: Can we just import this from `hil.model`? Same goes for the other definitions.

Izhmash: Yeah, that makes more sense. I've changed them to fit this.

hil/ext/obm/ipmi.py
```diff
@@ -23,11 +23,21 @@
 from subprocess import call, Popen, PIPE
 import os
 
+from os.path import join, dirname
```
naved001: note: since you modified this file here and in #835; you should rebase this PR just to be safe.

hil/ext/obm/ipmi.py
```diff
@@ -23,11 +23,21 @@
 from subprocess import call, Popen, PIPE
 import os
```
naved001: note: since you modified this file here and in #835; you should rebase this PR just to be safe.

hil/ext/obm/migrations/mock/df8d9f423f2b_rename_mockobm_table_for_flask.py
```diff
@@ -8,19 +8,17 @@
...
 from alembic import op
-import sqlalchemy as sa
 from hil.model import db
-from hil.flaskapp import app
 
 # revision identifiers, used by Alembic.
 revision = 'df8d9f423f2b'
 down_revision = None
-branch_labels = ('hil.ext.obm.mock',)
+branch_labels = None
```
naved001: why is this `None` now?

Izhmash:
Just rebased this to master

naved001:
why are diffs from the other merged PRs shown in this one? can someone with a higher gitfu rank explain?

zenhack:
I feel like we've hit this before. @henn, do you recall what we figured out?

Izhmash:
Can we just squash them when we merge it?

hil/ext/obm/migrations/mock/df8d9f423f2b_rename_mockobm_table_for_flask.py
Izhmash: Only the newest migration on a branch should have that branch's label. Otherwise, it should say `None`.

Izhmash:
Also I'm working on adding a migration test too.

Izhmash:
I'm having some auth issues with the migration test, for now I've left the new database dump so it can be visually inspected.

zenhack:
There are some tweaks needed to the db dump, which (I think) are described in the migrations docs. Some are for documentation, others are workarounds which may solve your auth issue.

Izhmash:
I'll add the documentation parts.  I think I've deleted the necessary lines mentioned in the docs, my auth issue is just figuring out how to populated the 'fresh' db in the migration test.  I think I'll try something similar to `initial_db` in `auth/database.py`

Izhmash:
I'm thinking I should redo that database dump actually. How thorough should the test be? I was thinking about adding one instance of every object since everything was upgraded.

zenhack:
Quoting Ian Ballou (2017-07-26 14:29:34)

>    I'm thinking I should redo that database dump actually. How thorough
>    should the test be? I was thinking about adding one instance of every
>    object since everything was upgraded.

Sounds good.

Izhmash:
Pretty much done with the migration test. Only issue is with the User table, since the password hash is different every time a fresh database is created.  Thus, the migration always fails when comparing the user hashed password between the psql dump and the fresh DB. Does anyone know any workarounds?

zenhack:
Have a look at what I did for the headnode uuids

naved001:
@zenhack  want to look at the latest commit?

tests/unit/migrations.py
```diff
@@ -163,6 +164,9 @@ def create_bigint_db():
 
         db.session.commit()
 
+        db.session.query(User).get(1).hashed_password = 'pass'
```
zenhack: Rather than setting the hashed_password field to the cleartext password, can we just hard-code the hashed output (and add a comment re: what the password is)? Would be nice if we kept the database contents actually valid.

Izhmash:
Everything should be all set now.

tests/unit/migrations.py
```diff
@@ -164,7 +164,10 @@ def create_bigint_db():
 
         db.session.commit()
 
-        db.session.query(User).get(1).hashed_password = 'pass'
+        db.session.query(User).get(1).hashed_password = \
+            ('$6$rounds=656000$iTyrApYTUhMx4b4g$YcaMExV'
+             'YtS0ut2yXWrT64OggFpE4lLg12QsAuyMA3YKX6Czth'
+             'XeisA47dJZW9GwU2q2CTIVrsbpxAVT64Pih2/')
```
zenhack: Could you add a comment mentioning what the password actually is?

Izhmash: @zenhack my latest commit fixes this.

PR #843, opened 2017-07-31
Production docs update

Izhmash:
- Docs updated to remedy confusion from recent production installation.
- The line about dropping a user usually isn't relevant to a standard user.  If an inexperienced admin is following the guide closely, accidentally dropping the `HIL` user would cause issues.
- Postgres doesn't require all naming to be the same, it just helps setup defaults.
- Also included minor Keystone doc-related issues.

### DIFF
--- a/docs/INSTALL.rst
+++ b/docs/INSTALL.rst
@@ -146,6 +146,16 @@ To enable the database backend, make sure the **[extensions]** section of
 
   hil.ext.auth.database =
 
+Keystone Backend
+^^^^^^^^^^^^^^^^
+
+To enable the Keystone backend, make sure the **[extensions]** section of
+``hil.cfg`` contains::
+
+  hil.ext.auth.keystone =
+
+Visit the `Keystone configuration guide <http://hil.readthedocs.io/en/latest/keystone-auth.html>`_ for more information.
+
 Null Backend
 ^^^^^^^^^^^^
 
--- a/docs/INSTALL.rst
+++ b/docs/INSTALL.rst
@@ -154,7 +154,7 @@ To enable the Keystone backend, make sure the **[extensions]** section of
 
   hil.ext.auth.keystone =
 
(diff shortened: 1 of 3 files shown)


docs/Install_configure_PostgreSQL_CENTOS7.md
```diff
@@ -99,13 +99,9 @@ $ psql -c '\dg'
...
-If you wish to delete the user, do the following:
-
-```
-$ sudo -u postgres dropuser hil
-```
-**Note**: Make sure that the database role you create corresponds to an existing system user. 
-eg. There has to be a system user `hil` to access database named `hil` as database role named `hil`.
+**Note**: It is recommended that the PostgreSQL role and database you create correspond to an existing system user. 
+eg. There should be a system user `hil` to access database named `hil` as database role named `hil`.
+Advanced user/roll/database configurations may not need to follow this rule.  More information is available in the [Database Roles and Privileges](https://www.postgresql.org/docs/9.0/static/user-manag.html) reference guide.
```
naved001: user/role*/database

Izhmash: Oops, thanks

docs/INSTALL.rst
```diff
@@ -146,6 +146,16 @@ To enable the database backend, make sure the **[extensions]** section of
...
 
+Keystone Backend
+^^^^^^^^^^^^^^^^
+
+To enable the Keystone backend, make sure the **[extensions]** section of
+``hil.cfg`` contains::
+
+  hil.ext.auth.keystone =
+
+Visit the `Keystone configuration guide <http://hil.readthedocs.io/en/latest/keystone-auth.html>`_ for more information.
```
zenhack: Most of the rest of the docs just reference the file in `docs/`, seems a bit odd to me to link to the web for just this one.

Izhmash: Fixed this in my latest commit

PR #850, opened 2017-08-06
Juniper driver 

SahilTikale:
Submitting a switch driver to manage Juniper switches.
It based on PyEZ library made by Juniper that python wraps around netconf. 
It depends on jinja2 templating and yaml based customized tables to gather and change configuration information. 

Redhat provided the switch for a week, and it was a steep learning curve, understanding the interface and the above mentioned dependencies, therefore deployment tests are still pending.

Potentially the same driver should be able to manage many of different juniper switches but that remains to be tested. 

Creating this PR, so that feedback can start as early as possible.


docs/network-drivers.md
```diff
@@ -213,6 +215,50 @@ The body of the api call request will look like:
 The brocade driver accepts interface names the same way they would be accepted
 in the console of the switch, ex. ``101/0/10``.
 
+### Juniper QFX5200 driver
+
+#### Switch preparation
+
+1. The driver relies on netconf protocol to manage the switch. 
+It is important switch is configured for responding to netconf connections.
```
naved001: Phrasing here. "The switch should be configured for responding to netconf connections" or something like that would be better.

docs/network-drivers.md
```diff
@@ -213,6 +215,50 @@ The body of the api call request will look like:
...
+configuration mode should set it up for netconf based connections. 
+
+```
+set system services netconf ssh
+```
+
+2. Driver relies on the PyEZ library, a python wrapper that for netconf 
+using `ncc client`. Make sure it is installed before using the driver. 
+
+3. VLANS that HIL will manage should be preconfigured on switch.
```
naved001: Do you want to add a bit on how to do that like we do it for all switches?

docs/network-drivers.md
```diff
@@ -213,6 +215,50 @@ The body of the api call request will look like:
...
+```
+
+2. Driver relies on the PyEZ library, a python wrapper that for netconf 
+using `ncc client`. Make sure it is installed before using the driver. 
+
+3. VLANS that HIL will manage should be preconfigured on switch. 
+
+4. Configure the initial state of the interface (ports) configuration manually.
+Then copy the configuration as a jinja2 template and store it at
+`hil/ext/switches/junos/jinja_templates`. This will used by `revert_port`
```
naved001: will *be* used by

hil/cli.py
```diff
@@ -411,7 +411,7 @@ def node_register(node, subtype, *args):
         "ipmi", <hostname>, <ipmi-username>, <ipmi-password>
     """
     obm_api = "http://schema.massopencloud.org/haas/v0/obm/"
-    obm_types = ["ipmi", "mock"]
+    obm_types = ["ipmi", "mock", "netns"]
```
naved001: why is this here?

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,294 @@
...
+        `_commit_config`. 
+        """
+
+        jun = self.session()
+        try:
+            with jun:
+                self._set_load_config(jun, set_command)
+                self._commit_config(jun)
+        except ConnectError as err:
+                print("connat connect to device: {0}".format(err))
```
naved001: typo: cannot*

docs/network-drivers.md
```diff
@@ -213,6 +215,50 @@ The body of the api call request will look like:
...
+
+To register a Juniper QFX5200 switch, the ``"type"`` field of the request body
+must have a value of::
+
+	http://schema.massopencloud.org/haas/v0/switches/juniper
+
+In addition, it requires three extra fields: 
+``"username"``, ``"hostname"``, and ``"pasword"``, which provide the necessary
+information to connect to the switch via netconf over ssh using PyEZ library.
+``"hostname"`` can be an IP address too.
```
naved001: the hostname has to have `http://` in front of it right?

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,294 @@
...
+    def _interface_info(self, port):
+        """Fetches latest committed configuration information about `port`"""
+        jun = self.session()
+        try: 
+            with jun:
+                interface_config = InterfaceConfigTables(jun)
+                interface_config.get(interface= port, options = {'database':'committed'})
+                ic_dict = json.loads(interface_config.to_json())
+        except ConnectError as err:
+                print("connat connect to device: {0}".format(err))
```
naved001: typo: can not*

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,294 @@
...
+        """
+        if mode == "trunk":
+            set_mode = """
+            set interfaces {port} unit 0 family ethernet-switching interface-mode trunk
+        """.format(port=port)
+            self._set_commit_config(set_mode)
+        else: 
+            self._remove_all_vlans_port(port)
+
+        #x = "set interfaces et-0/0/08:0 unit 0 family ethernet-switching interface-mode trunk"
```
naved001: commented out code should be removed

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,294 @@
...
+    def _remove_vlan_from_port(self, port, vlan_id):
+        """ removes a single vlan specified by `vlan_id` """
+        remove_vlan = """
+        delete interfaces {port} unit 0 family ethernet-switching vlan members {vlan_id}
+        """.format(port=port, vlan_id=vlan_id)
+        self._set_commit_config(remove_vlan)
+        
+    def _remove_all_vlans_port(self, port):
+        """ Removes all vlans from the port, including the native vlan.
+        Also converts the interface to access mode.
```
naved001: 2 questions:

1. If this method can remove all vlans, then why do we need the other revert port? 
2. How/Why does removing all the VLANs puts it into access mode? Is this how the switch works?

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,294 @@
...
+
+        if state == "enable":
+            set_port_state = """
+            delete interface {port} disable
+            """.format(port=port)
+        elif state == "disable":
+            set_port_state = """
+            set interface {port} disable
+            """.format(port=port)
+        self._set_commit_config(remove_vlan)
```
naved001: what is `remove_vlan` here? shouldn't we do `self._set_commit_config(set_port_state)`

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,294 @@
...
+                jun.cfg.load(rm_curr_port_config, format="set")
+                jun.cfg.load( 
+                        template_path = base_config, template_vars=port_name,
+                        format='text'
+                        )
+
+                if jun.cfg.commit_check:
+                    jun.cfg.commit
+                else:
+                    jun.cfg.rollback
```
naved001: We can call `_commit_config` here in place of this if-else block

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,294 @@
...
+        dev.bind(cfg=Config)
+        return dev
+
+    def _commit_config(self, loaded_session):
+        """ Takes an open session that has loaded configuration change. 
+        Commits it if it is correct, otherwise rollback
+        to recent commited state.
+        `loaded_session`: session object after new configuration is loaded.
+        """
+        jun = loaded_session
```
naved001: why reassign it to some variable?

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,294 @@
...
+
+    def _set_native_vlan(self, port, network_id):
+       """Sets native vlan for a trunked port. """
+       set_native_vlan = """
+       set interface {port} native-vlan-id {network_id}
+        """.format(port=port, network_id=network_id)
+       self._set_commit_config(set_native_vlan)
+
+    def _remove_native_vlan(self, port, network_id):
+        """Removes native_vlan tag from port. """
```
naved001: in case of this switch, does it actually remove the native vlan or set it to some default?

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,294 @@
...
+        if channel == 'vlan/native':
+            if network_id is None:
+                self._remove_native_vlan(interface)
+            else:
+                self._set_native_vlan(interface, network_id)
+        else:
+            match = re.match(re.compile(r'vlan/(\d+)'), channel)
+            assert match is not None, "HIL passed an invalid channel to the" \
+                    " switch!"
+             #Discuss this with Kristi
```
naved001: What's this comment for? can you please elaborate?

naved001:
Just remembered this, does this switch save the running config to flash by default (like the brocade we have) so that changes aren't lost in case of power failures?

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,294 @@
...
+        """ Takes an open session that has loaded configuration change. 
+        Commits it if it is correct, otherwise rollback
+        to recent commited state.
+        `loaded_session`: session object after new configuration is loaded.
+        """
+        jun = loaded_session
+        if jun.cfg.commit_check:
+            jun.cfg.commit
+        else:
+            jun.cfg.rollback
```
zenhack: I assume you meant to call these functions?

naved001:
@SahilTikale any updates on this?

hil/ext/switches/juniper.py
```diff
@@ -103,192 +107,242 @@ def _jinja_load_config(self, o_session, j_template, var_dict):
...
                )
 
+    def _commit_config(self, loaded_session):
+        """ Takes an open session that has loaded configuration change.
+        Commits it if it is correct, otherwise rollback
+        to recent commited state.
+        `loaded_session`: session object after new configuration is loaded.
+        """
+        jun = loaded_session
+        if jun.cfg.commit_check:
```
naved001: `commit_check()`?

hil/ext/switches/juniper.py
```diff
@@ -103,192 +107,242 @@ def _jinja_load_config(self, o_session, j_template, var_dict):
...
 
-        return ic_dict
+    def modify_port(self, port, channel, network_id):
+        """ Changes vlan assignment to the port. 
+        `node_connect_network` with 'vlan/native' flag: 
+            enable port; set port to trunk mode; assign native_vlan; 
+            remove default vlan
+        """
+
+        import pdb; pdb.set_trace()
```
naved001: i guess you forgot to take this out.

hil/ext/switches/juniper.py
```diff
@@ -103,192 +107,242 @@ def _jinja_load_config(self, o_session, j_template, var_dict):
...
+                assert network_id == vlan_id
+#                import pdb; pdb.set_trace()
+                try:
+                    self._add_vlan_to_trunk(interface, vlan_id)
+                except VlanAddError as e:
+                    return e
+
+    def _set_native_vlan(self, port, network_id):
+       """Sets native vlan for a trunked port. """
+#       import pdb; pdb.set_trace()
```
naved001: commented out debugging statement should be removed.

hil/ext/switches/juniper.py
```diff
@@ -103,192 +107,242 @@ def _jinja_load_config(self, o_session, j_template, var_dict):
...
+       set_native_vlan = """
+       {port_set} native-vlan-id {network_id}
+       {port_set} unit 0 family ethernet-switching vlan members {network_id}
+       """.format(port_set=port_set, network_id=network_id)
+       
+       remove_default_vlan = """
+       {port_del} unit 0 family ethernet-switching vlan members default
+       """.format(port_del=port_del)
+       
+       if port_info[port]['trunk_mode']:
```
naved001: can we do a `if not` here and put the else statement in this block?

hil/ext/switches/juniper.py
```diff
@@ -103,192 +107,242 @@ def _jinja_load_config(self, o_session, j_template, var_dict):
...
+    def _interface_info(self, port):
+        """Fetches latest committed configuration information about `port`"""
+        jun = self._create_session()
+        try:
+            with jun:
+                interface_config = InterfaceConfigTable(jun)
+                interface_config.get(interface= port, options = {'database':'committed'})
+                ic_dict = json.loads(interface_config.to_json())
+        except ConnectError as err:
+                print("connat connect to device: {0}".format(err))
```
naved001: typo: connat

hil/ext/switches/juniper.py
```diff
@@ -103,192 +107,242 @@ def _jinja_load_config(self, o_session, j_template, var_dict):
...
-        except Exception as err:
-            print (err)
+        self._set_commit_config(add_vlan)
 
-    def modify_port(self, port, channel, network_id):
-        """ Changes vlan assignment to the port.
-        """
+    def _remove_vlan_from_port(self, port, vlan_id):
+        """ removes a single vlan specified by `vlan_id` """
+        import pdb; pdb.set_trace()
```
naved001: debugging statement

docs/network-drivers.md
SahilTikale: addressed in the next commit.

docs/network-drivers.md
SahilTikale: example added in next commit. done

docs/network-drivers.md
SahilTikale: thanks for catching that. fixed it.

naved001:
@SahilTikale Let me know when this is ready to be reviewed.

docs/network-drivers.md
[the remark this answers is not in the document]
SahilTikale: No it does not have to be a url. In fact it should not be, since pyEZ makes ssh connections. It does not make any rest calls. `hostname` should be either IP address or a hostname or FQDN.

docs/network-drivers.md
```diff
@@ -213,6 +215,50 @@ The body of the api call request will look like:
...
+``"username"``, ``"hostname"``, and ``"pasword"``, which provide the necessary
+information to connect to the switch via netconf over ssh using PyEZ library.
+``"hostname"`` can be an IP address too. 
+The driver can potentially work with wide range of JUNOS based switches 
+without any changes but it has not been tested yet. 
+
+####
+
+Port naems must be of the same form accepted by the switch's cli interface,
+e.g. ``et-0/0/0`` is a physical port. A 40Gig port when distributed as a 4
```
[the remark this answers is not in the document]
SahilTikale: fixed it.

hil/cli.py
SahilTikale: working on multiple branches. will remove it in next commit. done.

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,294 @@
...
+from jnpr.junos import Device
+from jnpr.junos.exception import ConnectError
+from jnpr.junos.utils.config import Config
+
+
+from hil.model import db, Switch
+from hil.migrations import paths
+from os.path import dirname, join
+from hil.migrations import paths
+from hil.ext.switches.junos.config_tables.ConfigTables import InterfaceConfigTable
```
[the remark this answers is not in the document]
SahilTikale: it is used in function _interface_info(self, port)

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,294 @@
...
+logger = logging.getLogger(__name__)
+paths[__name__] = join(dirname(__file__), 'migrations', 'juniper')
```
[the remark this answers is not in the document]
SahilTikale: it will be required if you wish to add Juniper switch to a running HIL installation.

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,294 @@
...
+
+
+class juniper(Switch):
+    api_name = 'http://schema.massopencloud.org/haas/v0/switches/juniper'
+
+    __mapper_args__ = {
+        'polymorphic_identity': api_name,
+    }
+
+    id = db.Column(db.Integer, db.ForeignKey('switch.id'), primary_key=True)
```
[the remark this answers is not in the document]
SahilTikale: fixed it. Done

hil/ext/switches/juniper.py
SahilTikale: o as to keep the rest of the code uniform. the session name in each function is jun but depending on the context, it might be a session that is not yet opened; or an open session or an opened session that is loaded with configuration changes.

hil/ext/switches/juniper.py
SahilTikale: yup. thanks for catching that. fixed it.

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,294 @@
...
+        `o_session`: session object with connection to switch already open.
+        `set_command`: Multi-line `set` method based command string.
+        """
+        jun = o_session
+        jun.cfg.load(set_command, format="set")
+
+    def _jinja_load_config(self, o_session, j_template, var_dict):
+        """ Load configuration changes using the 'jinja_template' method.
+        `o_session`: session object with connection to switch already open.
+       `j_template`: jinja2 compliant configuration template. Samples is
```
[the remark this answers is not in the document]
SahilTikale: fixed it.

hil/ext/switches/juniper.py
SahilTikale: fixed it.

hil/ext/switches/juniper.py
SahilTikale: fixed it.

hil/ext/switches/juniper.py
SahilTikale: It is not commented out code. It is left as an example for future maintainers.

hil/ext/switches/juniper.py
SahilTikale: The function is changed. review it again and let me know if you have any questions.

hil/ext/switches/juniper.py
SahilTikale: Looks like I will not be needing this function after all. Removed it.

hil/ext/switches/juniper.py
SahilTikale: this function is removed from the driver.

hil/ext/switches/juniper.py
SahilTikale: I agree. replaced it as suggested.

hil/ext/switches/juniper.py
SahilTikale: it was a `note to myself`. removed it.

SahilTikale:
The code is ready for review. I have merge the changes from  the master.
Only thing remaining is the deployment test and how to import the pyez library on demand.

docs/network-drivers.md
```diff
@@ -213,6 +215,57 @@ The body of the api call request will look like:
...
+```
+set vlans hil-500 vlan-id 500
+``` 
+
+4. Configure the initial state of the interface (ports) configuration manually.
+Then copy the configuration as a jinja2 template and store it at
+`hil/ext/switches/junos/jinja_templates`. This will by used by `revert_port` 
+to reset the ports to its initial conditions when required. 
+
+#### switch_resister
```
naved001: typo: re**g**ister

docs/network-drivers.md
```diff
@@ -213,6 +215,57 @@ The body of the api call request will look like:
...
+In addition, it requires three extra fields: 
+``"username"``, ``"hostname"``, and ``"pasword"``, which provide the necessary
+information to connect to the switch via netconf over ssh using PyEZ library.
+``"hostname"`` can be an IP address too. 
+The driver can potentially work with wide range of JUNOS based switches 
+without any changes but it has not been tested yet. 
+
+####
+
+Port naems must be of the same form accepted by the switch's cli interface,
```
naved001: port n**am**es

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,373 @@
...
+from hil.migrations import paths
+from hil.ext.switches.junos.config_tables.ConfigTables import (
+        InterfaceConfigTable
+        )
+logger = logging.getLogger(__name__)
+paths[__name__] = join(dirname(__file__), 'migrations', 'juniper')
+
+
+class VlanAddError(Exception):
+    pass
```
naved001: I'd recommend having a docstring everywhere (see #861 )

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,373 @@
...
+
+
+class ConfigCommitError(Exception):
+    pass
```
naved001: same; a docstring would be nice.

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,373 @@
...
+            'password': basestring,
+        }).validate(kwargs)
+
+    def session(self):
+        return self
+
+    def disconnect(self):
+        pass
+
+    def _create_session(self):
```
naved001: why did you make a different session object here instead of at line 87 `session`?

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,373 @@
...
+        """ Load configuration changes using the 'set' method.
+        `o_session`: session object with connection to switch already open.
+        `set_command`: Multi-line `set` method based command string.
+        """
+        jun = o_session
+        try:
+            jun.cfg.load(set_command, format="set")
+        except ConfigLoadError as e:
+            if e.message == "warning: statement not found":
+                pass
```
naved001: sorry but it's' not clear what's happening here. can you add a comment here?

 If there's an error message then pass and then what? do you want to write it to logs or something?

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,373 @@
...
+            with jun:
+                self._set_load_config(jun, set_command)
+                try:
+                    self._commit_config(jun)
+                except ConfigCommitError as err:
+                    message = (
+                            "Inconsistent configuration, \
+                                    cannot commit. Exiting.".format(err)
+                                    )
+                    return sys.exit(1)
```
naved001: return sys.exit(1)? Why are you returning this? Can't you just exit here?

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,373 @@
...
+        except ConnectError as err:
+            message = "cannot connect to device: {0}".format(err)
+            return message
+
+    def revert_port(self, port):
+        """Resets the port to the factory default.
+        Running this function will replace the port settings
+        with the ones stored in the jinja template `jinja_templates/`
+        It reset any port to default port config as defined in
```
naved001: It reset**s**

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,373 @@
...
+                    return e
+
+    def _set_native_vlan(self, port, network_id):
+        """Sets native vlan for a trunked port.
+        It enables the port, if it is the first vlan for the port.
+        """
+        port_info = self._interface_info(port)
+        self._set_commands(port, network_id)
+        set_native_vlan = self.set_native_vlan
+        if port_info[port]['trunk_mode']:
```
naved001: `if not` and you can get rid of the else block

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,373 @@
...
+            add_vlan = self.enable_interface + add_vlan
+
+        if 'default' in port_info[port]['vlans']:
+            add_vlan = add_vlan + self.remove_default_vlan
+        self._set_commit_config(add_vlan)
+        self._clear_set_commands()
+
+    def _remove_vlan_from_port(self, port, vlan_id):
+        """ removes a single vlan specified by `vlan_id` """
+#        import pdb; pdb.set_trace()
```
naved001: should be removed

hil/ext/switches/juniper.py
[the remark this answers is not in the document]
naved001: in that case you should generate a migration script too with this PR. 

also, as far as I know, running `hil-admin db create` creates any new tables so a migration script might not be needed. @zenhack can you confirm this?

docs/network-drivers.md
SahilTikale: fixed it.

docs/network-drivers.md
SahilTikale: done.

hil/ext/switches/juniper.py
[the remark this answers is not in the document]
SahilTikale: I tested it, if a new driver is introduced to a running HIL installation it does not require a migration script. To add the new table to the database, stop the HIL servers and simple run `hil-admin db create` again. It will create any new tables based on the changes to the `hil.cfg`

docs/network-drivers.md
```diff
@@ -213,6 +215,57 @@ The body of the api call request will look like:
...
+in configuration mode run:
+
+```
+set vlans hil-500 vlan-id 500
+``` 
+
+4. Configure the initial state of the interface (ports) configuration manually.
+Then copy the configuration as a jinja2 template and store it at
+`hil/ext/switches/junos/jinja_templates`. This will by used by `revert_port` 
+to reset the ports to its initial conditions when required.
```
zenhack: This is not a reasonable interface. There are a few problems:

1. In general, users should not be storing things inside the HIL source tree. Remember, HIL may actually be installed somewhere, and not using the source tree directly, and the equivalent thing would be asking them to shove a file somewhere in `/usr/lib/python2.7/site-packages`, which is... horrible.
2. Sysadmins should not be asked to mess with templating languages, especially as part of routine set-up.

Can you point to some documentation explaining how all this works? There has to be a simpler way to manage this.

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,404 @@
...
+
+class VlanAddError(Exception):
+    pass
+
+
+class ConfigCommitError(Exception):
+    pass
+
+
+class Juniper(Switch):
```
zenhack: I don't even know where to begin with this class. It is an unreadable, umaintainable mess. I seriously can't follow it. A few concrete suggestions:

* Get rid of the global variables (enable_interface, add_vlan, etc). There's almost never a good reason for a mutable global variable in all but the tiniest programs.

* Rather than constructing the commands as strings, then passing them around untill they're eventually used, encapsulate the functionality you need in methods/functions -- the nexus driver provides a number of good examples of what this looks like:

```python
    def disable_vlan(self, vlan_id):
        self.console.sendline('sw trunk allowed vlan remove %s' % vlan_id)
```

Then just call the functions where you need them.

* Anywhere where you've got something like:

```python
try:
    with ...:
        try:
            ...
...
```

...or similar is a bad sign; that much non-local control flow is almost guaranteed to be unreadable. Simplify that sort of thing.

* You've got some kind of session related to the library that you're passing around, but the session() method just returns self -- might it make sense to actually return something else, and store some of the state there? I really can't follow this, so it's hard for me to be more specific.

SahilTikale:
```
============================================= test session starts =============================================
platform linux2 -- Python 2.7.13, pytest-3.1.3, py-1.4.34, pluggy-0.4.0
rootdir: /home/jack/git/hil, inifile: setup.cfg
plugins: xdist-1.18.0, cov-2.5.1, catchlog-1.2.2
collected 1 item s

========================================== 1 passed in 77.41 seconds ==========================================
(.venv) [jack@quadshil hil]$ py.test tests/deployment/
headnodes.py        native_networks.py  switch_config.py
ipmi.py             __pycache__/        vlan_networks.py
(.venv) [jack@quadshil hil]$ py.test tests/deployment/native_networks.py
============================================= test session starts 
=============================================
platform linux2 -- Python 2.7.13, pytest-3.1.3, py-1.4.34, pluggy-0.4.0
rootdir: /home/jack/git/hil, inifile: setup.cfg
plugins: xdist-1.18.0, cov-2.5.1, catchlog-1.2.2
collected 1 item s

========================================== 1 passed in 66.26 seconds ==========================================
(.venv) [jack@quadshil hil]$ git branch
* juniper_driver
```

docs/network-drivers.md
```diff
@@ -81,6 +81,8 @@ HIL:
   Nexus switches. Only the 3500 and 5500 have been tested, though it is
   possible that other models will work as well.
 * ``hil.ext.switches.brocade``, for the brocade VDX 6740.
+* ``hil.ext.switches.juniper.juniper``, for the juniper qfx 5200, 
+but potentially can be used to manage any junos based switch.
```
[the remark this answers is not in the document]
SahilTikale: done.

docs/network-drivers.md
SahilTikale: This is part of how Juniper uses the pyEZ library.  It is quite simple to use. I have provided the comments in the templates itself. I am assuming that the one who is setting up the initial switch configuration has a good idea of what a default port state should look like. They have to mimic that state in the jinja template. It is not complex at all.  A simpler way is I try to pass the template as a string or json or dictionary. In any case the sysadmin who will have to make some changes to the driver to decide what a default port state should look like. I thought this was cleaner and readable way of doing things. 
here is the link: https://www.juniper.net/documentation/en_US/junos-pyez/topics/task/program/junos-pyez-program-configuration-data-loading.htmlhttps://www.juniper.net/documentation/en_US/junos-pyez/topics/task/program/junos-pyez-program-configuration-data-loading.html

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,404 @@
...
+import os
+
+from jnpr.junos import Device
+from jnpr.junos.exception import ConnectError
+from jnpr.junos.utils.config import Config
+from jnpr.junos.utils.config import ConfigLoadError
+
+
+from hil.model import db, Switch
+from hil.model import BigIntegerType
```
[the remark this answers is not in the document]
SahilTikale: done.

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,404 @@
...
+from hil.migrations import paths
+from os.path import dirname, join
+from hil.migrations import paths
```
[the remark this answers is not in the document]
SahilTikale: removed.

hil/ext/switches/juniper.py
SahilTikale: sure. Added it.

hil/ext/switches/juniper.py
SahilTikale: done.

hil/ext/switches/juniper.py
SahilTikale: Here is the context, after getting the initial driver working. I took a second look at it to see if I can remove repetitive chunks of the code. I realized that there were some common set of commands that I was using to do most of the port related changes in the switch. 
Also, the some commands were long and had to be on a single line but I also had to stick to the 79 character limit. 

Considering this factors I did three things.
1. defined bunch of variables that other functions can use and initialized them to be None.
Their names reflect the commands that they will contain:
```
    enable_interface = None
    add_vlan = None
    remove_vlan = None
    set_native_vlan = None
    remove_native_vlan = None
    remove_default_vlan = None
    set_trunk_mode = None
```
2. defined a method that will initialized them for a given port and vlan
```
    def _set_commands(self, port, vlan_id):
        """ Set of common `set` commands used by other functions."""
```
3. and the wrote another function to clear the state at the end of each use.

I initially was reluctant to use global variables but did not find a better way to implement this. I will look at the nexus driver and see if I can improve this.

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,404 @@
...
+        }).validate(kwargs)
+
+    def session(self):
+        return self
+
+    def disconnect(self):
+        pass
+
+    @staticmethod
+    def validate_port_name(port):
```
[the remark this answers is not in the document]
SahilTikale: yup. after other comments are resolved.

hil/ext/switches/juniper.py
SahilTikale: if I create the session object, I have to put all the other function under the session method and that was too much indentation for me. 
To get some more horizontal space I reduced a level of identation by not using the `session` object. Instead made a private method.

Also, it only creates the object. Opening, loading, closing the session is done whenever a method requires the session.

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,404 @@
...
+        dev.bind(cfg=Config)
+        return dev
+
+    def _set_commands(self, port, vlan_id):
+        """ Set of common `set` commands used by other functions."""
+        port_set = "set interfaces {port}".format(port=port)
+        port_del = "delete interfaces {port}".format(port=port)
+
+        self.enable_interface = """
+        {port_del} disable""".format(port_del=port_del)
```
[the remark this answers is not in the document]
SahilTikale: Interesting you should say that. I found the using `\n` more distracting. 
This was cleaner in my opinion. Everything between the """ ... """ is statement string. Each new statement starts on a new line.

hil/ext/switches/juniper.py
SahilTikale: ok, I will add an comment there. sometimes the element on which the action is to be taken might not be present there. In that case this warning is  raised. It is harmless. To bring some sense of idempotency to the operations I implemented this. eg. to enable a port you have to delete the `disable` tag, if you do it again there is no `disable` tag for deletion, then the session will raise this warning.

hil/ext/switches/juniper.py
SahilTikale: my bad. fixed it.

hil/ext/switches/juniper.py
SahilTikale: done.

hil/ext/switches/juniper.py
SahilTikale: done.

hil/ext/switches/juniper.py
SahilTikale: it is not in my file. I guess some update issue. I will commit again.

SahilTikale:
I have address most of the comments by the @naved001 and @zenhack 
Will work on removing the global variables and the function `validate_port_name`.
Will also, rearrange the private functions such that the class is more readable.

hil/ext/switches/juniper.py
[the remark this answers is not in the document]
zenhack: It just looks odd visually; my brain wants to parse the second line as its own statement. Perhaps not worth worrying too much about though since I've asked you to remove these variables anyway.

hil/ext/switches/juniper.py
zenhack: If you've got a docstring, you don't need the `pass` statement.

hil/ext/switches/juniper.py
zenhack: same thing re: `pass`

hil/ext/switches/junos/config_tables/ConfigTables.py
```diff
@@ -3,5 +3,5 @@
 _YAML_ = splitext(__file__)[0] + '.yml'
 globals().update(loadyaml(_YAML_))
 
-# Reference: "https://www.juniper.net/documentation/en_US/junos-pyez/topics/task/program/junos-pyez-tables-views-external-importing.html"
-
+# Reference: "https://www.juniper.net/documentation/en_US/junos-pyez/topics/
+# task/program/junos-pyez-tables-views-external-importing.html"
```
zenhack: I don't think splitting this across two lines is a good idea. I assume you did this to satisfy pep8? I think this is a good place to just use `# noqa`

zenhack:
I made a few more minor comments. I'll have another look once you've reworked that class.

docs/network-drivers.md
zenhack: Ok, I'll read up on the documentation and weigh in again. I am regardless not open to storing the template (or any user-supplied data) in the source tree, but that can maybe go in the database or something, if it's the right way to go. I might be amenable if this is really the way juniper expects you to manage the switch normally.

hil/ext/switches/juniper.py
zenhack: For the long-line problem, you can do something like this:

```python
"foo bar baz quux " \
    "apple bannana foo " \
    "more stuff"
```

Python will concatenate string literals that are next to one another.

Like I said, I think the way to go is to just have a bunch of methods, rather than string variables, which take parameters (and the string construction happens in the method).

hil/ext/switches/juniper.py
zenhack: You can solve the indentation problem by declaring a wrapper class elsewhere in the file, and here just return that.

docs/network-drivers.md
naved001: For port revert there are 2 files and those contain some string and comments. Why can't we just put that string in the driver (the python file) itself?

hil/ext/switches/junos/config_tables/ConfigTables.py
SahilTikale: yes I did it for satisfying pep8. Thanks for the suggestion. will do that in the next commit.

docs/network-drivers.md
SahilTikale: I had to learn using `jinja` templates because pyez library relies on it, although there are other ways of getting the same results. It is simpler to make changes using templates which otherwise would require multiple `set` and `delete` statements to be executed on the switch. @naved001 I kept it in a separate file so as to hide the complexity from the user. The user does not need to go through the driver code to figure out how to set initial port of the driver. @zenhack Juniper has chosen `jinja templating` because, it can do lot of complex things with it. I provide a simpler template but complex templating including`for loops` and other constructs can be used to create a desired state. That is another reason I decided to make this its own file. That way it keeps things readable in their own files.

naved001: We should let the admin set the path for the templates in the config file (under a section for the juniper siwtch `[hil.ext.switch.juniper]`) and the driver then reads the path from there and loads them jinja templates. @henn thought this was a reasonable solution. What do you think @zenhack ?

zenhack: Having picked through the documentation a bit, it's still not clear to me why we need to have the user supply a template at all -- the operations we perform are pretty cookie-cutter, I would think we would just be able to write a template (or a few of them) ahead of time, and pass data to it at run-time.

zenhack:
@SahilTikale, ping.

It's been a month and a half. I'm tempted to suggest we just start closing PRs if the submitter is unresponsive, partially to keep things from cluttering up the tracker, but also to try to encourage shortening the pipeline.

naved001:
@SahilTikale could you post some updates about this?

SahilTikale:
I agree with @zenhack. 
But last month has been crazy, with two course this semester, MOC day preparation, and 3 demos that I had to prepare for RedHat (2 demo to perf. Scale team and 1 for open shit team on behalf of MOC) pertaining to the work I am doing for integrating HIL with QUADS, I was a little tight on time. 

Also, making the changes that were suggested broke the deployment tests and testing requires VPN access to the Juniper switch. I have talked to my manager at Red Hat about getting the access. Once I fix the code and pass the deployment tests, will update the PR.

Meanwhile, I wrote the driver for Openvswitch (gave a demo of it at RedHat).
Will soon make a PR of it.

docs/network-drivers.md
```diff
@@ -1,258 +1,319 @@
-# Network Drivers
+	# Network Drivers
```
naved001: Looks like you have put everything in a code block. which is why everything is indented and the whole file is shown in diff.

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,434 @@
...
+# IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied.  See the License for the specific language
+# governing permissions and limitations under the License.
+"""A switch driver for the juniper QFX series.
+
+But in theory can work for any Juniper switch that supports NETCONF.
+Driver uses the python library PyEZ made available by Juniper.
+By default it sets NETCONF session over ssh, i.e. make RPC calls over ssh.
+For this to work, make sure that the Juniper switch is configured
+appropriately to recieve NETCONF connections. Its default port is 830.
```
naved001: Since this is an important note for administrators, this should go in the documents (`network-drivers.md`).

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,434 @@
...
+import logging
+import schema
+import json
+import sys
+import os
+
+from jnpr.junos import Device
+from jnpr.junos.exception import ConnectError
+from jnpr.junos.utils.config import Config
+from jnpr.junos.utils.config import ConfigLoadError
```
naved001: these 2 imports can be collapsed into 1.

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,434 @@
...
+
+
+from hil.model import db, Switch, BigIntegerType
+from hil.migrations import paths
+from hil import config
+from hil.config import cfg
+from os.path import dirname, join
+
+import pdb; pdb.set_trace()
```
naved001: debugging statement

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,434 @@
...
+config.load()
```
naved001: I think config file is already loaded when hil is running ([Example](https://github.com/CCI-MOC/hil/blob/master/hil/ext/switches/_console.py#L123)). Did you add this for debugging?

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,434 @@
...
+j_driver = 'hil.ext.switches.juniper'
+if cfg.has_option('extensions', j_driver):
+    template_dir = cfg.get('extensions', j_driver)
+    sys.path.append(template_dir)
+    try:
+        from config_tables.ConfigTables import InterfaceConfigTable
+    except ImportError:
+        pass
```
naved001: sorry, but I don't understand what's going on here with this try-catch block? if there's something you are trying to import and it fails, you just silently pass?

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,434 @@
...
+
+logger = logging.getLogger(__name__)
+paths[__name__] = join(dirname(__file__), 'migrations', 'juniper')
```
naved001: Not needed https://github.com/CCI-MOC/hil/pull/897#discussion_r144885110

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,434 @@
...
+
+
+class VlanAddError(Exception):
+    """ Raise this exception when vlan cannot be added to the port."""
+    pass
```
naved001: `pass` can be removed since you have a docstring

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,434 @@
...
+
+
+class ConfigCommitError(Exception):
+    """ Raise this exception if there are commit conflicts."""
+    pass
+
+
+class Juniper(Switch):
```
naved001: inherit from both `Switch` and `SwitchSession`. [Example](https://github.com/CCI-MOC/hil/blob/master/hil/ext/switches/dellnos9.py#L38).

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,434 @@
...
+        return self
+
+    def disconnect(self):
+        pass
+
+    @staticmethod
+    def validate_port_name(port):
+        pass
+
+    def _create_session(self):
```
naved001: Looks like the switch is connection oriented. In that case your driver should look like how it is defined in `hil/model.py` (or like the nexus driver). This is important because the Networking Daemon caches the session object which avoids connecting-disconnecting for each change.

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,434 @@
...
+        self._set_commands(port, vlan_id)
+        remove_native_vlan = self.remove_native_vlan
+        if isinstance(port_info[port]['vlans'], (str, unicode)):
+            self.revert_port(port)
+        else:
+            self._set_commit_config(remove_native_vlan)
+        self._clear_set_commands()
+
+    def _get_mode(self, port):
+        """Returns True if the port is in trunk mode else returns False. """
```
naved001: the name implies that it will return either trunk or access, not some boolean. Should choose a more appropriate name, something like `is_mode_trunk()`

hil/ext/switches/juniper.py
```diff
@@ -0,0 +1,434 @@
...
+        """Fetches latest committed configuration information about `port`"""
+        jun = self._create_session()
+        try:
+            with jun:
+                interface_config = InterfaceConfigTable(jun)
+                interface_config.get(
+                        interface=port, options={'database': 'committed'}
+                        )
+                ic_dict = json.loads(interface_config.to_json())
+        except ConnectError as err:
```
naved001: I think errors should be logged for now so that admins can read those.

scripts/junos/jinja_templates/revert_nonchannelized_port.j2
```diff
@@ -0,0 +1,16 @@
+{# variable 'port_name' should be like 'et-0/0/11' #}
+{# or 'et-0/0/12' or 'et-0/0/8' #}
+{# Also make sure that dictionary passed to template_vars #}
+{# contains keys that matches the variable names that are used here #}
+interfaces {
+            {{ port_name }} {
+    unit 0 {
+        family ethernet-switching {
+            vlan {
+                members default;
```
naved001: what's the default vlan membership?

scripts/junos/jinja_templates/revert_nonchannelized_port.j2
```diff
@@ -0,0 +1,16 @@
...
+            }
+            storm-control default;
```
naved001: why are you setting storm-control only on a non-channelized port, why not in a channelized port?

naved001: these are the only 2 differences in the .j2 files for channelized and non channelized ports. 

Also, can you add a description to each `.j2` file what it's doing and why it's doing? because it's not obvious to someone like me who doesn't know much about jinja templates.

scripts/junos/config_tables/ConfigTables_old.yml
```diff
@@ -0,0 +1,24 @@
+# To understand the following definitions, refer to
```
naved001: I guess you wanted to get rid of this file since you don't use it anywhere in the code and have an another almost exactly same file `ConfigTables.yml`

scripts/junos/config_tables/InterfaceConfigTables.yml
```diff
@@ -0,0 +1,24 @@
+# To understand the following definitions, refer to
```
naved001: Please add a description on top of these `.yml` files so that it's easy to follow why it's there in the first place. 

This looks very similar to `ConfigTables.yml`. And idk where you are using this.

hil/ext/switches/juniper.py
naved001: see this comment [here](https://github.com/CCI-MOC/hil/pull/850#discussion_r144900753)

If that's the case, then the Juniper class will inherit from Switch like you did. And then there will be a class for Session which inherits from SwitchSession.

SahilTikale:
I have just addressed the issue of strings and global variables. 
I will fix the issue of putting junos in `/etc/hil`. that is proving to be challenging since it is outside the module tree of hil. 

Once this issue is resolved I will address all the other suggestions by reviewers.

zenhack:
@SahilTikale, this is much better. A few more general things I'd like to see improved:

* The details of splicing the strings together is now hidden behind functions, but when reading the code that makes the logical changes, I still have to look at code that's gluing the commands together. It would be more readable if it just looked like standard procedural code. For one thing, the whole thing is backwards; the commands get run in the opposite order of what's in the python source, because you're gluing the new commands onto the beginning. Just gluing them on to the end (and adjusting the code to still consturct the same set of commands) would make this much more readable. But I think you can do a bit better. Rather than gluing strings together explicitly, change your `_str_*` methods to accept an extra argument (probably at the front) that's a StringIO; then you can just write the string to it, rather than returning it and having the caller need to deal with it. e.g `_set_native_vlan` might look like:

```python
buf = StringIO()
if 'default' in port_info[port]['vlans']:
    self._str_remove_default_vlan(buf, port)
if port_info[port]['disabled']:
    self._str_enable_interface(buf, port)
if not port_info[port]['trunk_mode']:
    self._str_set_trunk_mode(buf, port)
self._str_set_native_vlan(buf, port, network_id)
self._set_commit_config(buf.getvalue())
```

You might also encapsulate the commands a bit further, e.g.

```python
cmd = self.make_command_builder()
if 'default' in port_info[port]['vlans']:
    cmd.remove_default_vlan(port)
if port_info[port]['disabled']:
    cmd.enable_interface(port)
if not port_info[port]['trunk_mode']:
    cmd.set_trunk_mode(port)
cmd.set_native_vlan(port, network_id)
cmd.run()
```

Where `self.make_command_builder()` returns an instance of some class that has these methods, and internally builds up the state.

This way the reader can just focus on the logic of what changes are being made to the switch, without worrying about how they're formatted/put on the wire at the same time.

* There are a whole bunch of places (like the above) where you've got this `port_info` dict, and only ever access things in `port_info[port]`; it would be lessy noisy if you made `_interface_info()` just return `port_info[port]` (is there anywhere you're using it where this isn't what you want?); then you could do:

```python
cmd = self.make_command_builder()
if 'default' in port_info['vlans']:
    cmd.remove_default_vlan(port)
if port_info['disabled']:
    cmd.enable_interface(port)
if not port_info['trunk_mode']:
    cmd.set_trunk_mode(port)
cmd.set_native_vlan(port, network_id)
cmd.run()
```

* You've got a bunch of functions where you're passing in the juniper session as one variable name, e.g. `o_session`, and then immediately assigning it to a new variable `jun`, which you use internally. Just pick a sensible name for the parameter and use it directly.

Take care of that stuff and I'll do an in-depth pass.

scripts/junos/jinja_templates/revert_nonchannelized_port.j2
SahilTikale: The settings in the file are what the administrators at Redhat wanted for their switch.
That was taken `as is`. Also that is the idea that they can set `any default state` for ports as they wish and HIL will comply with the requirements without having to know the reasons to do so. There can be settings that may affect the switch as a whole but it may have no effect on the operations that HIL administers.

SahilTikale:
@zenhack about string mangling, I will have to read your comments again and try out the options you suggested. I will update once I get back to doing that. 

I am still struggling with moving the config table into `/etc/hil` since it outside the domain of the hil module, importing it is turning out to be a challenge. Will update once I have a version that works as expected.

zenhack:
@SahilTikale, not sure I understand what you're trying to do with `/etc/hil`?

naved001:
@SahilTikale Are you talking about importing the yml files  [here](https://github.com/CCI-MOC/hil/pull/850/files#diff-71bf847bcd26698b7d1389d0063b4a59R40) ?

I don't know how that import worked in your code, but I found this way of parsing yml files.
```
>>> import yaml
>>> with open("/etc/junos/ConfigTables.yml", 'r') as stream:
...     print(yaml.load(stream))
... 
{'InterfaceConfigTable': {'get': 'interfaces/interface', 'view': 'InterfaceConfigView', 'required_keys': 'interface:name'}, 'InterfaceConfigView': None}
```

Other than that, you could import python modules by inserting into sys.path
```
>>> import sys
>>> sys.path.insert(0, '/etc/junos')
```
But this doesn't "import" the yml files for me, I thought you could only import python modules?

naved001:
@SahilTikale any updates on any of your PRs?

naved001:
I suggest we close this PR, @SahilTikale you can start a new one from your branch when you are ready. Maybe something is wrong with github, but PRs with lot of comments fail to load most of the times.

naved001:
@SahilTikale will start a fresh PR.

PR #851, opened 2017-08-08
Use setuptools extras to track optional dependencies

zenhack:
Per the first commit message, see:

https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies

This patch uses the extras feature to specify dependencies for our optional features: postgres & the keystone stuff. I've also split the test dependencies into an extra, so we can install those only during test -- no need to have pytest and several plugins installed on a production system.

This had floated through my mind a few times, and then @SahilTikale asked about what to do with the juniper driver, so I figured it was time to do it.


setup.py
```diff
@@ -95,12 +95,19 @@ def _get_readme():
...
+                'pytest-catchlog>=1.2.2,<2.0',
+                'pytest-cov>2.0,<3.0',
+                'pytest-xdist>=1.14,<2.0',
+                'pep8>=1.7.0',
+                'pylint>=1.6.0,<2.0',
+                'requests_mock>=1.0.0,<2.0',
+          ],
+          'postgres': ['psycopg2>=2.7,<3.0'],
+          'keystone-auth-backend': ['keystonemiddleware>=4.17,<5.0'],
+          'keystone-client': ['python-keystoneclient>=3.13,<4.0'],
```
naved001: why did you separate the two keystone packages? In most cases wouldn't we be installing both of them?

Izhmash: I was going to ask the same thing. I'm not sure when we'd only use one rather than both.

zenhack: I waffled on that. I'm fine combining tgem if you guys think it makes more sense.

I guess to some extent I think it would make sense to split the client library into a separate package altogether, but that's it's own can of worms.

Izhmash: If there's a chance that someone might need to use only one of those two, it could be good to keep them separate.  I'm fine with it how it is.

zenhack: Cool, I'm going to merge.

Issue #847, opened 2017-08-01
Last VLAN cannot be detached -> can't have port without VLAN


Izhmash:
- Strategy: 
    - Power off port when last VLAN (native) is detached
    - Power on port only if a native VLAN is attached
- Part of solution to removing the dummy VLAN

naved001:
@henn, @Izhmash @SahilTikale and I discussed how we handle native VLANs in HIL.

In most of the switches we have, we just can't remove the native VLAN and it always defaults to VLAN 1 in most cases. 

Our current workaround for this is to set it to some dummy VLAN and then disable it. It seems like we are relying on a weird glitch in the switch that we can set native vlan of a port to a disabled vlan. @radonm thought this should not be possible and is most likely a bug.

We think we should make it a requirement to connect a native vlan first before they can add any trunking VLANs and the port would remain turned off until a native vlan is added. This is how it will work:

* For a node in free pool, the port connected to any of it's nic would be turned off; so no traffic can flow at all.
* A user attaches a node to his project - the port is still off.
* User runs `node_connect_network` (channel = vlan/native) that attaches the native vlan. and in the driver, in `enable native`/`set_native` we turn on the port, and then do the VLANy bits. That way a user doesn't explicitly turn on the port. After this, a user can connect as many trunked vlans as he wants like the old days.
* when a user removes a native network, we remove the network and also turn off the port. (And if they have trunked vlans, we complain or warn them)

@zenhack @okrieg @pjd-nu what do you guys think?

naved001:
I think #846 and this would be bundled together

zenhack:
@naved001, that sounds reasonable to me.

naved001:
I am going to close this, since we'll be doing this on per switch basis 
* if a switch defaults to some vlan when the port is on, we will turn it off.  (dellnos9, brocade)
* if a switch has no vlans at all, then we don't have to turn it off. (powerconnects, nexus)

Issue #846, opened 2017-08-01
Require a native VLAN before making trunked ones


Izhmash:
- Most switches always require a native VLAN
- Part of the solution to getting rid of the dummy VLAN
- Points of discussion:
    - Should this requirement be configurable?

naved001:
Looks like we aren't gonna be forcing all switches to work the same way. #882 implements the title for the s3048-on switch, but not for other switches. 

We don't need to enforce this for other switches if a switch can just work fine without having a native vlan on a trunked port.

naved001:
Do we want to enforce this policy on all switches or not? I am not sure if the network topology could get complicated if some switches are allowed to do certain things while others aren't (because they can't anyway).

zenhack:
I think we're okay here; we can just enforce this on switches that need it. I'm going to close.

PR #852, opened 2017-08-08
Native VLAN requirement and native VLAN presense port power relation

Izhmash:
- In response to #846 and #847:
- API server now requires a native VLAN to exist if there are any trunked VLANs.
- When a port's native VLAN is removed from a switch port, the port powers off.
- When a port's native VLAN is created, the port powers on.
- This places direct port power out of the hands of the user.
- This PR is part of the solution to removing the dummy VLAN.
- Note: There are a few Travis errors that I'm aware of, but I wanted to put this out for review before going any further.

### DIFF
--- a/hil/api.py
+++ b/hil/api.py
@@ -407,6 +407,11 @@ def node_connect_network(node, nic, network, channel=None):
     if not allocator.is_legal_channel_for(channel, network.network_id):
         raise BadArgumentError("Channel %r, is not legal for this network." %
                                channel)
+    if channel != 'vlan/native' \
+            and db.session.query(model.NetworkAttachment) \
+            .filter(model.NetworkAttachment.channel == 'vlan/native') \
+            .count() < 1:
+        raise BlockedError("Please attach a native VLAN first.")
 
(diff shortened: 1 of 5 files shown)


hil/api.py
```diff
@@ -445,6 +454,15 @@ def node_detach_network(node, nic, network):
...
         raise BadArgumentError("The network is not attached to the nic.")
+    # Check if any other VLANs are attached before removing the native VLAN
+    if attachment.channel == 'vlan/native' \
+        and db.session.query(model.NetworkAttachment) \
+            .filter(model.NetworkAttachment.channel != 'vlan/native') \
+            .filter(model.NetworkAttachment.nic_id == nic.id) \
+            .count() > 0:
+        raise BlockedError(
+                "Please remove all trunked VLANs "
+                "before removing the native VLAN.")
```
zenhack: Having talked about this at the meeting, I think we're going to go with the approach outlined in #854, so I'm going to cut my review short. @Izhmash, sorry I didn't think of this before you went ahead and did all this work...

PR #855, opened 2017-08-09
Turn on more pylint warnings

zenhack:
* No more `import *`
* Check for unused variables.

I plan on doing a couple of these a week going forward.

### DIFF
--- a/hil/client/node.py
+++ b/hil/client/node.py
@@ -26,11 +26,11 @@ class Node(ClientBase):
         # Node requires which OBM, and knows arguments required
         # for successful node registration.
 
-        obm_api = "http://schema.massopencloud.org/haas/v0/obm/"
-        obm_types = ["ipmi", "mock"]
         # FIXME: In future obm_types should be dynamically fetched.
         # We need a new api call for querying available
         # and currently active drivers for HIL
+        # obm_api = "http://schema.massopencloud.org/haas/v0/obm/"
+        # obm_types = ["ipmi", "mock"]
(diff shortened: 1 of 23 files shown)


zenhack:
I don't understand why the github page says travis is still in progress -- if you click on the link it says it's done & passed.

hil/client/node.py
```diff
@@ -26,11 +26,11 @@ def register(self, node, subtype, *args):
         # Node requires which OBM, and knows arguments required
         # for successful node registration.
 
-        obm_api = "http://schema.massopencloud.org/haas/v0/obm/"
-        obm_types = ["ipmi", "mock"]
         # FIXME: In future obm_types should be dynamically fetched.
         # We need a new api call for querying available
         # and currently active drivers for HIL
+        # obm_api = "http://schema.massopencloud.org/haas/v0/obm/"
+        # obm_types = ["ipmi", "mock"]
```
naved001: off topic note: I read this the whole comment here, and now we have an API to get active drivers, so I guess we can implement this now (in a separate PR of course).

zenhack:
Quoting Naved Ansari (2017-08-10 14:34:35)

> -from hil import model
> +from hil import model, errors
>
>    if we import model here, then why are we importing db from hil.model
>    again in the next line?


We use the db variable often enough that it seemed a bit nicer to just
import that one directly. We *could* just do model.db.session...
everywhere; it's a question of style.

naved001:
Cool. I am happy.

PR #856, opened 2017-08-09
Perform networking actions one at a time

### LINKED ISSUE #314: record successful application of networks one at a time.

naved001:
Closes #314 

Collaborative effort with @henn 

I repackaged the PR (#673) raised by @matsuuran.

### DIFF
--- a/clearpost
+++ b/clearpost
@@ -1,0 +1,3 @@
+sudo service postgresql restart
+dropdb centos
+createdb centos
--- a/clearpost
+++ b/clearpost
@@ -1,3 +1,0 @@
-sudo service postgresql restart
-dropdb centos
-createdb centos
--- a/hil/deferred.py
+++ b/hil/deferred.py
@@ -93,7 +93,6 @@ def apply_networking():
     action = model.NetworkingAction.query \
         .order_by(model.NetworkingAction.id).first()
     if action is None:
-        # Optimization - avoid opening a switch session if there's no action
         db.session.commit()
         return False
 
@@ -106,6 +105,5 @@ def apply_networking():
         action = model.NetworkingAction.query \
             .order_by(model.NetworkingAction.id).first()
 
-    db.session.commit()  # Can only get here for action == None
     session.close()
     return True
--- a/hil/deferred.py
+++ b/hil/deferred.py
@@ -105,5 +105,9 @@ def apply_networking():
         action = model.NetworkingAction.query \
             .order_by(model.NetworkingAction.id).first()
 
+    # the last statement in the while loop opens a new db session that we must
+    # close when we exit the loop.
+    db.session.commit()
+
     session.close()
     return True
--- a/hil/deferred.py
+++ b/hil/deferred.py
@@ -82,26 +82,30 @@ def apply_networking():
     The networking server calls this function in a loop, to ensure that all
     pending network operations get processed within a reasonable amount of
     time.  The return value from this function lets the server know whether it
-    should check for new journal entries, or if it should wait.  If this
-    function does work, the server should immediately check again, because new
-    entries might have been added in the meantime.  But, if this function
-    returns immediately, the server should sleep, because there was no time
-    for new entries to be added.  This keeps the networking server from
+    should check for new journal entries immediately, or if it should wait.  If
+    this function does work, the server should immediately check again, because
+    new entries might have been added in the meantime.  But, if this function
+    returns immediately, the server should sleep, because there was no time for
+    new entries to be added.  This keeps the networking server from
     tight-looping.
     """
-    # Get the journal enries
-    actions = model.NetworkingAction.query \
-        .order_by(model.NetworkingAction.id).all()
 
-    if actions == []:
-        # No actions to perform.  Return False immediately.
+    action = model.NetworkingAction.query \
+        .order_by(model.NetworkingAction.id).first()
+    if action is None:
+        # Optimization - avoid opening a switch session if there's no action
         db.session.commit()
         return False
 
     session = DaemonSession()
-    for action in actions:
+    while action is not None:
         session.handle_action(action)
+        db.session.delete(action)
+        db.session.commit()
+        # Get the next action
+        action = model.NetworkingAction.query \
+            .order_by(model.NetworkingAction.id).first()
+
+    db.session.commit() # Can only get here for action == None
     session.close()
-    model.NetworkingAction.query.delete()
-    db.session.commit()
     return True
--- a/hil/deferred.py
+++ b/hil/deferred.py
@@ -106,6 +106,6 @@ def apply_networking():
         action = model.NetworkingAction.query \
             .order_by(model.NetworkingAction.id).first()
 
-    db.session.commit() # Can only get here for action == None
+    db.session.commit()  # Can only get here for action == None
     session.close()
     return True
--- a/tests/unit/deferred.py
+++ b/tests/unit/deferred.py
@@ -45,6 +45,7 @@ def configure():
     # if we are using sqlite's in memory db, then change uri to a db on disk
     uri = config.cfg.get('database', 'uri')
     if uri == 'sqlite:///:memory:':
+        assert False
         additional_config['database'] = {'uri': 'sqlite:////tmp/hil.db'}
 
     config_merge(additional_config)
--- a/tests/unit/deferred.py
+++ b/tests/unit/deferred.py
@@ -171,7 +171,6 @@ def test_apply_networking(switch, nic1, nic2, network, fresh_database):
     is thrown on the 3rd action, the 1st and 2nd action will have already been
     committed)
     '''
-    from hil.ext.obm.mock import MockObm
 
     port = model.Port(label=INTERFACE1, switch=switch)
     nic1.port = port
--- a/tests/unit/deferred.py
+++ b/tests/unit/deferred.py
@@ -80,6 +80,11 @@ def _deferred_test_switch_class():
         function.  It is needed for a custom implementation of the switch's
         apply_networking() that counts the networking actions as
         apply_networking() is called to ensure it is incremental.
+
+        It is defined as a fixture because if it is defined as a class with
+        global scope it's going to be defined for every test; so this will be
+        picked up by migration tests and those would fail because there will be
+        no migration scripts for this switch's table.
         '''
 
         api_name = 'http://schema.massopencloud.org/haas/v0/switches/deferred'
@@ -187,36 +192,36 @@ def test_apply_networking(switch, network, fresh_database):
                                              channel='vlan/native',
                                              type='modify_port'))
 
-    # this makes the last action invalid for the test switch which will raise
-    # an exception.
+    # this makes the last action invalid for the test switch because the switch
+    # raises an error when the networking action is of type revert port.
     action[2] = model.NetworkingAction(nic=nic[2],
                                        new_network=None,
                                        channel='',
                                        type='revert_port')
-    db.session.add(action[0])
-    db.session.add(action[1])
-    db.session.add(action[2])
+    for i in range(0, 3):
+        db.session.add(action[i])
     db.session.commit()
-    try:
+
+    with pytest.raises(SwitchError):
         deferred.apply_networking()
-    except SwitchError:
-        # close the session opened by `apply_networking` when `handle_actions`
-        # fails; without this the tests would just stall (when using postgres)
-        db.session.close()
 
-        local_db = new_db()
+    # close the session opened by `apply_networking` when `handle_actions`
+    # fails; without this the tests would just stall (when using postgres)
+    db.session.close()
 
-        pending_action = local_db.session \
-            .query(model.NetworkingAction) \
-            .order_by(model.NetworkingAction.id).first()
-        current_count = local_db.session \
-            .query(model.NetworkingAction).count()
+    local_db = new_db()
 
-        local_db.session.delete(pending_action)
-        local_db.session.commit()
-        local_db.session.close()
+    pending_action = local_db.session \
+        .query(model.NetworkingAction) \
+        .order_by(model.NetworkingAction.id).first()
+    current_count = local_db.session \
+        .query(model.NetworkingAction).count()
 
-        # test that there's only pending action in the queue and it is of type
-        # revert_port
-        assert current_count == 1
-        assert pending_action.type == 'revert_port'
+    local_db.session.delete(pending_action)
+    local_db.session.commit()
+    local_db.session.close()
+
+    # test that there's only pending action in the queue and it is of type
+    # revert_port
+    assert current_count == 1
+    assert pending_action.type == 'revert_port'
--- a/tests/unit/deferred.py
+++ b/tests/unit/deferred.py
@@ -15,7 +15,6 @@
 '''Functional test for deferred.py'''
 
 import pytest
-import os
 import tempfile
 
 from hil import config, deferred, model
@@ -34,7 +33,6 @@ INTERFACE2 = '104/0/18'
 INTERFACE3 = '104/0/20'
 
 DeferredTestSwitch = None
-temp_db = tempfile.NamedTemporaryFile()
 
 
 class SwitchError(Exception):
@@ -54,11 +52,16 @@ def configure():
     # if we are using sqlite's in memory db, then change uri to a db on disk
     uri = config.cfg.get('database', 'uri')
     if uri == 'sqlite:///:memory:':
-        uri = 'sqlite:///' + temp_db.name
-        additional_config['database'] = {'uri': uri}
-
-    config_merge(additional_config)
-    config.load_extensions()
+        with tempfile.NamedTemporaryFile() as temp_db:
+            uri = 'sqlite:///' + temp_db.name
+            additional_config['database'] = {'uri': uri}
+            config_merge(additional_config)
+            config.load_extensions()
+            yield
+    else:
+        config_merge(additional_config)
+        config.load_extensions()
+        yield
 
 
 @pytest.fixture()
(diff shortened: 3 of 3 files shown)


hil/deferred.py
```diff
@@ -82,26 +82,30 @@ def apply_networking():
...
-    for action in actions:
+    while action is not None:
         session.handle_action(action)
+        db.session.delete(action)
+        db.session.commit()
+        # Get the next action
+        action = model.NetworkingAction.query \
+            .order_by(model.NetworkingAction.id).first()
+
+    db.session.commit()  # Can only get here for action == None
```
zenhack: This could be a bit clearer; maybe something like "The last statement in the while loop will have opened a new database session, so we need to close it on exit." It took me a bit to figure out what was going on here, and the comment just re-states the condition in the while loop, which is unhelpful.

zenhack:
(Also, I fixed the link to the original PR in your comment; it was bothering me).

hil/deferred.py
```diff
@@ -82,26 +82,30 @@ def apply_networking():
...
-    # Get the journal enries
-    actions = model.NetworkingAction.query \
-        .order_by(model.NetworkingAction.id).all()
 
-    if actions == []:
-        # No actions to perform.  Return False immediately.
+    action = model.NetworkingAction.query \
+        .order_by(model.NetworkingAction.id).first()
+    if action is None:
+        # Optimization - avoid opening a switch session if there's no action
```
naved001: right, so a session is only created when we call `handle_action`. In that case we can remove the oart above the while loop. How would this look like then?

```python
    session = DaemonSession()

    while True:
        action = model.NetworkingAction.query \
            .order_by(model.NetworkingAction.id).first()
        if action is None:
            db.sesssion.commit()
            return False
        session.handle_action(action)
        db.session.delete(action)
        db.session.commit()

    session.close()
    return True
```
Ideally I would want a do-while here, but python doesn't have it so I am using the `while True` and `break on condition` combo.

@henn what do you think?

naved001:
> Would you address that as well?

I am reading about it right now and then I'll address it.

Izhmash:
I agree with the concern in #673, but other than that this looks fine to me.

naved001:
@zenhack 
> my question was about whether the loop in apply_networking and the check in the switch driver are sharing a transaction, regardless of when it starts or ends.

You're right. The switch and `apply_networking` are sharing a transaction.  Because if you remove  `db.session.commit` in `apply_networking`, the tests still pass; implying that the switch could see uncommitted changes.

naved001:
I was trying to setup a different transaction in the switch driver. But it doesn't find my table. If I use a db on disk for this test, then I can have 2 connections open (1 by `apply_networking` and another by the switch). Am I doing something silly? 

```python

        def modify_port(self, port, channel, network_id):

            # setup a different SQL Alchemy connection
            local_app = Flask('hil')
            uri = 'sqlite:///:memory:'
            local_app.config.update(SQLALCHEMY_TRACK_MODIFICATIONS=False)
            local_app.config.update(SQLALCHEMY_DATABASE_URI=uri)
            local_db = SQLAlchemy(local_app)
            
            current_count = local_db.session.query(model.NetworkingAction).count()
```

hil/deferred.py
naved001: The code I wrote above would never return True. : O

naved001:
Don't know what's wrong with postrgres tests

tests/unit/deferred.py
```diff
@@ -34,11 +35,19 @@
...
             'hil.ext.obm.mock': ''
-        },
-    })
+            }
+        }
+
+    # if we are using sqlite's in memory db, then change uri to a db on disk
+    uri = config.cfg.get('database', 'uri')
+    if uri == 'sqlite:///:memory:':
+        additional_config['database'] = {'uri': 'sqlite:////tmp/hil.db'}
```
zenhack: Probably better to use the `tempfile` module, and delete the file when the test exits.

I'm surprised this is passing on sqlite -- we're running the tests in parallel, so I would expect tests to trample on each other using the same db. Can you run travis with an assert False in here, just to verify that it's actually going into the if statement?

zenhack:
Other options include binary search with an assert False (still going to be slow and painful), or doing something like this:

https://github.com/zenhack/hil/commit/197960eddca1fd708bad50f8766a6f0a7fa3f817

(and doing netcat -l PORT on the machine you specify).

naved001:
@zenhack Fixed the tests, it was a dumb mistake (see the commit message). What threw me off was that it was passing on sqlite. Also, I have made the other changes you requested.

@djfinn14 can you review this? Thanks!

tests/unit/deferred.py
```diff
@@ -0,0 +1,201 @@
...
+    class DeferredTestSwitch_(Switch):
+        '''DeferredTestSwitch
+
+        This is a switch implemented to test the deferred.apply_networking()
+        function.  It is needed for a custom implementation of the switch's
+        apply_networking() that counts the networking actions as
+        apply_networking() is called to ensure it is incremental.
+        '''
+
+        api_name = 'http://schema.massopencloud.org/haas/v0/switches/deferred'
```
henn: Please s/haas/hil/

tests/unit/deferred.py
```diff
@@ -0,0 +1,201 @@
...
+
+        def disconnect(self):
+            pass
+
+        def modify_port(self, port, channel, network_id):
+
+            # setup a different connection to database so that this method does
+            # not see uncommited changes by `apply_networking`
+
+            local_app = Flask(__name__.split('.')[0])
```
henn: Do you know of an easier way of getting a separate flask context than doing all of this inline?

Maybe [app_context()](http://flask.pocoo.org/docs/0.10/appcontext/)?

tests/unit/deferred.py
```diff
@@ -0,0 +1,201 @@
...
+@pytest.fixture()
+def network():
+    project = model.Project('anvil-nextgen')
+    return model.Network(project, [project], True, '102', 'hammernet')
+
+pytestmark = pytest.mark.usefixtures('configure',
+                                     'fail_on_log_warnings')
+
+
+def test_apply_networking(switch, nic1, nic2, network, fresh_database):
```
henn: Would it be possible to add another test that ensures that if action2 fails, action1 successfully was committed?

tests/unit/deferred.py
naved001: We still use haas in schema URLs [everywhere](https://github.com/CCI-MOC/hil/commit/9ab69122f762196da5c56478850531e98c8d796d) which is why I left it here.

zenhack:
I agree with @henn's comments, but I'm otherwise happy.

tests/unit/deferred.py
```diff
@@ -0,0 +1,201 @@
...
+    additional_config = {
+        'extensions': {
+            'hil.ext.obm.mock': ''
+            }
+        }
+
+    # if we are using sqlite's in memory db, then change uri to a db on disk
+    uri = config.cfg.get('database', 'uri')
+    if uri == 'sqlite:///:memory:':
+        uri = 'sqlite:///' + temp_db.name
```
naved001: In that case I had the yield inside the if statement, and if that branch is not taken (i.e. running tests with postgres), the tests error out with the following error:
`.venv/lib/python2.7/site-packages/_pytest/fixtures.py:699: StopIteration`

```python
    if uri == 'sqlite:///:memory:':
        with tempfile.NamedTemporaryFile() as temp_db:
            uri = 'sqlite:///' + temp_db.name
            additional_config['database'] = {'uri': uri}
            config_merge(additional_config)
            config.load_extensions()
            yield
    else:
        config_merge(additional_config)
        config.load_extensions()
```

If i get rid of yield, the db is still available to the rest of the code and is cleaned up on exit.

zenhack: One way around that is to just do a yield in the else branch as well; this way the fixture machinery still does the right thing, even if there isn't actually any cleanup to do.

The bigger issue is that by initializing temp_db at global scope, all the tests are sharing that one file, as opposed to having their own. As @henn mentioned, this will likely cause problems for parallelizing the tests. I suspect those problems will just be that the tests will have to acquire a lock on the db, and thus serialize, which is why they're still passing, but that's still sub-optimal.

naved001: > One way around that is to just do a yield in the else branch as well

Cool, that worked.

> The bigger issue is that by initializing temp_db at global scope, all the tests are sharing that one file, as opposed to having their own

How are other tests sharing this database? As far as I understand, only this test (`tests/unit/deferred.py`) is using the disk on db. Other tests are creating their own in-memory db.

naved001:
@henn I have made the changes you requested.

tests/unit/deferred.py
henn: When/if we introduce a second test in this file, and if someone were to run `pytest -n auto ...`, then the global could be shared.

tests/unit/deferred.py
henn: `haas` in the schema URLs was kept for backwards compatibility/migration issues, but since this is new code we could change that with minimal fuss.

That being said, I don't feel strongly on this.

tests/unit/deferred.py
```diff
@@ -96,17 +105,9 @@ def modify_port(self, port, channel, network_id):
...
-            uri = config.cfg.get('database', 'uri')
-            local_app.config.update(SQLALCHEMY_TRACK_MODIFICATIONS=False)
-            local_app.config.update(SQLALCHEMY_DATABASE_URI=uri)
-            local_db = SQLAlchemy(local_app)
-
-            current_count = local_db.session \
-                .query(model.NetworkingAction).count()
-
-            local_db.session.commit()
+            with app.app_context():
```
henn: @zenhack - does this look right to you?

Skimming the docs again, I'm not sure whether app_context() or [scoped_session()](http://docs.sqlalchemy.org/en/latest/orm/contextual.html#unitofwork-contextual) is better.

tests/unit/deferred.py
```diff
@@ -140,23 +143,23 @@ def nic1():
                 host="ipmihost",
                 user="root",
                 password="tapeworm")),
-        'ipmi',
+        name,
         '00:11:22:33:44:55')
 
 
 @pytest.fixture()
+def nic1():
```
henn: Could we remove these as fixtures and inline these to the test itself?

I'm not sure that having them as fixtures helps unless we have additional tests beyond the first one.

naved001: done

tests/unit/deferred.py
zenhack: Honestly, now that I think about it I like @naved001's original solution better than either. The problem with `app_context` is that here we have *nested* app contexts, and I can't find any documentation talking about that situation at all; I wouldn't want to just assume it's valid, although it may be. The issue with `scoped_session` is that I'm not sure where `Flask-SQLAlchemy` stores our session factory, and more generally, the details of that sort of integration are poorly documented. The original code is clearly correct without having to delve into arcane details of how the frameworks operate; that's worth a few extra lines of code (wish I'd spoken up a bit earlier).

tests/unit/deferred.py
```diff
@@ -193,9 +196,26 @@ def test_apply_networking(switch, nic1, nic2, network, fresh_database):
                                             channel='vlan/native',
                                             type='modify_port')
 
+    action_native3 = model.NetworkingAction(nic=nic3,
+                                            new_network=None,
+                                            channel='',
+                                            type='revert_port')
```
zenhack: 1. I'd suggest putting these in a list; variables like foo1 foo2 foo3 are generally a code smell.
2. This could use a comment pointing out that action_native3 is invalid, to trigger the failure below. Similarly, you should update the docstring to mention the additional thing this test is checking.

tests/unit/deferred.py
```diff
@@ -193,9 +196,26 @@ def test_apply_networking(switch, nic1, nic2, network, fresh_database):
...
     db.session.commit()
+    try:
+        deferred.apply_networking()
+    except SwitchError:
+        with app.app_context():
+            pending_action = db.session \
+                .query(model.NetworkingAction). \
+                order_by(model.NetworkingAction.id).first()
+            current_count = db.session \
+                .query(model.NetworkingAction).count()
```
zenhack: I have my same worries about the nested app context here.

tests/unit/deferred.py
zenhack: Actually, now that I look at this, I'm not sure this code works -- that check is in the except so it may just be dead code. I don't think we actually validate the information in the networking daemon in a way that would catch this.

tests/unit/deferred.py
henn: Ok. I'll defer to you on this.

henn:
Interestingly, this brings up the point that we don't try to catch any exceptions within the networking action server.

Even if we did, I'm not sure that we could do something besides exiting anyhow, since an error indicates either a config or switch problem that needs attention.

zenhack:
@henn, yep, that's true. I suppose under normal operation that should "never" happen. But of course stuff happens. It might be prudent to try a couple times before just grinding the system to a halt on the first blip, and maybe somehow try to alert an adminstrator if the failures seem to be persistent.

tests/unit/deferred.py
```diff
@@ -196,15 +206,23 @@ def test_apply_networking(switch, network, fresh_database):
...
-                .query(model.NetworkingAction).count()
-
-        # assert that there's only 1 remaining action and that has type
-        # revert_port. confirming that previous modify_port operations were
-        # completed successfully.
+        local_app = Flask(__name__.split('.')[0])
+        uri = config.cfg.get('database', 'uri')
+        local_app.config.update(SQLALCHEMY_TRACK_MODIFICATIONS=False)
+        local_app.config.update(SQLALCHEMY_DATABASE_URI=uri)
+        local_db = SQLAlchemy(local_app)
```
zenhack: Should probably factor this out into its own function; this is copypasta from above.

clearpost
```diff
@@ -0,0 +1,3 @@
+sudo service postgresql restart
+dropdb centos
+createdb centos
```
zenhack: Was this committed by accident? If not, what's it for?

naved001:
Oh yeah. Sorry. I wanted to save my work so I committed by accident. The pr isn't final yet; I need to test some more stuff.

tests/unit/deferred.py
naved001: Done

tests/unit/deferred.py
naved001: see [this failure](https://travis-ci.org/CCI-MOC/hil/jobs/266135520#L2363) in my previous travis build where I made an incorrect assertion and that caused it to fail.

clearpost
naved001: yes, I removed it now.

naved001:
@zenhack it's ready for review now. I think I have addressed all your comments.

tests/unit/deferred.py
```diff
@@ -0,0 +1,222 @@
...
+            config.load_extensions()
+            yield
+    else:
+        config_merge(additional_config)
+        config.load_extensions()
+        yield
+
+
+@pytest.fixture()
+def _deferred_test_switch_class():
```
zenhack: We should add a docstring explaining why this is a fixture, instead of just declaring a class once at global scope -- that's something that's not immediately apparent from the code.

tests/unit/deferred.py
```diff
@@ -0,0 +1,222 @@
...
+
+    # this makes the last action invalid for the test switch which will raise
+    # an exception.
+    action[2] = model.NetworkingAction(nic=nic[2],
+                                       new_network=None,
+                                       channel='',
+                                       type='revert_port')
+    db.session.add(action[0])
+    db.session.add(action[1])
+    db.session.add(action[2])
```
zenhack: This can be a for loop. It only saves one line, but makes the pattern more clear.

tests/unit/deferred.py
```diff
@@ -0,0 +1,222 @@
...
+    for i in range(0, 3):
+        interface = 'gi1/0/%d' % (i)
+        nic.append(new_nic(str(i)))
+        nic[i].port = model.Port(label=interface, switch=switch)
+        action.append(model.NetworkingAction(nic=nic[i], new_network=network,
+                                             channel='vlan/native',
+                                             type='modify_port'))
+
+    # this makes the last action invalid for the test switch which will raise
+    # an exception.
```
zenhack: It could still be made more clear *why* this makes it invalid -- it isn't obvious unless you go read the implementation of revert_port.

tests/unit/deferred.py
```diff
@@ -0,0 +1,222 @@
...
+            .query(model.NetworkingAction).count()
+
+        local_db.session.delete(pending_action)
+        local_db.session.commit()
+        local_db.session.close()
+
+        # test that there's only pending action in the queue and it is of type
+        # revert_port
+        assert current_count == 1
+        assert pending_action.type == 'revert_port'
```
zenhack: The way this is organized now, if something goes wrong and a `SwitchError` doesn't get raised at all, this check will just get skipped. I'd suggest:

```python
with pytest.raises(SwitchError):
     deferred.apply_networking()

# Everything that's currently in the except block.
```

... which will fail the test if the exception is not raised.

naved001: yeah, that's better!

naved001:
> There's no reason to use range here

Good point. Changed it.

PR #857, opened 2017-08-14
Enable a few more pylint warnings.

zenhack:


naved001:
@jeremyfreudberg want to review this?

hil/model.py
```diff
@@ -313,10 +313,13 @@ def validate(kwargs):
...
         assert False, "Subclasses MUST override the validate method "
 
-    def power_cycle(self):
+    def power_cycle(self, force):
         """Power cycles the node.
 
         Exact implementation is left to the subclasses.
+
+        ``force`` indicates that the node should be *forced* off, as opposed to
+        e.g. given an ACPI shutdown signal, to which the node respond.
```
jeremyfreudberg: responds? would respond?

zenhack:
Quoting Jeremy Freudberg (2017-08-15 13:36:14)

> >    In hil/model.py:
> >          """Power cycles the node.
>
>          Exact implementation is left to the subclasses.
> +
> +        ``force`` indicates that the node should be *forced* off, as opposed to
> +        e.g. given an ACPI shutdown signal, to which the node respond.
>
>    responds? would respond?

Good catch; fixed.

PR #862, opened 2017-09-04
Improve checks for bad format string usage with logging.

zenhack:


tests/custom_lint.py
```diff
@@ -74,7 +74,14 @@ def visit_Call(self, node):
         #    in a variable (don't do that). We could be smarter about
         #    figuring out what function is being called, but this is
         #    probably good enough:
-        logfunc_names = {'critical', 'error', 'warn', 'info', 'debug'}
+        logfunc_names = {
+            'critical',
+            'error',
+            'warn',
+            'warning',
```
naved001: difference between warn and warning?

zenhack: They do the same thing.

zenhack:
@naved001, per the commit message, pylint's check missed a number of instances of the problem when I first tried it, so it didn't seem prudent to rely on it exclusively.

tests/custom_lint.py
naved001: okay maybe my `grep`ing skills aren't that good, but I couldn't find any place where we are using `logger.warning`

Why did you add it to this list then?

naved001: Looks like it's in config.py (thanks @jeremyfreudberg!) why don't we change warn to warning  (or vice versa) everywhere? Just so that we are consistent.

zenhack: I'm fine with that; I'll push the change. But I think it still makes sense to add this to the list.

zenhack: (pushed)

PR #866, opened 2017-09-06
Driver for Dell S3048-ON

### LINKED ISSUE #844: Dell s3048-on switch driver/Dell NOS9

naved001:
Closes #844 

* I took the skeleton of the brocade driver and built on top of it. Added some additional helper functions.

* Link to [API guide](http://downloads.dell.com/manuals/all-products/esuprt_ser_stor_net/esuprt_networking/esuprt_net_netwrkng_sw/force10-open-automation_service%20manual9_en-us.pdf) and [CLI Guide](https://topics-cdn.dell.com/pdf/force10-s3048-on_Service%20Manual_en-us.pdf)

* There's one issue that still needs to be solved. I can't remove the Native vlan on this switch, it always defaults to 1 (just like the brocade). To run the deployment tests, I have put in a temporary workaround (it's in the comments) to make sure there are no other issues and it passes both deployment tests.  Any suggestions?

* I am doing some string parsings/manipulation, and if you have better suggestions on how to do that then please let me know.

* I have two of these switches (with all ports available), so if you want to play with it let me know (it would be nice if someone besides me gets to know the switch)


naved001:
oh no, I got @zenhack's name wrong in the last commit :(

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,392 @@
...
+            interface: interface to remove the vlan from
+            vlan: vlan to remove
+        """
+        url = self._construct_url()
+        command = 'interface vlan ' + vlan + '\r\n no tagged ' + \
+            self.interface_type + ' ' + interface
+        payload = self._make_payload(CONFIG, command)
+        self._make_request('POST', url, data=payload)
+
+    def _remove_all_vlans_from_trunk(self, interface):
```
naved001: now that I think of it, I could just call `_remove_vlan_from_trunk` in a loop; because that is exactly what I am doing here.

Or maybe not define this method at all. And do the loop thing in `revert_port` directly.

naved001:
Passing the deployment tests (with the hack)

```
collected 2 items                                                                                                                                                   

==================================================================== 2 passed in 326.73 seconds ====================================================================
```

It's quite slow, but not as slow as our pexpect dells

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,387 @@
...
+        (port,) = filter(lambda p: p.label == port, self.ports)
+        interface = port.label
+
+        if channel == 'vlan/native':
+            if new_network is None:
+                self._remove_native_vlan(interface)
+            else:
+                self._set_native_vlan(interface, new_network)
+        else:
+            match = re.match(re.compile(r'vlan/(\d+)'), channel)
```
zenhack: Maybe just call `get_network_allocator().is_legal_channel_for`?

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,387 @@
...
+
+        Returns: Dictionary containing the configuration of the form:
+        {
+            Port<"port-3">: [("vlan/native", "23"), ("vlan/52", "52")],
+            Port<"port-7">: [("vlan/23", "23")],
+            Port<"port-8">: [("vlan/native", "52")],
+            ...
+        }
+
+        """
```
zenhack: You don't really need to duplicate the docstring from the parent class, and doing so is just going to introduce a hazard of having them get out of sync.

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,387 @@
...
+        response = {}
+        for port in ports:
+            response[port] = filter(None,
+                                    [self._get_native_vlan(port.label)]) \
+                                    + self._get_vlans(port.label)
```
zenhack: This is a bit convoluted; the use of filter on a one-element list seems like overkill; I think a simple if statement would be clearer.

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,387 @@
...
+        Similar to _get_vlans()
+        """
+
+        url = self._construct_url()
+        command = 'interfaces switchport %s %s' % \
+            (self.interface_type, interface)
+        payload = self._make_payload(SHOW, command)
+        response = self._make_request('POST', url, data=payload)
+        # find the Native Vlan ID from the response
+        response = response.text.replace(' ', '')
```
zenhack: Everything from here up is identical to _get_vlans(); factor this out.

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,387 @@
...
+        if response.find('NativeVlanId:') == -1:
+            return None
+        begin = response.find('NativeVlanId:') + len('NativeVlanId:')
+        end = response.find('.', begin)
+        vlan = response[begin:end]
+        # a really bad temporary workaround to get the deployment tests to pass
```
zenhack: Explain what it's doing, why exactly the tests are failing without it, and what's wrong with it.

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,387 @@
...
+        """ Set the native vlan of an interface.
+
+        Args:
+            interface: interface to set the native vlan to
+            vlan: vlan to set as the native vlan
+
+        Method relies on the REST API CLI which is slow
+        """
+        if not self._is_port_on(interface):
+            self._port_on(interface)
```
zenhack: There's a (security sensitive) race-condition here; in the time between turning on the port and setting the right value for the vlan, the node has access to whatever vlan it was on before.

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,387 @@
...
+        url = self._construct_url()
+        command = 'interface vlan ' + vlan + '\r\n untagged ' + \
+            self.interface_type + ' ' + interface
+        payload = self._make_payload(CONFIG, command)
+        self._make_request('POST', url, data=payload)
```
zenhack: This should probably get factored out into a helper method; it looks like almost the same copypasta from _get_vlans() and _get_native_vlan().

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,387 @@
...
+            interface: interface to remove the native vlan from.vlan
+
+        Method relies on the REST API CLI which is slow.
+        """
+        vlan = self._get_native_vlan(interface)[1]
+        url = self._construct_url()
+        command = 'interface vlan ' + vlan + '\r\n no untagged ' + \
+            self.interface_type + ' ' + interface
+        payload = self._make_payload(CONFIG, command)
+        self._make_request('POST', url, data=payload)
```
zenhack: Same thing, helper method.

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,387 @@
...
+
+    def _port_shutdown(self, interface):
+        """ Shuts down <interface>
+
+        Turn off portmode hybrid, disable switchport, and then shut down the
+        port. All non-default vlans must be removed before calling this.
+
+        Fast method that uses the REST API (no hackery).
```
zenhack: This doesn't really belong in the docstring, as it's more of an implementation question (same goes elsewhere). If you want to include it, I would do it in a comment just below this, but what I think makes more sense is to put TODO/XXX comments int he places where improvement is desired.

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,387 @@
...
+    def _convert_interface(interface_type):
+        """ Convert the interface name from switch CLI form to what the API
+        server understands.
+
+        Args:
+            interface: the interface in the CLI-form
+
+        Returns: string interface
+        """
+        iftypes = {'GigabitEthernet': 'gige-', 'TenGigabitEthernet': 'tengig-',
```
zenhack: Keep these to one key-value pair per line; having two on some of the lines makes it much harder to read.

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,387 @@
...
+            interface_type = 'vlan-'
+
+        return '%(hostname)s/api/running/dell/interfaces/interface/' \
+            '%(interface_type)s%(interface)s%(suffix)s' \
+            % {
+                'hostname': self.hostname,
+                'interface_type': interface_type,
+                'interface': interface,
+                'suffix': '/%s' % suffix if suffix else ''
+            }
```
zenhack: Given that you're not actually using anything but %s here, it might make more sense to just use `''.join` instead of a format string.

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,387 @@
...
+                  '<shutdown>false</shutdown></interface>' % interface
+
+        self._make_request('PUT', url, data=payload)
+
+    def _is_port_on(self, port):
+        """ Returns a boolean that tells the status of a switchport
+        Fast method that uses the REST API (no hackery)
+        """
+
+        url = self._construct_url(interface=port, suffix='\?with-defaults')
```
zenhack: As far as I can tell this is the only place you're using the suffix argument; why not remove it and just do `+ '\?with-defaults'` here? keeps the complexity out of everything else.

Also: did you really want a backslash there?

hil/ext/switches/dellnos9.py
naved001: This is copy pasta from the brocade code. But yeah, I could change that.

hil/ext/switches/dellnos9.py
naved001: Cool. I'll remove it. I just copied this from the brocade.

hil/ext/switches/dellnos9.py
naved001: I honestly didn't think much about it because I copied this from the brocade too. I'll make it simpler though.

hil/ext/switches/dellnos9.py
naved001: I wanted to leave it as it is because it might be helpful later when we do some more stuff that might need suffix.

> did you really want a backslash there?

The documentation had that, so I kept it as it is. But you can put any combination of forward and backslashes and it still works.

hil/ext/switches/dellnos9.py
naved001: I don't think so. A port can only be off if there are no vlans on it.l so when a port is turned on it starts from a default state.

Earlier I thought I could implement revert port by just toggling the port off and on, but the switch doesn't allow that. That's why I added a method to remove all trunked vlans.

hil/ext/switches/dellnos9.py
naved001: Do you want me to add it as a comment here or explain it to you (because I am pretty sure you already know what's up) ?

hil/ext/switches/dellnos9.py
zenhack: Is it not on any vlans when it's first turned on? my understanding was we were turning off the port because we couldn't do that. Otherwise, it's on *some* vlan, so unless that's already the one we want to put it on, there is a brief moment where it's on the wrong network. If something goes wrong and we get disconnected from the switch in the middle of this, that leaves us in a situation where that node has illicit access to that network.

hil/ext/switches/dellnos9.py
zenhack: OK. add a comment explaining what's going on there?

hil/ext/switches/dellnos9.py
zenhack: In that case, it should probably get changed there as well.

Actually, this whole method feels like boilerplate that should probably get factored out into a super class or something (similarly, the powerconnect and nexus drivers share a lot of structure like this, hence _console.py).

hil/ext/switches/dellnos9.py
zenhack: as a comment, sorry if that wasn't clear.

hil/ext/switches/dellnos9.py
naved001: That's a good point!

You are right, when a switch is first turned on it's starts with a Native Vlan of 1. But I don't really know how to handle that in the driver. A port must be turned on before we do anything to it; and it will start with a default native vlan. 

This seems to be an issue with the [brocade too](https://github.com/CCI-MOC/hil/blob/master/hil/ext/switches/brocade.py#L159)

Doesn't look like we are handling it?

hil/ext/switches/dellnos9.py
naved001: cool. will do that.

zenhack:
>    This seems to be an issue with the brocade too
>
>    Doesn't look like we are handling it?

OK, well we can't just leave this as is (though I might be convinced to
merge this and open a separate issue for the race, since it's the same
bug that we've already got, but see also my comments about factoring out
common logic).  It's not immediately obvious to me what to do about it,
but I will think on the matter.

-Ian

naved001:
> OK, well we can't just leave this as is (though I might be convinced to merge this and open a separate issue for the race, since it's the same bug that we've already got, but see also my comments about factoring out common logic). It's not immediately obvious to me what to do about it, but I will think on the matter.

I had a chat with @okrieg and he proposed a solution (similar to  #852 ):

HIL should enforce that a node can only be powered on when a native vlan has been set. 
So when a user gets a node, they must first set a native vlan. The call to `node_power_on` should ensure that a native vlan has been set on all connected nics (make a call to the switch to get the native vlan) before it actually powers on the node.

The idea seems simple enough, if we agree on this we could think about the implementation details. 

@henn curious to your thoughts on this as well.

zenhack:
That sounds like a basically sane solution to me. I'm a little worried about the proliferation of stateful safety constraints in our code, since it's going to make implementation more hazardous.

naved001:
I think I have factored out what I thought was reasonable and addressed the comments you made.

Re: using `is_legal_channel_for()`, I made that changed but now this happened when I was running deployment tests.
```
self = <hil.ext.network_allocators.vlan_pool.VlanAllocator object at 0x7f141919af10>, net_id = None

    def legal_channels_for(self, net_id):
        return ["vlan/native",
>               "vlan/" + net_id]
E       TypeError: cannot concatenate 'str' and 'NoneType' objects
```

Before I fix that, I am curious why are we doing this check again in the driver? We already do it in the [API here](https://github.com/CCI-MOC/hil/blob/master/hil/api.py#L419)

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,377 @@
...
+    def revert_port(self, port):
+        self._remove_all_vlans_from_trunk(port)
+        if self._get_native_vlan(port) is not None:
+            self._remove_native_vlan(port)
+
+    def get_port_networks(self, ports):
+        response = {}
+        for port in ports:
+            native = self._get_native_vlan(port.label)
+            # return an empty list if native is None
```
zenhack: Better to just change the expression to make this more obvious; options:

* `native = [native] if native is not None else []`
* `native = [] if native is None else [native]`
* An if-statement, rather than expression:

```python
if native is None:
    native = []
else:
    native = [native]
```

zenhack: Better yet, you could just do:

```python
response[port] = self._get_vlans(port.label)
if native is not None:
    response[port].append(native)
```

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,377 @@
...
+            return None
+        begin = response.find('NativeVlanId:') + len('NativeVlanId:')
+        end = response.find('.', begin)
+        vlan = response[begin:end]
+
+        # At this point, we are unable to remove the default native vlan.
+        # The switchport defaults to native vlan 1. Our deployment tests make
+        # assertions that a switchport has no default vlan at start.
+        # To get the tests to pass, we return None if we see the switchport has
+        # a default native vlan (=1). This needs to be fixed ASAP.
```
zenhack: maybe put the work FIXME in here, just so it's easier to grep for.

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,377 @@
...
+        the caller.
+
+        Sample Response:
+        u"<outputxmlns='http://www.dell.com/ns/dell:0.1/root'>\n<command>
+        showinterfacesswitchportGigabitEthernet1/3\r\n\r\nCodes:U-Untagged,
+        T-Tagged\r\nx-Dot1xuntagged,X-Dot1xtagged\r\nG-GVRPtagged,M-Trunk\r\n
+        i-Internaluntagged,I-Internaltagged,v-VLTuntagged,V-VLTtagged\r\n\r\n
+        Name:GigabitEthernet1/3\r\n802.1QTagged:Hybrid\r\nVlanmembership:\r\n
+        QVlans\r\nU1512\r\n T1511 \r\n\r\n NativeVlanId:1512 .\r\n\r\n\r\n\r\n
+        MOC-Dell-S3048-ON#</command>\n</output>\n"
```
zenhack: You could probably format this in a way that's easier to read, e.g:

```
u"<output xmlns='http://www.dell.com/ns/dell:0.1/root'>\n" \
    "<command>showinterfacesswitchportGigabitEthernet1/3\r\n" \
    "\r\n" \
    "Codes:U-Untagged,..." \
    ...
```

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,377 @@
...
+        """ Remove all vlan from a trunk port.
+
+        Args:
+            interface: interface to remove the vlan from
+
+        this is kind slow for a number of reasons
+        1. we are using the REST API CLI
+        2. we have to go to every vlan and then remove our port from it;
+           that makes multiple calls.
+        """
```
zenhack: I don't think we should do it right now, but question: is it possible to parallelize this, or will the switch choke on that? I suspect we can improve on the latency if we have more than one request in flight at a time. Also, how slow are we talking?

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,377 @@
...
+
+    def _is_port_on(self, port):
+        """ Returns a boolean that tells the status of a switchport"""
+
+        # the url here requires a suffix to GET the shutdown tag in response.
+        url = self._construct_url(interface=port) + '\?with-defaults'
+        response = self._make_request('GET', url)
+        root = etree.fromstring(response.text)
+        shutdown = root.find(self._construct_tag('shutdown')).text
+        # is error handling needed here?
```
zenhack: Probably for robustness, we should do something like:

```python
if shutdown not in ('true', 'false'):
    raise SomeKindOfError()
return shutdown == 'false'
```

...rather than treating possibly bogus data as the port being off, which could result in various information leaks.

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,377 @@
...
+            # interface refers to `vlan`
+            interface_type = 'vlan-'
+
+        return ''.join([self.hostname, '/api/running/dell/interfaces/'
+                       'interface/', interface_type, interface])
+
+    @staticmethod
+    def _convert_interface(interface_type):
+        """ Convert the interface name from switch CLI form to what the API
+        server understands.
```
zenhack: This docstring doesn't look like it's quite right -- you're converting interface *types*, not full names of specific interfaces, unless I'm mistaken.

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,377 @@
...
+
+        return '<input><%s>%s</%s></input>' % (command, command_type, command)
+
+    @staticmethod
+    def _construct_tag(name):
+        """ Construct the xml tag by prepending the dell tag prefix. """
+        return '{http://www.dell.com/ns/dell:0.1/root}%s' % name
+
+    def _make_request(self, method, url, data=None,
+                      acceptable_error_codes=()):
```
zenhack: afaik, you don't use this parameter anywhere. More copypasta from the brocade driver.

In the future, please don't copy-and-paste; if there's enough in common to warrant it, there's enough in common to factor it out into its own code.

zenhack:
This is much better. I left a number of other comments, now that it's easier to follow.

hil/ext/switches/dellnos9.py
naved001: I think it's doable within this PR. 

This is what `_remove_native_vlan` does:

```python
        command = 'interface vlan ' + vlan + '\r\n no tagged ' + \
            self.interface_type + ' ' + interface
```
that translates to 2 commands in one call to the switch. The commands are separated by \r\n
```
switch(conf)# interface vlan vlanno
switch(conf-if-vl-vlanno)# no tagged gigabitethernet 1/3
```

So for removing multiple vlans, I should be able to generate a huge command (separated by \r\n) and send that in one call. That would probably be faster.

> Also, how slow are we talking?

Both deployment tests take around 340 seconds to complete.

zenhack:
Quoting Naved Ansari (2017-09-12 10:22:59)

>    So for removing multiple vlans, I should be able to generate a huge
>    command (separated by \r\n) and send that in one call. That would
>    probably be faster.

Yeah, that would avoid doing so many round-trips. Sure, give it a shot.

hil/ext/switches/dellnos9.py
naved001: So, I did this. The deployment tests don't seem to be any faster. But I guess the speedup is proportional to part of program that can be parallelized.

naved001:
@zenhack I think addressed all the comments. 

@shwsun @SahilTikale can we get a second review here? Please ask any questions you have. Would be nice to have another exhaustive review.

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,380 @@
+# Copyright 2017 Massachusetts Open Cloud Contributors
```
geedo0: Do you need to consider renaming the hil.ext.switches.dell module to something more explicit now?

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,380 @@
...
+        # It uses the REST API CLI which is slow but it is the only way
+        # because the switch is VLAN centric. Doing a GET on interface won't
+        # return the VLANs on it, we would have to do get on all vlans (if that
+        # worked reliably in the first place) and then find our interface there
+        # which is not feasible.
+
+        response = self._get_port_info(interface).splitlines()
+
+        try:
+            index = response.index("Vlanmembership:") + 3
```
geedo0: The 3 here is kind of troubling. Can we add some more checks here to make sure you're parsing out the right thing? You can also do a tighter regex to match exactly the line with the "T". Bounds checking on the array would be nice as well to avoid an out of bounds exception.

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,380 @@
...
+            interface: interface to return the native vlan of
+
+        Returns: Tuple of the form ('vlan/native', vlan) or None
+
+        Similar to _get_vlans()
+        """
+        response = self._get_port_info(interface)
+        if response.find('NativeVlanId:') == -1:
+            return None
+        begin = response.find('NativeVlanId:') + len('NativeVlanId:')
```
geedo0: Use a regular expression to figure out if the native vlan exists and to get its id.

r'NativeVlanId:(\d+)\.'

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,380 @@
...
+
+        If the port is not trunked, its mode will be set to trunk.
+
+        Args:
+            interface: interface to add the vlan to
+            vlan: vlan to add
+        """
+        if not self._is_port_on(interface):
+            self._port_on(interface)
+        command = 'interface vlan ' + vlan + '\r\n tagged ' + \
```
geedo0: This driver constructs enough commands with the \r\n line endings that I think it's worthwhile to use a different pattern to reduce the bugs.

Something like this might work well
```
self._build_cmd = lambda cmd: '\r\n'.join(cmd)
```

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,380 @@
...
+            self.interface_type + ' ' + interface
+        self._execute(interface, CONFIG, command)
+
+    def _remove_native_vlan(self, interface):
+        """ Remove the native vlan from an interface.
+
+        Args:
+            interface: interface to remove the native vlan from.vlan
+        """
+        vlan = self._get_native_vlan(interface)[1]
```
geedo0: This will throw a TypeError if _get_native_vlan returns None

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,380 @@
...
+
+        Returns: string with the url for a specific interface and operation
+
+        If interface is None, then it returns the URL for REST API CLI.
+        """
+
+        if interface is None:
+            return '%s/api/running/dell/_operations/cli' % self.hostname
+
+        val = re.compile(r'^\d+/\d+(/\d+)?$')
```
geedo0: You can call validate_port_name(interface) here instead to improve reuse.

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,380 @@
...
+
+    def session(self):
+        return self
+
+    @staticmethod
+    def validate_port_name(port):
+        """Valid port names for this switch are of the form 1/0/1 or 1/2"""
+
+        val = re.compile(r'^\d+/\d+(/\d+)?$')
+        if not val.match(port):
```
geedo0: This syntax is cleaner since the regex pattern isn't reused.
```
if not re.match(r'^\d+/\d+(/\d+)?$', port):
```

hil/ext/switches/dellnos9.py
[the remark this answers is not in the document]
zenhack: Yeah, at some point we should re-name it to powerconnect. That gets into some not-entirely trivial issues with migrating existing setups though, so my inclination is to defer and open a separate issue.

docs/network-drivers.md
```diff
@@ -213,6 +213,40 @@ The body of the api call request will look like:
...
+The ``type`` field for the Dell Networking OS 9 driver has the value:
+
+    http://schema.massopencloud.org/haas/v0/switches/dellnos9
+
+In addition to ``type``, you need to supply ``username``, ``password``,
+``hostname``, and ``interface_type``. ``interface_type`` is of the form
+"GigabitEthernet", "TenGigabitEthernet". Please look at the switch guide for
+more valid interface types.
+
+The API server either runs on port 8008 (HTTP) or 8888 (HTTPS), so be sure to
```
zenhack: I would say the switch's API server to make this a bit more clear -- it took me a second to figure out you weren't talking about the HIL API server, and I was momentarily confused.

docs/network-drivers.md
```diff
@@ -213,6 +213,40 @@ The body of the api call request will look like:
...
+specify that in the ``hostname``.
+
+If you have multiple types of ports on the same switch, register the switch
+multiple times with different parameters for ``interface_type``.
```
zenhack: This is kindof gross, is it really much work to do this as port_register?

naved001: I thought this was reasonable and generally accepted since I got this from the brocade driver.

Doing this in port register means, I would have to read the whole interface_type + interface like `TenGigabitEthernet 1/3`. 

Anyway, making this change would require a significant amount of changes everywhere.

zenhack: Ug, okay, I guess we can live with it. I'm realizing I'm somewhat dissatisfied with the quality of the brocade driver...

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,380 @@
...
+            raise BadArgumentError("Invalid port name. Valid port names for "
+                                   "this switch are of the from 1/0/1 or 1/2")
+        return
+
+    def disconnect(self):
+        pass
```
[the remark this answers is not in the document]
naved001: I added a docstring to explain this.

In the base [Switch class](https://github.com/CCI-MOC/hil/blob/master/hil/model.py#L243) we explain that for switches that are not connection-oriented, we don't need to worry about setting up and/or disconnecting a session.

hil/ext/switches/dellnos9.py
```diff
@@ -0,0 +1,380 @@
...
+        if channel == 'vlan/native':
+            if new_network is None:
+                self._remove_native_vlan(interface)
+            else:
+                self._set_native_vlan(interface, new_network)
+        else:
+            vlan_id = channel.replace('vlan/', '')
+            legal = get_network_allocator(). \
+                is_legal_channel_for(channel, vlan_id)
+            assert legal, "HIL passed an invalid channel to the switch!"
```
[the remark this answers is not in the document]
naved001: True, it actually needs to be changed in the brocade too. That will be done in a separate PR though.

hil/ext/switches/dellnos9.py
naved001: Good call,  I should have been using regex all along.

hil/ext/switches/dellnos9.py
naved001: Could you explain how this reduces bugs? I avoid using lambas because they make the code slightly less readable imo.

docs/network-drivers.md
naved001: yeah, I'll avoid copy-pasting in the future.

naved001:
@geedo0 @zenhack 

pushed another commit addressing your changes.

hil/ext/switches/dellnos9.py
jeremyfreudberg: sorry to butt in, but please do not assign names to anonymous functions. if you want a named function use def.

naved001: @jeremyfreudberg feel free to butt in.

geedo0: Yeah, I was just being really lazy. My main point though was to define a function that would build these command sequences with the /r/n instead of constructing them inline. It would take in an array of commands and then return the string delimited by windows newlines.

hil/ext/switches/dellnos9.py
```diff
@@ -153,11 +147,9 @@ def _get_native_vlan(self, interface):
         Similar to _get_vlans()
         """
         response = self._get_port_info(interface)
-        if response.find('NativeVlanId:') == -1:
-            return None
-        begin = response.find('NativeVlanId:') + len('NativeVlanId:')
-        end = response.find('.', begin)
-        vlan = response[begin:end]
+        match = re.search(re.compile(r'NativeVlanId:(\d+)\.'), response)
+        vlan = re.search(re.compile(r'(\d+)'), match.group())
```
geedo0: We lost handling for the error condition where the NativeVlanId was not found. Also, the vlanid is already captured in the original regex so something like this should work.

```
match = re.search(r'NativeVlanId:(\d+)\.', response)
if match is not None:
    vlan = match.group(1)
else:
    return None
```

hil/ext/switches/dellnos9.py
```diff
@@ -128,18 +128,12 @@ def _get_vlans(self, interface):
...
-            index = response.index("Vlanmembership:") + 3
-        except ValueError:
-            # happens when our response has almost no info on a switchport
-            # because it's off. the response lacks the word "vlanmembership"
-            return []
-        if response[index] == '':
-            # a port with no tagged interface has no section with "T"
+        response = self._get_port_info(interface)
+        # finds a comma separated list of integers starting with "T"
+        match = re.search(re.compile(r'T(\d+)((,\d+)?)*'), response)
```
geedo0: You don't need the re.compile(), search will accept the regex string directly :)

hil/ext/switches/dellnos9.py
naved001: oh, I totally forgot about that haha. Good catch.

And, I was doing `match.group()[1]` which was just returning 'a'.

hil/ext/switches/dellnos9.py
naved001: @geedo0 Done!

zenhack:
LGTM, merging.

PR #867, opened 2017-09-07
Use Correct SQL Alchemy methods

### LINKED ISSUE #308: Use correct SQLAlchemy methods
We should be using .one(), .count() or .exists() in many cases where we are just using .first() for DB queries.

naved001:
* Closes #308 

Please see my [comment here](https://github.com/CCI-MOC/hil/issues/308#issuecomment-327642054) to see what I did here and why.

### DIFF
--- a/hil/api.py
+++ b/hil/api.py
@@ -453,7 +453,7 @@ def node_detach_network(node, nic, network):
         raise errors.BlockedError(
             "A networking operation is already active on the nic.")
     attachment = model.NetworkAttachment.query \
-        .filter_by(nic=nic, network=network).first()
+        .filter_by(nic=nic, network=network).one_or_none()
     if attachment is None:
         raise errors.BadArgumentError(
             "The network is not attached to the nic.")
--- a/hil/api.py
+++ b/hil/api.py
@@ -379,7 +379,7 @@ def node_connect_network(node, nic, network, channel=None):
         return model.NetworkAttachment.query.filter(
             model.NetworkAttachment.nic == nic,
             query,
-        ).count() != 0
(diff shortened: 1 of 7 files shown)


zenhack:
I'm not sure whether I agree with the count() -> first() change; we're also not interested in the contents of the tuple. What we really want is exists; was there a reason not to do that?

naved001:
> I'm not sure whether I agree with the count() -> first() change; we're also not interested in the contents of the tuple.

Let's say there were 10,726 matches (highly unlikely obviously), then count would take some time but first() would just return on the first match (worst case would be returning a `None`). I know this is a very unlikely scenario with little to no performance benefits, but that was my reasoning.

> What we really want is exists; was there a reason not to do that?

I couldn't find it in the docs. Let me dig around more and I'll use that.

naved001:
`exists()` seems needlessly complicated to me honestly. It makes the code less readable. 
I am wondering if I should even change it [here](https://github.com/CCI-MOC/hil/blob/master/hil/ext/network_allocators/vlan_pool.py#L57)

zenhack:
I actually think I agree with you re: exists; maybe just revert that commit. 

Re: the vlan thing, I think leave it as is.

naved001:
> I actually think I agree with you re: exists; maybe just revert that commit.

done.

naved001:
@knikolla @shwsun @Izhmash can either of you review this? Thanks!

naved001:
@SahilTikale want to review this? it's a small PR.

naved001:
Thanks @shwsun for looking at it! :)

PR #871, opened 2017-09-12
pylint: check for duplicate imports

zenhack:
Pylint was complaining about this in 3 places:

* A duplicate import of paths
* Arguably a false positive, we were importing two different mock
  drivers locally, for the side-effects only, under the same name.
  I changed these to import the modules fully-qualified, which satisfies
  pylint.
* hil.model.db was being imported locally in the migration tests, *in
  addition to* being imported at the top of the file. I removed the
  redundant local import. While there, I also moved some local imports
  of non-extensions to global scope, as these are safe to import at the
  top of the file.


naved001:
Looks good to me!

@djfinn14 can you review this?

zenhack:
Cool, merging.

PR #874, opened 2017-09-13
Show network now shows node and nic connected to network

### LINKED ISSUE #496: show_network() should display connected nodes
Currently, one has to dig through the DB or iterate through every node to figure out what's connected to a network.

This is needed for times like these:

```
$ haas network_delete random-network
Unexpected status code: 409
Response text:
{"msg": "Network still connected to nodes", "type": "BlockedError"}
```

If we showed connected nodes (+ maybe the channel, NIC,  switch and port name), this would allow all sorts of configuration goodness.
Maybe we could just include the output of `show_node()` for each node?

naved001:
* fixes #496 
* show_network will return a list of node,nic pairs that are connected to a
network. This is useful when a user wants to delete a network but can't delete it
because they don't know what nodes is the network connected to.

* added a small api test

* added a simple method in api tests, that can help reduce repepetitive code. 
* updated the REST API docs

### DIFF
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -200,7 +200,7 @@ The result must contain the following fields:
 * "owner", the name of the project which created the network, or
   "admin", if it was created by an administrator.
 * "access", a list of projects that have access to the network or null if the network is public
-* "node, nic": node and nic connected to network
+* "connected-nics": nodes and list of nics connected to network
 
 Response body (on success):
 
@@ -209,7 +209,7 @@ Response body (on success):
         "channels": <chanel-id-list>,
         "owner": <project or "admin">,
         "access": <project(s) with access to the network/null>
-        "node, nic": <list of node, nic>
+        "connected-nodes": {"<node>": [<list of nics connected to network]}
     }
 
 Authorization requirements:
@@ -217,6 +217,8 @@ Authorization requirements:
 * If the network is public, no special access is required.
 * Otherwise, access to a project in the "access" list or
 administrative access is required.
+* Admin access to view all connected-nodes is required; otherwise only the
+nodes that a non-admin user has access to will be shown.
 
 #### Channel Formats
 
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -200,7 +200,7 @@ The result must contain the following fields:
 * "owner", the name of the project which created the network, or
   "admin", if it was created by an administrator.
 * "access", a list of projects that have access to the network or null if the network is public
-* "connected-nics": nodes and list of nics connected to network
+* "connected-nodes": nodes and list of nics connected to network
 
 Response body (on success):
 
@@ -217,8 +217,8 @@ Authorization requirements:
 * If the network is public, no special access is required.
 * Otherwise, access to a project in the "access" list or
 administrative access is required.
-* Admin access to view all connected-nodes is required; otherwise only the
-nodes that a non-admin user has access to will be shown.
+* Admins and network owners can see all nodes connected to network; other users
+only see connected nodes that they have access to.
 
 #### Channel Formats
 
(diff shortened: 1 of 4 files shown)


docs/rest_api.md
```diff
@@ -208,6 +209,7 @@ Response body (on success):
         "channels": <chanel-id-list>,
         "owner": <project or "admin">,
         "access": <project(s) with access to the network/null>
+        "node, nic": <list of node, nic>
```
zenhack: Maybe change the form of this a bit; the name "node, nic" smells funny, and using a list for this is a little weird. You could rename the field to "connected-nics" or something and make the value a dictionary mapping node -> list of nics, e.g.

```json
...
"connected-nodes": {
    "node-1": ["eth0", "pxe"],
    "node-72": ["eth1"]
}
...
```

naved001: yeah, I actually wanted to use tuples, but it took me a while to realize that tuples aren't a thing in json.

I'll make this change.

naved001:
> I think it's worth discussing how broadly we want this information to be available; right now it's anyone who can show_network; we may want to e.g. restrict this to the owner, since they are the only ones who need this information for the already-described use cases,

Good point. So we should show this information to admins (obviously), and for regular users with access to the network we could do 1) show all nodes connected to that network. 2) show only their nodes that are on that network.

naved001:
I went ahead and did 2)

Now I need to format the output in case multiple nics on one node are connected to the same network. And update the tests.

Izhmash:
We'll want a test to make sure that users cannot access each other's nodes.

naved001:
> We'll want a test to make sure that users cannot access each other's nodes.

@Izhmash Done!

@zenhack How does this look now?

naved001:
@zenhack @Izhmash ping

hil/api.py
```diff
@@ -895,6 +894,17 @@ def show_network(network):
     else:
         result['access'] = None
 
+    x = {}
```
zenhack: More descriptive variable name. Maybe `connected nodes`?

docs/rest_api.md
```diff
@@ -200,6 +200,7 @@ The result must contain the following fields:
 * "owner", the name of the project which created the network, or
   "admin", if it was created by an administrator.
 * "access", a list of projects that have access to the network or null if the network is public
+* "connected-nics": nodes and list of nics connected to network
```
zenhack: Should be `"connected-nodes"`.

naved001:
@zenhack 
> Also, what happens if a user has shared a network with other projects, but can't delete it because other projects still have nodes on that network? right now the user would not be able to see them.

good point. I think network owners should also be able to see what nodes are connected to their network (pushed that in the most recent commit)

 But what would a network owner do with this information?
Network owners can't revoke access unless all nodes are disconnected from that network, and it's not possible to remove someone else's nodes from your network. So they are stuck and can't delete a network they own.
At least, they can see that there are some nodes on their network and then tell an admin to do something.

Or maybe we should have a `revert_network()` that removes all nics/nodes from a network. And that can only be run by network owners or admins.

Izhmash:
I think the `revert_network()` call would be a good idea.  Or better yet, I wonder if giving the user who owns a private network the power to disconnect other users' nodes would be appropriate.  It would make the private owner a kind of 'sub-admin' of the private network.

zenhack:
I think for the time being the thing to do is exactly what this pr is doing -- let the owner see things, but no other changes. This at least gives them the information they need to proceed.

There are other improvements tgat could be made, but they're out of scope for this pr.

naved001:
@zenhack so is there anything else that I can do in this PR to get your approval?

zenhack:
LGTM, merging.

Issue #410, opened 2015-04-07
Mechanism for expressing network driver capabilities


henn:
For work like what @zenhack is doing in #408, it would be good to have a mechanism for registering the optional capabilities for drivers. One example would be for a driver to advertise its ability to do channels, where multiple networks can be communicated over a single NIC/port.

As HaaS grows and has more code & contributors, it may not always be feasible to support new features (like channels) in all existing drivers.

Other capabilities could include the bridging of networks between HaaS instances (#409), QoS options or enabling jumbo frames.

Updated:   
**Examples include:**
- Jumbo frames are per-switch on our Dell switches and per-port on our Ciscos (#755)
-  QoS (#384)
- Disabling STP (#779)
- DHCP/ARP Snooping (#471)
- Enabling the tagging of DHCP packets with the port number (from @okrieg or @pjd-nu)

henn:
CC @naved001 @zenhack

henn:
Examples include:
- Jumbo frames are per-switch on our Dell switches and per-port on our Ciscos
-  QOS
- Disabling STP (#779)
- DHCP/ARP Snooping (#471)
- Enabling the tagging of DHCP packets with the port number (from @okrieg or @pjd-nu)

SahilTikale:
@shwsun @naved001 @Izhmash @zenhack 
We discussed this how do we wish to expose to the user, what different services HIL provides for a given switch. Currently switches are not exposed to the users. They only know about nic and networks. 
We think we should augment the one of the calls `show nic` or `show network` where we can list what other services HIL provides for that given switch.

zenhack:
To clarify/expand a bit on @SahilTikale's summary:

* The obvious thing would be to just query the switch, but the switch topology is currently only exposed to admins.
* We've talked about wanting to expose topology to the users in the future; one that happens we should probably move this information there.
* We *can* report this information via the affected visible objects, i.e. networks and NICs
* There was some disagreement about whether networks or NICs made more sense. I think NICs make more sense; with networks you have the problem of possibly needing to do multi-switch queries to determine the answer to a question. @SahilTikale said he thought it was better to do it on the network. We didn't want to get bogged down with that argument during the meeting. @SahilTikale, do you want to make the case for doing it on the network here?

SahilTikale:
@naved001 @Izhmash and I are in the meeting today. 
[...]

PR #876, opened 2017-09-15
Spec for expressing switch driver capabilities

naved001:
Making the first spec here for Issue #410

### DIFF
--- a/specs/switch-driver-capabilities.md
+++ b/specs/switch-driver-capabilities.md
@@ -15,16 +15,15 @@ We want to expose switch driver capabilities. Capabilities could include:
 # Solution
 -----------
 
-* All switch drivers would have a method called get_capabilities.
+* All switch drivers would have a method called `get_capabilities()`.
 
 * This method would return a list with all capabilities supported by the switch.
 Its implementation could be flexible; it could simply return a hardcoded list, or generate it from some database.
 
-* The superclass will have a method `has_capbility(self, name)` that would return
-`name in self.get_capabilities()`. Drivers can overwrite if need be.
 
-* an API that needs to check for capabilities will call `has_capability()` and
-proceed based on the response.
+* method `ensure_legal_operations()` will be updated to check for capabilities.
+Currently, this method is called by the API before anything is queued for the networking
+daemon.
 
 * `show_switch()` will be updated to show switch capabilities.
 
--- a/specs/switch-driver-capabilities.md
+++ b/specs/switch-driver-capabilities.md
@@ -1,0 +1,48 @@
+
+# Expose network driver capabilites
+-----------
+
+[Issue 410](https://github.com/CCI-MOC/hil/issues/410)
+
+We want to expose switch driver capabilities. Capabilities include:
+
+- Jumbo frames are per-switch on our Dell switches and per-port on our Ciscos (#755)
+-  QoS (#384)
+- Disabling STP (#779)
+- DHCP/ARP Snooping (#471)
+- Enabling the tagging of DHCP packets with the port number (from @okrieg or @pjd-nu)
+
+# Solution
+-----------
+
+* Every switch driver will have a list called `capabilities` in the class. So it would look like
+`capabilities = ["require-native-network", "per-port-jumbo-frames"]`
+
+* and then there'll be a method  `get_capabilities()` that returns a boolean if a capability exists or not by checking it in the list.
+
+* an API that needs to test this, can call that method `switch.get_capabilities('require-native-network`) to ensure that the switch supports that method, and then proceed with the original request or error out.
+
+* `show_node()` can be updated to show the capabilities they support.
+
+# Alternative Solutions
+-----------------------
+
+@henn proposed that we store the capabilites in a database and those will be mapped
+to switch objects. That way we can control indiviual switches instead of a whole
+class of switches.
+
+Rest of the steps would remain the same.
+
+# Arch Impact
+-----------------
+
+Are there any new architectural assumptions we are now making?
+I am making an assumption that we would treat all switches that belong to a class
+the same. That is, all brocade switches would have those capabilities as specified
+in the Class.
+
+# Security
+----------
+
+Doesn't look like it.
+```
--- a/specs/switch-driver-capabilities.md
+++ b/specs/switch-driver-capabilities.md
@@ -36,15 +36,19 @@ a new switch capability.
 
 # Alternative Solutions
 -----------------------
-
+We discussed having another method called `has_capability()` that would return if a
+capability exists in the list returned by `get_capabilities()`. But at this point
+it seems redundant, because `ensure_legal_operations()` can just do the check by itself,
+instead of requiring a new method.
 
 # Arch Impact
 -----------------
 
-None
+Doesn't affect the current architecture, but opens up the debate how network topologies might
+look like if a network spans across switches with different capabilities.
 
 # Security
 ----------
 
-Doesn't look like it.
+This doesn't seem to have any impact on security.
 ```
--- a/specs/switch-driver-capabilities.md
+++ b/specs/switch-driver-capabilities.md
@@ -15,25 +15,22 @@ We want to expose switch driver capabilities. Capabilities could include:
 # Solution
 -----------
 
-* Every switch driver will have a list called `capabilities` in the class. So it would look like
-`capabilities = ["require-native-network", "per-port-jumbo-frames"]`
+* All switch drivers would have a method called get_capabilities.
 
-* A new column (text) will be added to the switch table. This will store list of
-capabilities encoded as JSON. This could probably default to all capabilities as
-specified in the Class.
+* This method would return a list with all capabilities supported by the switch.
+It's implementation could be flexible; it could simply return a hardcoded list, or generate it from some database.
 
-* A method called `get_capabilities()` would return all capabilities for a switch
-
-* and then there'll be a method  `have_capability()` that returns a boolean if a capability exists or not by checking it in the list.
-
-* an API that needs to test this, can call that method `switch.get_capabilities('require-native-network`) to ensure that the switch supports that method, and then proceed with the original request or error out.
+* an API that needs to check for capabilities, can call `get_capabilties()` and then
+check if it exists in the list or not.
 
 * `show_switch()` will be updated to show switch capabilities.
 
 * `show_node()` can be updated to show the capabilities supported by a nic. This
-will be useful for end users to pick nodes based on capabilities.
+will be useful for end users to pick nodes based on capabilities. Admins would
+see all capabilities while non-admin users would only see certain limited capabilities.
+What to show and what not show under this can be decided as we add support for
+a new switch capability.
 
-*
 
 # Alternative Solutions
 -----------------------
--- a/specs/switch-driver-capabilities.md
+++ b/specs/switch-driver-capabilities.md
@@ -18,28 +18,31 @@ We want to expose switch driver capabilities. Capabilities include:
 * Every switch driver will have a list called `capabilities` in the class. So it would look like
 `capabilities = ["require-native-network", "per-port-jumbo-frames"]`
 
-* and then there'll be a method  `get_capabilities()` that returns a boolean if a capability exists or not by checking it in the list.
+* A new column (text) will be added to the switch table. This will store list of
+capabilities encoded as JSON. This could probably default to all capabilities as
+specified in the Class.
+
+* A method called `get_capabilities()` would return all capabilities for a switch
+
+* and then there'll be a method  `have_capability()` that returns a boolean if a capability exists or not by checking it in the list.
 
 * an API that needs to test this, can call that method `switch.get_capabilities('require-native-network`) to ensure that the switch supports that method, and then proceed with the original request or error out.
 
-* `show_node()` can be updated to show the capabilities they support.
+* `show_switch()` will be updated to show switch capabilities.
+
+* `show_node()` can be updated to show the capabilities supported by a nic. This
+will be useful for end users to pick nodes based on capabilities.
+
+*
 
 # Alternative Solutions
 -----------------------
 
-@henn proposed that we store the capabilites in a database and those will be mapped
-to switch objects. That way we can control indiviual switches instead of a whole
-class of switches.
-
-Rest of the steps would remain the same.
 
 # Arch Impact
 -----------------
 
-Are there any new architectural assumptions we are now making?
-I am making an assumption that we would treat all switches that belong to a class
-the same. That is, all brocade switches would have those capabilities as specified
-in the Class.
+None
 
 # Security
 ----------
--- a/specs/switch-driver-capabilities.md
+++ b/specs/switch-driver-capabilities.md
@@ -18,17 +18,20 @@ We want to expose switch driver capabilities. Capabilities could include:
 * All switch drivers would have a method called get_capabilities.
 
 * This method would return a list with all capabilities supported by the switch.
-It's implementation could be flexible; it could simply return a hardcoded list, or generate it from some database.
+Its implementation could be flexible; it could simply return a hardcoded list, or generate it from some database.
 
-* an API that needs to check for capabilities, can call `get_capabilties()` and then
-check if it exists in the list or not.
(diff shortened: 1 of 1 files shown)


specs/switch-driver-capabilities.md
```diff
@@ -0,0 +1,48 @@
...
+- DHCP/ARP Snooping (#471)
+- Enabling the tagging of DHCP packets with the port number (from @okrieg or @pjd-nu)
+
+# Solution
+-----------
+
+* Every switch driver will have a list called `capabilities` in the class. So it would look like
+`capabilities = ["require-native-network", "per-port-jumbo-frames"]`
+
+* and then there'll be a method  `get_capabilities()` that returns a boolean if a capability exists or not by checking it in the list.
```
zenhack: Nitpick: maybe call it `have_capability`; `get_capabilities` suggests to me that it would return the list itself.

specs/switch-driver-capabilities.md
```diff
@@ -0,0 +1,48 @@
...
+
+Rest of the steps would remain the same.
+
+# Arch Impact
+-----------------
+
+Are there any new architectural assumptions we are now making?
+I am making an assumption that we would treat all switches that belong to a class
+the same. That is, all brocade switches would have those capabilities as specified
+in the Class.
```
zenhack: If we don't make the list itself public, and only specify the `have_capability` method, drivers could report the capability based on other information stored about the switch, so they wouldn't necessarily have to return the same value for all switches. That makes it unclear how we would list the caps in `show_node` though. Maybe we can just have the method `get_capabilities`, that returns the list, but not actually make the list itself public?

specs/switch-driver-capabilities.md
```diff
@@ -0,0 +1,48 @@
...
+-----------
+
+* Every switch driver will have a list called `capabilities` in the class. So it would look like
+`capabilities = ["require-native-network", "per-port-jumbo-frames"]`
+
+* and then there'll be a method  `get_capabilities()` that returns a boolean if a capability exists or not by checking it in the list.
+
+* an API that needs to test this, can call that method `switch.get_capabilities('require-native-network`) to ensure that the switch supports that method, and then proceed with the original request or error out.
+
+* `show_node()` can be updated to show the capabilities they support.
```
zenhack: Ideally we'd have a `show_switch` that would report this information. Might be worth considering whether we want to just add that first.

naved001: True. `show_switch` should show this information. But I think we should show this information under `show_node()` just so that users can pick nodes with capabilities they require.

specs/switch-driver-capabilities.md
naved001: > drivers could report the capability based on other information stored about the switch

where would this other information be stored? in the switch table?

> Maybe we can just have the method get_capabilities, that returns the list, but not actually make the list itself public?

yep, and that will be called by `show_node()`

zenhack: Yeah, probably in a driver-specific table.

naved001: Hmm, in that case we would need multiple tables. A table to store switch info (we already have that), a table to store all possible switch capabilities, and then a table to map capabilities to switches.

``` SQL
CREATE TABLE Switches(SwitchID integer PRIMARY KEY, SwitchName text);

/* Create few records in this table */
INSERT INTO Switches VALUES(1,'Brocade');
INSERT INTO Switches VALUES(2,'Dell');

CREATE TABLE AllSwitchCapabilities(CapID integer PRIMARY KEY, Capability text);

INSERT INTO AllSwitchCapabilities Values(1,'jumbo');
INSERT INTO AllSwitchCapabilities Values(2,'native');
INSERT INTO AllSwitchCapabilities Values(3,'something');

CREATE TABLE SwitchCapability(ID integer PRIMARY KEY, SwitchID int, CapID int);
INSERT INTO SwitchCapability Values(1, 2, 1);
INSERT INTO SwitchCapability Values(2, 2, 2);
```
and then the query to list all capabilities would look like this:

```SQL
SELECT SwitchName, Capability from Switches, AllSwitchCapabilities, SwitchCapability where
Switches.SwitchID == SwitchCapability.SwitchID and SwitchCapability.CapID == 
AllSwitchCapabilities.CapID

```

Is there a simpler way to do this?

zenhack: Depending on the switch, you could just have a column of type text that stores the list encoded as JSON. The point is that implementations can do whatever makes sense for them.

naved001:
@shwsun @knikolla thoughts?

specs/switch-driver-capabilities.md
zenhack: I think you misunderstood what I was getting at. The notion is that get_capabilities would be implemented by the switch driver. Nothing needs to be added to the Switch table; if schema changes occur they would be isolated to the driver. But the point is it's an implementation detail; by making it a function the driver can do whatever it wants to create that list. The driver *could* store the info in the database (under one of its own tables), or it could just return a hard-coded list (probably makes sense for most drivers), or compute it some other way.

naved001: > The notion is that get_capabilities would be implemented by the switch driver.

Yes, I agree. And like you said, it could get/generate this list from a database, or a hardcoded list in the Class.

What I am trying to address is that, let's say I have 2 switches of type brocade, and for whatever reason I want to enable a certain capability only on one of them. 

> or it could just return a hard-coded list (probably makes sense for most drivers),

If I hard-code this list in the switch driver, then I can't differentiate between the capabilities of 2 different switches of the same type. So I must store it somewhere, right? And I thought keeping it as JSON in the db made sense.

Or should we not expose the per-switch capability in HIL? and enforce same capabilities if they belong to the same Class.

specs/switch-driver-capabilities.md
```diff
@@ -0,0 +1,51 @@
...
+
+# Solution
+-----------
+
+* Every switch driver will have a list called `capabilities` in the class. So it would look like
+`capabilities = ["require-native-network", "per-port-jumbo-frames"]`
+
+* A new column (text) will be added to the switch table. This will store list of
+capabilities encoded as JSON. This could probably default to all capabilities as
+specified in the Class.
```
henn: If we don't support a way for modifying the capabilities per-switch yet, why not just keep this in the driver class and out of the DB?

naved001: Yeah right. The question was should we support per-switch yet or not. If not, then we can just keep the capabilities in the class.

henn: Ya. And once we got per-switch, it might make sense to have a separate table since the relation is many to many.

henn:
Reiterating my above question (since I made it in an edit): are there capabilities that we might want to hide from end-users? For example, the user might not need to know that a port gets shut off on a particular switch.

specs/switch-driver-capabilities.md
zenhack: I think you misunderstood what I was proposing. The column is just one possible implementation strategy. The point is that, if we limit the interface to `get_capabilities`, the driver can come up with the list however it wants, whereas if we have a static attribute on the class, it's stuck leaving it the same for all switches. For an individual driver, it may not actualy make sense to store any extra information.

Also, the notion was to add a column to the *driver's* table, e.g. if the powerconnect driver wanted to track this information, it would add a column to a table defined in that extension. The table defined in hil.model doesn't change.

naved001:
> are there capabilities that we might want to hide from end-users?

I think, to begin with, we should only expose limited capabilities to end users. 

And things like DHCP/ARP snooping should only be visible to admins. Actually, I don't know if they should be exposed at all since they seem like one time configuration changes. We already recommend a bunch of config settings to make when setting up a switch for HIL anyway.

>  the user might not need to know that a port gets shut off on a particular switch.

This is not a switch feature, right? So I don't understand how is this related to switch capabilities. But yeah, the end users don't need to be aware about switchport status.

henn:
> And things like DHCP/ARP snooping should only be visible to admins

I envisioned DHCP/ARP snooping being exposed to the HIL tenant, since tenants could select which ports they want to be enabled for DHCP serving. I expect there will be an associates API calls for `enable_arp_snooping()` and `enable/disable_dhcp_server()` that takes a host, NIC and network argument.

This actually brings up an interesting point -- DHCP/ARP snooping support can differ based on the switch, but it's something that would only make sense to have globally enabled on all the switches or none of them.

How do folks think we should do that?

One way would be to have that capability enabled globally (either through an API call that gets stored in the DB or via the config file), then reject any switch registrations for switches that didn't support that feature.

Other capabilities (like disabling STP) might be hidden though.

specs/switch-driver-capabilities.md
henn: Just catching up on this thread.

I agree with @zenhack - having driver API calls gives us a lot of flexibility in how we keep the backend. Please see my comments in [the code](https://github.com/CCI-MOC/hil/pull/876#discussion_r140098278). Also, by giving it an API, the switch driver can do whatever it wants. For now, we can keep a hard-coded list in the class, and in the future we can use a per-driver or global DB table to make it per-switch instantiation.

specs/switch-driver-capabilities.md
```diff
@@ -0,0 +1,48 @@
...
+-----------
+
+* All switch drivers would have a method called get_capabilities.
+
+* This method would return a list with all capabilities supported by the switch.
+It's implementation could be flexible; it could simply return a hardcoded list, or generate it from some database.
+
+* an API that needs to check for capabilities, can call `get_capabilties()` and then
+check if it exists in the list or not.
+
```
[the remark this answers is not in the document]
zenhack: I'm fine putting has_capability in there. we could supply a default implementation in the superclass that's just:

```python
def has_capability(self, name):
    return name in self.get_capabilities()
```

...Which implementations could overwrite if they want to, or just leave as-is.

I agree with @naved001 re: a global list of caps.

specs/switch-driver-capabilities.md
```diff
@@ -0,0 +1,48 @@
...
+- DHCP/ARP Snooping (#471)
+- Enabling the tagging of DHCP packets with the port number (from @okrieg or @pjd-nu)
+
+# Solution
+-----------
+
+* All switch drivers would have a method called get_capabilities.
+
+* This method would return a list with all capabilities supported by the switch.
+It's implementation could be flexible; it could simply return a hardcoded list, or generate it from some database.
```
zenhack: s/It's/Its/

naved001:
@zenhack  @henn: updated with your recent suggestions.

>  This actually brings up an interesting point -- DHCP/ARP snooping support can differ based on the switch, but it's something that would only make sense to have globally enabled on all the switches or none of them.
> 
> How do folks think we should do that?
> 
> One way would be to have that capability enabled globally (either through an API call that gets stored in the DB or via the config file), then reject any switch registrations for switches that didn't support that feature.

Putting it in the config file sounds good to me.

specs/switch-driver-capabilities.md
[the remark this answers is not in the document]
henn: Do we really need get_capabilities() if a driver has has_capability() ?

henn:
Do we really need get_capabilities() now that we have has_capability() ?
I don't feel too strongly on this, but I think we should be intentional.

It's always better not to commit to an API until we're sure we need it, since then we'll have to support it forever.

zenhack:
@henn, show_node wants to list all of the capabilities. Imagine a driver that fishes these out of the database: implementing get_capabilities in terms of have_capability means doing a round-trip to the db for every potential element of the list.

I don't have an objection to only supporting one of these, but I think if we do that get_capabilities should be the primitive.

henn:
> @henn, show_node wants to list all of the capabilities. Imagine a driver that fishes these out of the database: implementing get_capabilities in terms of have_capability means doing a round-trip to the db for every potential element of the list.

Ya - that might be worth having both. I had been thinking that the common case would be has_capability(), which could eventually be implemented as an SQLAlchemy *exists()* call when we move the capabilities into the DB. *Exists()* would likely be much faster/more efficient than fetching + collating all of the entries in the table.

zenhack:
The point of hiding this behind a method at all is to keep the implementation independent from the interface -- using `.exists()` assumes DB storage, which may not be the case for all drivers.

naved001:
@henn and @zenhack 

so what's the consensus? We are going with `get_capabilities()` then?

zenhack:
`get_capabilities()` would be my preference. @henn?

specs/switch-driver-capabilities.md
```diff
@@ -0,0 +1,51 @@
...
+
+* All switch drivers would have a method called get_capabilities.
+
+* This method would return a list with all capabilities supported by the switch.
+Its implementation could be flexible; it could simply return a hardcoded list, or generate it from some database.
+
+* The superclass will have a method `has_capbility(self, name)` that would return
+`name in self.get_capabilities()`. Drivers can overwrite if need be.
+
+* an API that needs to check for capabilities will call `has_capability()` and
```
naved001: now that we have `ensure_legal_operation()` in the switches, I think `has_capability()` becomes redundant. capability checks can also be done be in `ensure_legal_operation()`

zenhack: We've already established that get_capabilities is sufficient. We can leave has_capability out and add it later if we want.

naved001: coo. I got rid of it, and rebased this PR.

specs/switch-driver-capabilities.md
```diff
@@ -0,0 +1,50 @@
...
+
+* `show_node()` can be updated to show the capabilities supported by a nic. This
+will be useful for end users to pick nodes based on capabilities. Admins would
+see all capabilities while non-admin users would only see certain limited capabilities.
+What to show and what not to show under this can be decided as we add support for
+a new switch capability.
+
+
+# Alternative Solutions
+-----------------------
```
zenhack: maybe use this section to summarize the debate we had re: has_capability? might be good to have for posterity.

specs/switch-driver-capabilities.md
```diff
@@ -0,0 +1,50 @@
...
+
+# Arch Impact
+-----------------
+
+None
+
+# Security
+----------
+
+Doesn't look like it.
```
zenhack: Empty sections like these make me somewhat dubious of the template we've drawn up. Regardless, if we're actually going merge this into the tree, I'd feel a little weird just having this not filled out. Though I don't think there's really anything else that needs discussion.

specs/switch-driver-capabilities.md
```diff
@@ -0,0 +1,54 @@
...
+-----------
+
+* All switch drivers would have a method called `get_capabilities()`.
+
+* This method would return a list with all capabilities supported by the switch.
+Its implementation could be flexible; it could simply return a hardcoded list, or generate it from some database.
+
+
+* method `ensure_legal_operations()` will be updated to check for capabilities.
+Currently, this method is called by the API before anything is queued for the networking
```
naved001: I think I might be wrong about this. Any other switch capability we add will be exposed via a different API. `ensure_legal_ops()` is only called by `node_connect/detach_network()`. So I can leave `ensure_legal_ops()` as it is. Any other API would just do `capability in switch.get_capabilities()`.

zenhack: Yeah, I think this bullet point probably isn't quite right.

naved001: done. now that I think of it, the spec thing for this is a bit of an overkill. I think of adding at least one capability when introducing this (perhaps jumbo frames on the brocade).

naved001:
sent an email to @henn, if he doesn't respond by today I'll dismiss his review and get someone else to review.

zenhack:
No response from @henn, so I've dismissed the review.

zenhack:
@Izhmash, would you do a second review?

Izhmash:
After some clarifications by @naved001, this looks good to me.

okrieg:
One comment, the only capabilities that should be exposed are capabilities needed to interact with HIL, e.g., capability of jumbo frames, or capability of supporting no native VLANs; these are capabilities that depend on both the nature of the hardware and how HILs drivers interact with it.  What you don't want is to make HIL the source of truth for the characteristics of the hardware.  We want discovery services.. for that.

PR #879, opened 2017-09-20
Add revert port call in the client library

naved001:
* Also add a unit test for it

Realized that I had it only in one of my feature branches and we kinda need this for developing HIL client.


SahilTikale:
Change the corresponding CLI function to use it.

naved001:
@SahilTikale Done

PR #881, opened 2017-09-21
Add mechanism to handle switch capabilities

naved001:
* the first capability introduced here is the ability to have no native vlan for a trunk port.
* modified the network allocator backend to return a native network
* add checks in the api to to ensure that a native network is always set for a
switch that requires it
* modified all switch drivers except for brocade (that requires more work).

### DIFF
--- a/hil/api.py
+++ b/hil/api.py
@@ -382,6 +382,7 @@ def node_connect_network(node, nic, network, channel=None):
         ).first() is not None
 
     auth_backend = get_auth_backend()
+    allocator = get_network_allocator()
 
     node = _must_find(model.Node, node)
     nic = _must_find_n(node, model.Nic, nic)
@@ -392,8 +393,6 @@ def node_connect_network(node, nic, network, channel=None):
     auth_backend.require_project_access(node.project)
 
     project = node.project
-
-    allocator = get_network_allocator()
 
     if nic.port is None:
         raise errors.NotFoundError("No port is connected to given nic.")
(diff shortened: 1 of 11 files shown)


naved001:
contingent on #876

hil/api.py
```diff
@@ -420,6 +419,19 @@ def _have_attachment(nic, query):
         raise errors.BadArgumentError(
             "Channel %r, is not legal for this network." % channel)
 
+    # First, check if the switch supports not having any native networks.
+    # then check if the passed channel is native, if yes, then check in the db
+    # if there's a native network already in use.
+    switch = nic.port.owner
+    native_channel = allocator.get_native_channel()
```
knikolla: You can only use methods of the network allocator which are part of the base class `hil.network_allocator.NetworkAllocator`. The way it is not it will break for all other network allocators that don't define it.

It would actually be "ok" to put this inside the if condition below, **but** we must clearly document that there is now a dependency that **all** switches that don't have the capability to not require a native network must be used with an allocator that has the `get_native_channel()` method.

But then, I don't like that this introduces a requirement on **all** switches that don't know anything about native networks, to declare a capability that they don't require one.

hil/ext/network_allocators/null.py
```diff
@@ -52,6 +52,9 @@ def claim_network_id(self, net_id):
     def is_network_id_in_pool(self, net_id):
         return True
 
+    def get_native_channel(self):
```
knikolla: No

hil/network_allocator.py
```diff
@@ -87,6 +87,10 @@ def claim_network_id(self, net_id):
     def is_network_id_in_pool(self, net_id):
         """returns true if net_id is part of the allocation pool"""
 
+    @abstractmethod
+    def get_native_channel(self):
```
knikolla: No

hil/api.py
naved001: > must be used with an allocator that has the get_native_channel() method.

PR #882, opened 2017-09-22
Introduce a method in the switch drivers that check if an operation i…

naved001:
…s legal

* this method is called by the API right before it queues any networking
action.

* for dell nos9 based switches, the checks ensures that a native network is
always attached first before any trunked vlans, and the native network is
also the last network to be removed.

* also turn off the port when revert_port is called.

* update get_port_info to turn on the port before querying it.

### DIFF
--- a/hil/ext/switches/dellnos9.py
+++ b/hil/ext/switches/dellnos9.py
@@ -23,8 +23,9 @@ import re
 import requests
 import schema
 
+from hil import model
 from hil.model import db, Switch, SwitchSession
-from hil.errors import BadArgumentError
+from hil.errors import BadArgumentError, BlockedError
 from hil.model import BigIntegerType
 from hil.network_allocator import get_network_allocator
 
@@ -61,6 +62,24 @@ class DellNOS9(Switch, SwitchSession):
     def session(self):
         return self
 
+    def ensure_legal_operation(self, nic, op_type, channel):
+        table = model.NetworkAttachment
+        if channel != 'vlan/native' and op_type == 'connect' and \
+            db.session.query(table).filter(table.nic_id == nic.id) \
+                .filter(table.channel == 'vlan/native').count() == 0:
+            # checks if it is trying to attach a trunked network, and then in
+            # in the db see if nic does not have any networks attached natively
+            raise BlockedError("Please attach a native network first")
+        elif channel is None and op_type == 'detach' and \
+            db.session.query(table).filter(table.nic_id == nic.id) \
+                .filter(table.channel != 'vlan/native').count() > 0:
+            # if it is detaching a network, then check in the database if there
+            # are any trunked vlans.
+            raise BlockedError("Please remove all trunked Vlans"
+                               " before removing the native vlan")
+        else:
+            return
+
     @staticmethod
     def validate_port_name(port):
         """Valid port names for this switch are of the form 1/0/1 or 1/2"""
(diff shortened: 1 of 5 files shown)


hil/model.py
```diff
@@ -246,6 +246,18 @@ def session(self):
         and have ``session`` just return ``self``.
         """
 
+    def ensure_legal_operation(self, nic, op_type, channel):
+        """Checks with the switch if the operation is legal before queueing it.
+
+        channel is network channel
+        interface is Port object
```
naved001: uh, need to uipdate this. "nic is Nic object"

hil/ext/switches/dellnos9.py
```diff
@@ -61,6 +62,24 @@ def validate(kwargs):
     def session(self):
         return self
 
+    def ensure_legal_operation(self, nic, op_type, channel):
+        table = model.NetworkAttachment
+        if channel != 'vlan/native' and op_type == 'connect' and \
+            db.session.query(table).filter(table.nic_id == nic.id) \
+                .filter(table.channel == 'vlan/native').count() == 0:
```
knikolla: this overly complex db query is performed twice. make a private helper method with a descriptive and meaningful name.

Izhmash:
We'll want to update unit and deployment tests for this, perhaps in a separate PR.

hil/ext/switches/dellnos9.py
naved001: the queries are slightly different though. One checks for `channel = 'vlan/native'` while the other one checks for the opposite. The only thing I could put in a helper method is the first part of the query where it finds the `nic_id` in table. Should I do that then?

Izhmash:
The logic of checking for native vlan presence looks sound to me, plus the logic for powering on and shutting off the ports.  We'll just need to make sure that when the tests call `_get_port_info`, if a port powers on, that it gets shut back off during cleanup.

naved001:
@Izhmash yeah, the deployment tests for native_networks (`tests/deployment/native_networks.py`) run fine for this, but the ones for vlan_networks don't because those tests use tagged networks without setting a native one. Fixing that would require more thought.

naved001:
>  We'll just need to make sure that when the tests call _get_port_info, if a port powers on, that it gets shut back off during cleanup.

When the tests finish, they remove all the networks (or call revert_port) and when all networks are removed, the switchport is shutoff.

naved001:
> Also make a test for the api that correctly returns the error message to the user in case of a raised exception.

Done.

hil/ext/switches/dellnos9.py
```diff
@@ -179,7 +201,8 @@ def _get_port_info(self, interface):
         \r\n\r\n Native Vlan Id: 1512.\r\n\r\n\r\n\r\n
         MOC-Dell-S3048-ON#</command>\n</output>\n"
         """
-
+        if not self._is_port_on(interface):
+            self._port_on(interface)
```
knikolla: Leaving it on after getting the information doesn't seem correct.

naved001: okay. I'll go without turning on the port at all. If the switchport is off that means there are no trunked or native vlans on it anyway.  (All VLANs must be removed from a switchport before it can be turned off)

naved001:
@zenhack @henn @SahilTikale Can I get your thoughts on this PR? At least on the idea. Thanks!

naved001:
> The only thing I would add is a comment on why this solves both the race condition on between port turn on and assigning a native vlan

A port is turned on only when setting a native vlan, so those 2 actions are bundled together in the networking action queue. And since the networking action queue is serialized, ports are turned on serially too which prevents the condition of two nodes sitting on the same default native network.

PR #884, opened 2017-09-25
refactor api tests (just a tad bit)

naved001:
* replace the calls to api.reigster_node() with the method `new_node()`. This
removes some repeated code since the api call was like 4 lines each.
 
There might be more places where we can do something like this, this one was the most obvious. EZ PZ


naved001:
I have purposely left some places with regular api node register calls because I thought it was important to be a bit more explicit there.

PR #886, opened 2017-09-26
Fix documentation

vsemp:
I went through rest_api.md documentation and checked that it matches api.py. I suggest few minor changes to the documentation.

### DIFF
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -421,8 +421,8 @@ Request Body:
              "type": <obm-subtype>, <additional sub-type specific values>
             }
 	    "metadata": {
-            "label_1": "value_1",
-	 	    "label_2": "value_2"
+            "label_1": "value_1", 
+            "label_2": "value_2"
             } (Optional)
 	}
 
@@ -984,7 +984,9 @@ Response body:
         "name": <headnode>,
         "project": <projectname>,
         "nics": [<nic1>, <nic2>, ...],
-        "vncport": <port number>
+        "vncport": <port number>,
+        "uuid": <headnode uuid>,
+        "base_img": <headnode base_img>
     }
 
 Authorization requirements:
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -415,10 +415,15 @@ see the documentation of the OBM driver in question (read `docs/obm-drivers.md`)
 `PUT /node/<node>`
 
 Request Body:
-	{"obm": { "type": <obm-subtype>,
-		<additional sub-type specific values>}
-	 "metadata": {"label_1": "value_1",
-	 	     "label_2": "value_2"} (Optional)
+
+	{
+        "obm": {
+             "type": <obm-subtype>, <additional sub-type specific values>
+            }
+	    "metadata": {
+            "label_1": "value_1",
+	 	    "label_2": "value_2"
+            } (Optional)
 	}
 
 example provided in USING.rst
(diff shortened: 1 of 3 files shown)


docs/rest_api.md
```diff
@@ -967,7 +984,9 @@ Response body:
         "name": <headnode>,
         "project": <projectname>,
         "nics": [<nic1>, <nic2>, ...],
-        "vncport": <port number>
+        "vncport": <port number>,
+        "uuid": <headnode uuid>,
+        "base_img": <headnode base_img>
```
zenhack: Should add descriptions for these to the bullet list above.

hil/api.py
```diff
@@ -1248,7 +1248,9 @@ def show_headnode(nodename):
     Example:  '{"name": "headnode1",
                 "project": "project1",
                 "hnics": ["hnic1", "hnic2"],
-                "vncport": 5900
+                "vncport": 5900,
+                "uuid": <headnode uuid>, 
+                "base_img": <headnode base_img>
```
zenhack: I think a while back we decided to remove a bunch of this info from the docstrings, in favor of keeping the detailed api descriptions in rest_api.md. Some stuff in this file still has examples because of legacy, but I think the thing to do here is just delete this, rather than try to keep it in sync with the same info elsewhere.

docs/rest_api.md
```diff
@@ -479,30 +484,42 @@ Authorization requirements:
...
 Accepts one optional boolean argument that determines whether to soft (default)
 or hard reboot the system.
 
-Request body:
-    {
-    	"force": <boolean> (Optional, defaults to False)
-    }
+Authorization requirements:
+
+* If the owner is a project, access to that project is required.
```
zenhack: I'm assuming this wording was copied from one of the networking related calls. "If the owner is a project" is a bit awkward in this case, since nodes can only be owned by projects (or be "free"), whereas with networks we have a notion of an admin-owned network. Maybe something more like:

* If the node is assigned to a project, access to that project is required.

Same goes for the other call you added this to.

docs/rest_api.md
```diff
@@ -415,10 +415,15 @@ see the documentation of the OBM driver in question (read `docs/obm-drivers.md`)
...
-	 	     "label_2": "value_2"} (Optional)
+
+	{
+        "obm": {
+             "type": <obm-subtype>, <additional sub-type specific values>
+            }
+	    "metadata": {
+            "label_1": "value_1", 
+            "label_2": "value_2"
+            } (Optional)
```
zenhack: I'm not sure I think this indentation is any better.

docs/rest_api.md
```diff
@@ -479,30 +484,42 @@ Authorization requirements:
...
-Request body:
-    {
-    	"force": <boolean> (Optional, defaults to False)
-    }
+Authorization requirements:
+
+* If the owner is a project, access to that project is required.
+* Otherwise, administrative access is required.
 
 #### node_set_bootdev
```
zenhack: We should probably add something here about the possible values for bootdev.

docs/rest_api.md
vsemp: Unfortunately, I don't have sufficient knowledge about "show_headnode" to do this.

docs/rest_api.md
vsemp: I followed the suggested format in the very beginning. There is an example in there.

docs/rest_api.md
vsemp: Changed this

hil/api.py
vsemp: Done

docs/rest_api.md
naved001: it's already there https://github.com/CCI-MOC/hil/blob/master/docs/rest_api.md#for-ipmi-devices

docs/rest_api.md
naved001: * "uuid", UUID for the headnode
* "base_img", the os image that the headnode is running 


(see this [link](https://github.com/CCI-MOC/hil/blob/master/examples/hil.cfg#L29))

vsemp: Thanks

docs/rest_api.md
zenhack: Ah, fair enough.

docs/rest_api.md
zenhack: It's probably easier for me to just submit a pr with the diff I have in mind than try to explain it in english; I'll do that once this is merged.

zenhack:
Looks good, merging.

docs/rest_api.md
zenhack: Edit: what's going on here is we're mixing tabs & spaces; if you render tabs as 4 spaces it looks right. I'm going to submit a PR getting rid of the tabs.

vsemp: Ok. Sorry if this became an issue. Actually, I just parsed / re-indented this JSON-like format thought Visual Studio.

zenhack: `python -m json.tool` formats things in the correct style. A formatting tool might not necessarily work here though, since this isn't quite valid json.

PR #887, opened 2017-09-27
add link to the individual contributor license agreement that contribut…

naved001:
…ors must sign before they can contribute to HIL


naved001:
@pns005 told me that we need contributors to sign that agreement for them to commit code to any moc repos.

Issue #873, opened 2017-09-13
Set up a linter to check for trailing whitespace in non-python files (e.g. docs).


zenhack:
Every now and then somebody submits a patch that removes some trailing whitespace, e.g.

https://github.com/CCI-MOC/hil/pull/866/commits/a5cec0ec4bf54c1e908e97d43633ddccc001094f#diff-96f699aa7db6da1ea1765db754a87a45L186

My editor is set up to do this automatically on save. This clutters up the diff a bit, but it never seems worth it to do a separate patch for that.

It would be nice if we had  a linter that would check these files so this doesn't come up.

I consider this low priority, but worth tossing up here on the bug tracker.

PR #889, opened 2017-09-29
Get logging to work and add examples in config file

naved001:
* Logging wasn't working because `import logging` wouldn't import the
submodule `handlers`; so we explicitly do it.

* Also, it wasn't apparent how to set the path for the log file in
the config file, so added that.

### DIFF
--- a/examples/hil.cfg
+++ b/examples/hil.cfg
@@ -14,7 +14,7 @@
 # logged.  Options include: debug, info, warn/warning, error, critical/fatal.
 log_level = debug
 
-# Set the directory for log file. Comment the line to disable logging to file.
+# Set the directory for the log file. Comment the line to disable logging to file.
 log_dir = /var/log/
 
 [auth]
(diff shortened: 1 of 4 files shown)


examples/hil.cfg
```diff
@@ -90,7 +93,7 @@ uri = sqlite:////absolute/path/to/hil.db
 #dry_run=
 
 [network-daemon]
-# The amount of time in seconds to sleep after attempting to empty the journal 
+# The amount of time in seconds to sleep after attempting to empty the journal
```
naved001: whitespace removed by my text editor.

zenhack: See also #873 (though obviously doesn't need to be dealt with here).

examples/hil.cfg
```diff
@@ -14,6 +14,9 @@
 # logged.  Options include: debug, info, warn/warning, error, critical/fatal.
 log_level = debug
 
+# Set the directory for log file. Comment the line to disable logging to file.
```
zenhack: For *the* log file.

naved001: done

PR #890, opened 2017-10-02
Fix #873

### LINKED ISSUE #873: Set up a linter to check for trailing whitespace in non-python files (e.g. docs).

zenhack:


naved001:
* it looks like the linter is replacing `:%VIRTUAL_ENV%/` in `ci/wsgi.conf` with a real value.

* also, we should only run this test [`if DB == sqlite`](https://github.com/CCI-MOC/hil/blob/master/ci/pep8_tests.sh#L4)

zenhack:
It was actually the existing apache CI stuff that was modifying wsgi.conf; I changed that so it shouldn't modify the one in the source tree anymore. Fixed the SQLite thing too.

PR #891, opened 2017-10-06
Maintenance Pool Service

Izhmash:
- If enabled and properly configured, `project_detach_node` will move the node to the maintenance project and POST to the maintenance service URL.
- `examples/hil.cfg` has been updated with the new config options.pool
- `docs/maintenance-pool` introduces the maintenance pool, shows an example use-case, and describes the configuration process.

### DIFF
--- a/docs/maintenance-pool.md
+++ b/docs/maintenance-pool.md
@@ -14,7 +14,7 @@ frees the node from the maintenance pool with `hil project_detach_node`.
 
 ## Configuration
 
-The maintenance pool will be active if the `maintenance` section exists in hil.cfg
-and `enable` is set to True.  If so, `maintenance_project` must be set equal to the
+The maintenance pool will be active if the `maintenance` section exists in hil.cfg.
+If so, `maintenance_project` must be set equal to the
 name of the maintenance project registered in HIL and `url` must point at the maintenance
 service.
--- a/docs/maintenance-pool.md
+++ b/docs/maintenance-pool.md
@@ -1,0 +1,20 @@
+# Maintenance Pool
+
+## Overview
+
+The maintenance pool is an optional service implemented by server administrators
+that ideally performs extra operations on nodes after they have been removed from a project
+by `hil project_detach_node`. If enabled, HIL will move the node from its original project
+into the designated maintenance project.  Then, HIL will POST to the URL of the service with
+the name of the node.
+
+Maintenance service example: Once the POST is received, the maintenance service
+logs when the node was detached, wipes the disks, sets the boot device to PXE, and then
+frees the node from the maintenance pool with `hil project_detach_node`.
+
+## Configuration
+
+The maintenance pool will be active if the `maintenance` section exists in hil.cfg
+and `enable` is set to True.  If so, `maintenance_project` must be set equal to the
+name of the maintenance project registered in HIL and `url` must point at the maintenance
+service.
--- a/examples/hil.cfg
+++ b/examples/hil.cfg
@@ -105,7 +105,7 @@ uri = sqlite:////absolute/path/to/hil.db
 #
 # If ``shutdown`` is present, the node will power off before moving to
 # the maintenance pool.
-# shutdown = 
+# shutdown =
 
 
 [network-daemon]
--- a/examples/hil.cfg
+++ b/examples/hil.cfg
@@ -95,14 +95,17 @@ uri = sqlite:////absolute/path/to/hil.db
 [maintenance]
 # Options for configuring the maintenance pool.
 #
-# Enables maintenance pool functionality if True. Default is False:
-# enable = False
+# Maintenance pool is enabled if the project and the url exist.
 #
 # Name of the maintenance project:
 # maintenance_project = maintenance
 #
 # URL of maintenance pool service:
 # url = <service url>
+#
+# If ``shutdown`` is present, the node will power off before moving to
+# the maintenance pool.
+# shutdown = 
 
 
 [network-daemon]
--- a/examples/hil.cfg
+++ b/examples/hil.cfg
@@ -92,6 +92,19 @@ uri = sqlite:////absolute/path/to/hil.db
 # enabled:
 #dry_run=
 
+[maintenance]
+# Options for configuring the maintenance pool.
+#
+# Enables maintenance pool functionality if True. Default is False:
+# enable = False
+#
+# Name of the maintenance project:
+# maintenance_project = maintenance
+#
+# URL of maintenance pool service:
+# url = <service url>
+
+
 [network-daemon]
 # The amount of time in seconds to sleep after attempting to empty the journal
 # when running serve_networks. If set, must be > 0 and < 3600 (1 hour). A
--- a/hil/api.py
+++ b/hil/api.py
@@ -1395,6 +1395,8 @@ def _maintain(project, node, node_label):
         raise errors.NotFoundError("Maintenance URL not in hil.cfg.")
     elif (cfg.has_option('maintenance', 'url')):
         raise errors.NotFoundError("Maintenance project not in hil.cfg.")
+    else:
+        return
 
     if (cfg.has_option('maintenance', 'shutdown')):
         node.obm.power_off()
--- a/hil/api.py
+++ b/hil/api.py
@@ -123,7 +123,6 @@ def project_detach_node(project, node):
     get_auth_backend().require_project_access(project)
     node_label = node
     node = _must_find(model.Node, node)
-    maintaining = False
     if node not in project.nodes:
         raise errors.NotFoundError("Node not in project")
     num_attachments = model.NetworkAttachment.query \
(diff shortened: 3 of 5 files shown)


hil/api.py
```diff
@@ -116,10 +117,15 @@ def project_detach_node(project, node):
 
     If the node has network attachments or pending network actions, a
     BlockedError will be raised.
+
+    If using a maintenance pool, please update the hil config accordingly.
     """
     project = _must_find(model.Project, project)
     get_auth_backend().require_project_access(project)
+    node_json = node
+    print(node_json)
```
naved001: debugging statement? Also, isn't `node` just a string? Can you post what it looks like when you print it?

hil/api.py
```diff
@@ -130,9 +136,27 @@ def project_detach_node(project, node):
...
+                raise errors.NotFoundError("Maintenance URL not in hil.cfg.")
+            else:
+                url = cfg.get('maintenance', 'url')
+            payload = json.dumps({'node': node_json})
     node.obm.stop_console()
     node.obm.delete_console()
     project.nodes.remove(node)
+    if (maintaining):
+        maintenance_proj.nodes.append(node)
+        requests.post(url, data=payload)
```
naved001: why not move this block in the else part of the block where you get the url (line 151)? that way you wouldn't need tthe `maintaining` variable.

naved001: Not sure if worth it, but should we get the response from the POST you do and then check if was succesful or not? (response >= 400?); and then either raise an error or write it to logs.

hil/api.py
Izhmash: I'm going to remove that print and replace the name with node_label.

hil/api.py
Izhmash: Yeah, checking the response is a good idea.  Also the reason I added the block there is because we need to catch all errors before removing the node from its project, and only then can we move it to the maintenance pool.

Izhmash:
Naved and I talked about some options for error checking the POST response.  We suggest that even if the POST fails, the node still be taken away from the user's project and put in the maintenance pool.  Then, an error will be logged for the admin to see.

We also thought it would be helpful to add a configurable for powering off the node when it gets moved into the maintenance pool.  This would be for users who don't want to setup a full-fledged service but still have basic maintenance done.

As for testing, would it be appropriate to have my tests in a new file? The API test is getting rather large.

examples/hil.cfg
```diff
@@ -89,6 +89,19 @@ uri = sqlite:////absolute/path/to/hil.db
 # enabled:
 #dry_run=
 
+[maintenance]
+# Options for configuring the maintenance pool.
+#
+# Enables maintenance pool functionality if True. Default is False:
+# enable = False
```
zenhack: Might be better to just have this be implied by the presence or absence of the other two options.

hil/api.py
```diff
@@ -116,10 +117,15 @@ def project_detach_node(project, node):
 
     If the node has network attachments or pending network actions, a
     BlockedError will be raised.
+
+    If using a maintenance pool, please update the hil config accordingly.
```
zenhack: This doesn't seem like a very good place for advice to sysadmins.

It might make sense to say something along the lines of "if a maintenance pool is configured, the maintenance service will be notified and the node will be moved to the maintenance project. otherwise it will be returned directly to the free pool."

...but we should probably put that in rest_api.md instead.

hil/api.py
zenhack: There's enough logic here that it might make sense to split it off into a helper function. That said, I'd do the change I suggested in the config file first, since it's going to affect the logic a bit.

Izhmash:
After some discussion with @naved001, we decided it would probably be more appropriate to add the sample service in a second pull request.
One issue came up when writing the tests: running `project_detach_node` with maintenance enabled requires a server to POST to.  As a placeholder I put a (real) dummy website that will take the POST.  Since this might not be appropriate for unit tests, we can either keep the POST out of the tests or try using `requests_mock`.
Are there any other options for the dummy POST that would be worth exploring?

hil/api.py
```diff
@@ -158,11 +158,12 @@ def project_detach_node(project, node):
...
-        # Give warning if response > 400
-        if (response > 400):
-            logger.warn('POST to maintenance service' \
-                    ' failed with response: %s' % response)
+        response = requests.post(url,
+                                 headers={'Content-Type': 'application/json'},
+                                 data=payload)
+        if (response.status_code > 400):
+            logger.warn('POST to maintenance service'
+                        ' failed with response: %s' % response)
```
naved001: shouldn't this be `response.text`?

hil/api.py
```diff
@@ -158,11 +158,12 @@ def project_detach_node(project, node):
...
         maintenance_proj.nodes.append(node)
-        response = requests.post(url, data=payload)
-        # Give warning if response > 400
-        if (response > 400):
-            logger.warn('POST to maintenance service' \
-                    ' failed with response: %s' % response)
+        response = requests.post(url,
+                                 headers={'Content-Type': 'application/json'},
+                                 data=payload)
+        if (response.status_code > 400):
```
naved001: should be **>=**

or maybe just `>= 300` because that also is a failure, isn't it?

Izhmash: Fixed.

zenhack: 3xx aren't failures, they're things like redirects (which we should follow) "not modified" responses (which aren't applicable here) etc. I think requests is configured to follow redirects by default, but be should be defensive. Probably the thing to do is to check `200 <= response_code < 300`, and assume the library already follows redirects.

tests/unit/api/maintenance-pool.py
```diff
@@ -0,0 +1,170 @@
...
+                    'configure',
+                    'fresh_database',
+                    'server_init',
+                    'with_request_context',
+                    'set_admin_auth']
+
+pytestmark = pytest.mark.usefixtures(*default_fixtures)
+
+
+class TestProjectDetachNodeMaintenance:
```
zenhack: These tests look identical to the existing ones. find a way to factor out the common logic, don't just copy/paste.

hil/api.py
```diff
@@ -1424,3 +1413,19 @@ def _must_find_n(obj_outer, cls_inner, name_inner):
                                     obj_outer.__class__.__name__,
                                     obj_outer.label))
     return obj_inner
+
+
+def _maintain(project, node, node_label):
+    """Perform maintenance tasks."""
```
zenhack: Could use a more detaild docstring here.

Izhmash: The checks have been moved.

tests/unit/api/maintenance-pool.py
Izhmash: I removed some unnecessary bits, but my idea was to keep the test relatively similar to the other test case.  Here, I wanted to test simply that the node goes into the maintenance pool when `project_detach_node` is run.

Izhmash:
I discovered that my maintenance pool test is causing the `cli.py` integration test to fail somehow, which I will be investigating.

docs/maintenance-pool.md
```diff
@@ -0,0 +1,20 @@
...
+the name of the node.
+
+Maintenance service example: Once the POST is received, the maintenance service
+logs when the node was detached, wipes the disks, sets the boot device to PXE, and then
+frees the node from the maintenance pool with `hil project_detach_node`.
+
+## Configuration
+
+The maintenance pool will be active if the `maintenance` section exists in hil.cfg
+and `enable` is set to True.  If so, `maintenance_project` must be set equal to the
```
naved001: `enable` doesn't exist anymore, yeah? We only check for the existence of `maintenance` section.

hil/api.py
```diff
@@ -119,6 +121,7 @@ def project_detach_node(project, node):
     """
     project = _must_find(model.Project, project)
     get_auth_backend().require_project_access(project)
+    node_label = node
```
naved001: just a suggestion, take it or leave it.
 don't really need to keep a copy here. we could just use `node.label` in the code later after we get Node object.

docs/maintenance-pool.md
Izhmash: Oops, missed that.  Fixed now.

hil/api.py
Izhmash: Yeah I definitely agree with that.  I'll remove node_label.

tests/unit/api/maintenance-pool.py
Izhmash: @zenhack do you still require further changes to the test?

tests/unit/api/maintenance-pool.py
```diff
@@ -0,0 +1,98 @@
...
+    """Test project_detach_node."""
+
+    def test_project_detach_node_maintenance(self, maintenance_proj_init):
+        """Test that project_detach_node removes the node from the project.
+        Note that the maintenance server has a fake url.  We expect it to
```
zenhack: This docstring could use some tweaking; it doesn't capture the salient point that we're checking that the node ends up in the maintenance project.

tests/unit/api/maintenance-pool.py
```diff
@@ -0,0 +1,98 @@
...
+        api.project_create('anvil-nextgen')
+        new_node('node-99')
+        api.project_connect_node('anvil-nextgen', 'node-99')
+        with pytest.raises(LoggedWarningError):
+            api.project_detach_node('anvil-nextgen', 'node-99')
+        project = api._must_find(model.Project, 'anvil-nextgen')
+        maintenance_proj = api._must_find(model.Project, 'maintenance')
+        node = api._must_find(model.Node, 'node-99')
+        assert node not in project.nodes
+        assert node.project is not project
```
zenhack: These two lines (a) do the same thing (sqlalchemy maintains the relationship between project.nodes and node.project), and (b) are implied by `node.project is mainenance_proj`, below, assuming `project is not maintenance_project`.

Also, rather than checking the projects with `is`, I would compare their labels with `==`. `is` checks pointer identity, so if somehow we have two objects in memory that represent the same project in the database, but aren't at physically the same spot in memory, this will fail. And I don't know off-hand what `==` will do on models. comparing the labels is more straightforward.
Also, use `==` not `is`. `is` checks pointer identity, not equality. We use it with `None` because it's in the pep8 guidelines (and always works there because there's only one `None`), but `is` is almost never what you want.

tests/unit/api/maintenance-pool.py
```diff
@@ -0,0 +1,98 @@
...
+            'hil.ext.network_allocators.null': None,
+            'hil.ext.network_allocators.vlan_pool': '',
+        },
+        'hil.ext.network_allocators.vlan_pool': {
+            'vlans': '40-80',
+        },
+        'maintenance': {
+            'maintenance_project': 'maintenance',
+            # Keystone url acts as dummy for posting
+            'url': 'http://127.0.0.1:35357/v3/'
```
zenhack: I would just pick something arbitrary; there' s no reason to confuse this with keystone.

zenhack:
Works for me, merging.

PR #894, opened 2017-10-12
Update _get_vlans to correctly gather all vlans.

naved001:
The switch can return a range of vlans (e.g. 14-18). That needs to parsed
correctly to get all the VLANs.


Izhmash:
Sometimes the switches do ranges of 2, e.g. 16-17.  Might want to add a test for that.

PR #895, opened 2017-10-13
Add get_capabilities to switches and use that to update deployment tests.

naved001:
* Add `get_capabilities(()` to switch drivers as described in #876 
* Update deployment tests for vlan_networks to use `get_capabilities()` and modify the tests accordingly.
* Tested it on Dell S3048-on (running Dell nos9) and our cisco nexus switch (with an assert False in the branches it shouldn't take). Passed both tests.

Question:
Is there an elegant way to re-use `ensure_legal_operations()` from `dellnos9.py` in `brocade.py`? And should I do that in this PR since I am dealing with deployment tests.

Note: should probably rebase this after #894 is merged.


tests/deployment/vlan_networks.py
```diff
@@ -78,8 +78,12 @@ def create_networks():
             # Create two networks
             network_create_simple('net-0', 'anvil-nextgen')
             network_create_simple('net-1', 'anvil-nextgen')
+            network_create_simple('net-2', 'anvil-nextgen')
+            network_create_simple('net-3', 'anvil-nextgen')
```
mosayyebzadeh: You forgot to delete net-2 and net-3 in network_delete.

naved001:
@mosayyebzadeh fixed. Let me know if there any other issues.

hil/model.py
```diff
@@ -258,6 +258,14 @@ def ensure_legal_operation(self, nic, op_type, channel):
 
         return
 
+    def get_capabilities(self):
+        """Returns the list of capabilities supported by a switch
+
+        default behaviour is to return a list with capabilities supported by
+        most switches"""
+
+        return ['nativeless-trunk-mode']
```
zenhack: I think I'd prefer to not have a default implementation, and mandate that drivers be explcit. Or, if we do want to have a default, it should return the empty list, so a driver author has to explicitly say their driver can do something.

tests/deployment/vlan_networks.py
```diff
@@ -173,6 +216,15 @@ def delete_networks():
...
                                             node.nics[0].label,
                                             attachment.network.label))
+
+            switch = nodes[0].nics[0].port.owner
+            # in some switches, the native network can only be disconnected
+            # after we remove all tagged networks first. The following checks
+            # for that and rearranges the networks (all_attachments) such that
+            # tagged networks are removed first.
+            if 'nativeless-trunk-mode' not in switch.get_capabilities():
+                all_attachments = sorted(all_attachments)
```
zenhack: Since you've run the tests I imagine this *does* work, but it would otherwise take a bit of staring at the code for me to be sure that sorting this actually results in the tagged ones first. Can you do this in a way that's more self-explanatory (even if it's a bit more verbose)?

tests/deployment/vlan_networks.py
```diff
@@ -133,10 +161,25 @@ def create_networks():
 
             # Verify that we can put nodes on more than one network, with
             # different channels:
+
+            # this is a little clumsy. `get_networks` only returns the node
+            # if it's attached natively. So I must reconnect node[2] to net-1
```
zenhack: Are we actually relying on this behavior anywhere? If so, we should document that more clearly, since it isn't obvious. If not, let's just fix it instead of hacking around it here.

tests/deployment/vlan_networks.py
```diff
@@ -215,17 +215,21 @@ def delete_networks():
                 for attachment in attachments:
                     all_attachments.append((node.label,
                                             node.nics[0].label,
-                                            attachment.network.label))
+                                            attachment.network.label,
+                                            attachment.channel))
```
zenhack: 4 elements in a tuple is enough that it starts to get hard to keep track of. Suggestion: use a namedtuple.

tests/deployment/vlan_networks.py
naved001: yeah, let's fix this within this PR. will do it in the next commit.

naved001: I was wrong about it. In the previous steps, we do connect tagged networks and those are correctly returned by get_networks. I'll get rid of this workaround.

naved001:
>  we should have a list of known switch capabilities somewhere, where we document what they mean and how HIL treats them.

in `network-drivers.md`?

naved001:
travis ci is so [dead right now](https://twitter.com/traviscistatus)

naved001:
@zenhack I think I made the changes you suggested. 
@mosayyebzadeh I have added some more commits after your approval; have a look at those once.

docs/network-drivers.md
```diff
@@ -81,9 +81,19 @@ HIL:
   Nexus switches. Only the 3500 and 5500 have been tested, though it is
   possible that other models will work as well.
 * ``hil.ext.switches.brocade``, for the brocade VDX 6740.
+* ``hil.ext.switches.n3000``, for Dell N3000 series switches.
+* ``hil.ext.switches.dellnos9``, for Dell switches running Dell Networking OS 9.
 
-None of the drivers require any extension-specific config options. Per the
-information in `rest_api.md`, the details of certain API calls are
+All switches may or may not have certain capabilities. The `show_switch` call
```
mosayyebzadeh: This is better sentence:
Different switches can have different capabilities.

tests/deployment/vlan_networks.py
```diff
@@ -166,14 +200,31 @@ def delete_networks():
...
+
+            if 'nativeless-trunk-mode' not in switch.get_capabilities():
+                # sort by channel; so vlan/<integer> comes before vlan/native
+                all_attachments = sorted(all_attachments,
+                                         key=lambda net: net.channel)
+
             for attachment in all_attachments:
+                attachment = (attachment.node, attachment.nic,
+                              attachment.network)
                 api.node_detach_network(*attachment)
```
zenhack: Could just collapse these two into:

```python
api.node_detach_network(attachment.node, attachment.nic, attachment.network)
```

tests/deployment/vlan_networks.py
```diff
@@ -166,14 +200,31 @@ def delete_networks():
...
+            switch = nodes[0].nics[0].port.owner
+            # in some switches, the native network can only be disconnected
+            # after we remove all tagged networks first. The following checks
+            # for that and rearranges the networks (all_attachments) such that
+            # tagged networks are removed first.
+
+            if 'nativeless-trunk-mode' not in switch.get_capabilities():
+                # sort by channel; so vlan/<integer> comes before vlan/native
+                all_attachments = sorted(all_attachments,
+                                         key=lambda net: net.channel)
```
zenhack: I'm still a bit bothered by the sorting thing; it's serendipitous that digits happen to come before letters in ascii/unicode, but that's a weird bit of knowledge to lean on for this, since it's so conceptually orthoginal. I'd be okay with it if we just improved the comment to detail exactly why this works.

naved001:
@zenhack @mosayyebzadeh addressed your  comments and rebased.

naved001:
@zenhack ping

Issue #761, opened 2017-04-24
Revert port when registering in HIL


naved001:
So I came across this when I was playing with the dell n3048. So when you enable trunking on it (`sw mode trunk`) by default the port has all the vlans enabled which is not cool. This may be the case in other switches too.  

What we should do is, during port registration, call `revert_port` on the port being registered. That way, when a new port is registered we can be sure that it has no vlans enabled. (Or are we doing this already?)

cc: @zenhack

zenhack:
Yeah, I thought there was a note somwehre about admins needing to shut that off when setting up. But now that we have the functionality we probably ought to just do it automatically.

naved001:
@zenhack just realised that I can't call port revert on a port that is not attached to any nic. Why is that so?

naved001:
@zenhack ping

zenhack:
It's an artifact of the fact that NewtorkingActions are bound to nics, not ports, so to actually shove something in the queue we need a nic. This probably shouldn't be the case, just a limitation of the existing implementation.

It would require a bit of refactoring to fix. I *think* we could just switch what object the networking action was bound to, though there are likely a few subtleties in the places where we check for the existance of an action.

naved001:
hm, maybe we can make it tunable via the configuration file. 

`[installation]`
`complete-wipe` or `sneaky`

that way we can do other stuff too, like turn off nodes when they are registered.

zenhack:
maybe a more stragithforward name for the option, like `clean_resources_on_register = True`, but that seems reasonable.

naved001:
so we can make it tunable via the config file, and call `revert_port()` when we do a `port_connect_nic()`. And since that would be an admin only operation that'll be performed before the nodes are given to users, I think it'll be safe to do. 

What do you think @zenhack?

naved001:
Now that I think of it, this may not always work and is probably a bad idea. Sometimes a switchport may have more configuration than just VLANs. Our revert_port call only removes networks, not any other configuration on a switchport (like ip-address, qos etc etc). So it's best to leave it up to admins to set the state right before registering in HIL; otherwise there may be unwanted side affects.

PR #896, opened 2017-10-13
Call revert port with port_connect_nic

naved001:
* Issue #761 
* This puts the switchport in a state HIL expects* it to be. So admins don't have
to worry about clearing the switchport before registration

* Update the deployment tests and the sample config files to work with it.

If this seems reasonable I'll add some unit tests for this.

Again, should probably rebase this one once/if my other 2 PRs are merged (just to be safe).

### DIFF
--- a/tests/unit/migrations/pending-networking-actions.sql
+++ b/tests/unit/migrations/pending-networking-actions.sql
@@ -599,7 +599,7 @@ INSERT INTO network_projects (project_id, network_id) VALUES (1, 1);
 -- Data for Name: networking_action; Type: TABLE DATA; Schema: public; Owner: hil
 --
 
-INSERT INTO networking_action (id, nic_id, new_network_id, channel) VALUES (1, 1, 1, 'null');
+INSERT INTO networking_action (id, nic_id, new_network_id, channel) VALUES (2, 1, 1, 'null');
 
 
 --
--- a/examples/hil.cfg
+++ b/examples/hil.cfg
@@ -17,6 +17,12 @@ log_level = debug
 # Set the directory for the log file. Comment the line to disable logging to file.
 log_dir = /var/log/
 
+# Set this to true to clean resources during HIL registration. If this option
+# is enabled, then when a switchport is registered, HIL will remove all vlans
+# from that switchport (by calling the revert port API).
+# Set the value to False or comment out the line to disable this feature.
+clean_resources_on_register = True
+
 [auth]
 # There are a handful of API calls that require no special access to execute.
 # By default, in this case the user must still successfully authenticate as
--- a/examples/hil.cfg
+++ b/examples/hil.cfg
@@ -17,12 +17,6 @@ log_level = debug
 # Set the directory for the log file. Comment the line to disable logging to file.
 log_dir = /var/log/
 
-# Set this to true to clean resources during HIL registration. If this option
-# is enabled, then when a switchport is registered, HIL will remove all vlans
-# from that switchport (by calling the revert port API).
-# Set the value to False or comment out the line to disable this feature.
-clean_resources_on_register = True
-
 [auth]
 # There are a handful of API calls that require no special access to execute.
 # By default, in this case the user must still successfully authenticate as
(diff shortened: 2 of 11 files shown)


SahilTikale:
Why do we need this flag ? 
I think this should be the default behaviour for all new ports that we register with HIL.
Is there a user-case where this flag is set to `False` ?

naved001:
This [comment](https://github.com/CCI-MOC/hil/issues/761#issuecomment-327867789) by @zenhack 

In short, if we wanted to deploy HIL in an environment without disturbing the pre-existing configuration; this patch would always remove the old config and start afresh. 

But now that I think of it, if there's some pre-existing configuration then we would want HIL to have it in it's database; so are we expecting HIL admins to manually edit the database? because I can't think of any other way of syncing a switchport status with HIL database. cc @zenhack

zenhack:
Okay, yeah, I want to think more carefully about the details of retrofitting HIL onto an existing setup before we put some option to try to support it. Will mull it over.

naved001:
>  because I can't think of any other way of syncing a switchport status with HIL database

`get_port_networks()` gets half the info. We would still need to create networks in HIL (network label with a net id). And what if the networks on the switch fall outside the range of HIL allocation pool? do they become admin networks then?

naved001:
@SahilTikale @zenhack 
any thoughts?

zenhack:
I'm tempted to suggest that we ignore that use case entirely for now, and deal with it separately. We're not really breaking anything, since you'd have to do some careful mucking with things (bordering on black magic) to get this to work as-is.

naved001:
@SahilTikale @zenhack got rid of the config file changes, and rebased this PR.

naved001:
I just gotta fix the plethora of failed tests, sigh.

naved001:
Fixing the unit tests require some work. So everywhere we do `port_connect_nic()` we need to follow it with `deferred.apply_networking()`. Is there a smart of doing this, or do I go about changing all files?

naved001:
Some of the tests fail locally for postgres but they don't stall like travis ci. :/c

tests/unit/migrations/pending-networking-actions.sql
```diff
@@ -599,7 +599,7 @@ INSERT INTO network_projects (project_id, network_id) VALUES (1, 1);
 -- Data for Name: networking_action; Type: TABLE DATA; Schema: public; Owner: hil
 --
 
-INSERT INTO networking_action (id, nic_id, new_network_id, channel) VALUES (1, 1, 1, 'null');
+INSERT INTO networking_action (id, nic_id, new_network_id, channel) VALUES (2, 1, 1, 'null');
```
naved001: don't know how I feel about this change, but I can't think of any other solution right now. @zenhack any ideas on this?

naved001:
@SahilTikale @zenhack ping

tests/unit/migrations/pending-networking-actions.sql
zenhack: I am *very* uncomfortable with this.

I also kindof don't like the fact that we're caling one of the request handlers from another. It feels wrong.

Adding the config option would solve needing to modify the database dump, but I don't think that justifies it by itself.

I'm not totally sure what I think the right thing to do here is. Maybe we're losing the plot a little bit here -- The HIL server is supposed to be bare-bones. It's not intended to be a high-level tool. One approach is to just build separate tools for admins to use that do this correctly; we don't actually need server-side support I don't think.

Still waffling a bit. Thoughts?

naved001: We should do this in the CLI then. Perhaps, when we have a new CLI, we could do `port_connect_nic --clean` which would make 2 calls, one for `port_connect` and another for `revert_port`. Or maybe get rid of the flag and make it the default behavior. That way we don't touch the APIs

zenhack: OK. do you want to just close this then?

naved001:
closing this, we'll implement something like this in the CLI then.

Issue #899, opened 2017-10-16
Saving to flash memory for all switches


naved001:
Old issue #688 

We did not add a method for the brocade switch because we thought it automatically saves to the flash memory. The reason for that behavior is that our brocade switches are stacked ([logical chassis mode](http://www.brocade.com/content/html/en/administration-guide/nos-601a-adminguide/GUID-BF9E6BA9-A949-48CB-AE57-ADBBC376DE0C.html)) and in that case all switches sync the running config to startup config automatically. 

I think we should always provide a method in our drivers for every switch to save to flash memory because a switch may not always be running in stacked configuration. 

I'll make a PR that does this for the brocade and dell s3048-on.

@SahilTikale how does stacking work in the Juniper world?
cc: @zenhack

naved001:
I haven't really figured out how to do this on the brocade switch using the rest api.

http://www.brocade.com/content/dam/common/documents/content-types/api-reference-guide/nos-700-restapiguide.pdf

```python
    def save_running_config(self):
        url = self.hostname + '/rest/' + 'operational-state/bna-config-cmd'
        body = '<bna-config-cmd><src>running-config</src><dest>startup-config \
                </dest></bna-config-cmd>'
        response = self._make_request('POST', url, data=body)
```

this doesn't seem to save though and it's not very clear from the docs what bna-cfg is? but this was only command doing something with the running and startup config. 

Other than that, I could only read the `running-config` using the rest api, and not the `startup-config`. (So we can't implement `get_config()` for this switch). 

@Izhmash do you think you could give a second look at the documentation, maybe I overlooked something? Even if there's a REST API CLI mode that would be helpfui too (that way we can package any cli command in the REST calls, just like the dellnos9 driver)

If there's no way to do this using the rest api then we have the following options:

1. Document that we don't support saving on this switch.
2. use pexpect to do it.

naved001:
@zenhack any thoughts about how to move forward with this?

zenhack:
What does the response body from that request look like? From the docs it looks like the call is asynchronus; you may have to poll the id it returns and wait for it to finish.

naved001:
yeah, the body returned a status_id to poll on. [Here's the](https://github.com/CCI-MOC/hil/compare/master...naved001:save-on-brocade?expand=1) code for it. Even after I wait for it to finish, it doesn't look like it saves it. And it's not clear to me from the docs what it is saving.

naved001:
[...]

PR #901, opened 2017-10-18
Refactor some console-based switch driver code

naved001:
I wanted to work on #899 but when I started working on it, I figured our old code could use some refactoring which will help in cleanly implementing #899 (which I'll do in a separate PR now). All I wanted to do was re-use `_should_save()`.

The commit messages have enough information I think.

### DIFF
--- a/hil/ext/switches/dell.py
+++ b/hil/ext/switches/dell.py
@@ -86,11 +86,6 @@ class _PowerConnect55xxSession(_BaseSession):
         self.switch = switch
         self.console = console
 
-    def _sendline(self, line):
-        logger.debug('Sending to switch %r: %r',
-                     self.switch, line)
-        self.console.sendline(line)
-
     @staticmethod
     def connect(switch):
         """connect to the switch, and log in."""
--- a/hil/ext/switches/dell.py
+++ b/hil/ext/switches/dell.py
@@ -29,6 +29,7 @@ from hil.ext.switches._dell_base import _BaseSession
 from os.path import dirname, join
 from hil.errors import BadArgumentError
 from hil.model import BigIntegerType
+from hil.ext.switches.switch_common import _should_save
 
 paths[__name__] = join(dirname(__file__), 'migrations', 'dell')
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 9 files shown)


naved001:
Note: I ran deployment tests (vlan_networks, native_networks, switch_config) on the dell n3000 and the cisco nexus switch, and all tests passed.

naved001:
@mosayyebzadeh @razaaliraza this PR is mostly just refactoring; so should be easier to review.

naved001:
@izhmash I think it will be useful going forward. For example, for enabling jumbo frames we would be checking the config file to do that, so a helper method to do it could live in that file. Basically, anything to do with just DB checks, config file checks could be put in that file.

naved001:
@zenhack would appreciate your review of this PR.

hil/ext/switches/dell.py
```diff
@@ -29,6 +29,7 @@
 from os.path import dirname, join
 from hil.errors import BadArgumentError
 from hil.model import BigIntegerType
+from hil.ext.switches.switch_common import _should_save
```
zenhack: Two things:

1. The leading underscore convention means "this is an implementation detail," so you shouldn't name it that if it's meant to be exported from another module.
2. Instead, of `switches.switch_common`, maybe just call it `switches.common`.

hil/ext/switches/n3000.py
```diff
@@ -195,3 +196,8 @@ def get_port_networks(self, ports):
                 networks.append(('vlan/native', native))
             result[k] = networks
         return result
+
+    def disconnect(self):
+        if _should_save('n3000'):
+            self._save_running_config()
+        self._disconnect()
```
zenhack: Maybe factor this bit out into `_dell_base`? the only bit that's different is the parameter to `_should_save`.

Idea: change the implementation of `_should_save` so that instead of a name, it takes a switch object. Then, it gets the name of the module it needs by doing `obj.__class__.__module__`. Then this `disconnect` can be in some superclass and just pass in `self.`

naved001:
@zenhack made the changes. I ran all switch related deployment tests again, just to be sure.

zenhack:
This looks good to me, but the CI is failing because of the keystone thing. Rebasing on top of master should fix that. Once travis passes, I'll approve.

naved001:
I just realized that I didn't update the nexus driver in my last commit. Let me update that.

naved001:
@zenhack rebased this and updated the nexus.
@Izhmash want to have another look? I made some changes after your approval.

Izhmash:
Still looks good to me!

Issue #536, opened 2016-02-25
docs/rest_api.md is out of sync with implementation


zenhack:
People have been changing the API without changing the docs. We need to get these back in sync, and we need to make sure this doesn't happen again.

The file `docs/rest_api.md` contains a detailed description of the API. that description _MUST_ be updated with _EVERY_ change to the API.

Submitters: If you change the API, change the docs.

Reviewers: Do not **under any circumstances** merge a patch which changes the API without updating the docs.

@SahilTikale, in an attempt to keep this _really important_ issue from just getting forgotten about, I'm assigning this to you. Make sure the recent changes to the API are documented. At least some of the changes in question are yours, but I haven't done a full audit. Submit a pr with the documentation updates. I'd be interested in being asked to review that patch.

A few of the API changes themselves seem somewhat dubious, but I don't really have the time to pick through them all right now.

//cc @henn, @gsilvis, @lokI8, @SahilTikale, @knikolla, @okrieg, @pjd-nu, @pns005

zenhack:
Also, for each of these documentation sync issues we find, we should make sure there's a test! It shouldn't be possible to change the output of an API call without causing test failures.

SahilTikale:
I was not aware of this.  Will take care of it in my pull requests and will monitor for all other pull requests that involve changing API calls.

zenhack:
@SahilTikale, thanks. Sorry to be so stern about it.

I suspect you're not the only one who's made this mistake; it would be good to go through and audit the changes that have been made, since there's probably other drift. Looking at the history for `api.py` might be a good place to start, maybe going back through september or so, which is when I left.

zenhack:
Apparently I'm guilty as well (#537). Oops.

Glad people are tackling this. I'd like to help with review, but since I'm billing you guys by the hour for that I feel like folks have to explicitly ask me to first. Ping me if you need a pair of eyes on these things.

SahilTikale:
knock knock @vsemp

SahilTikale:
any updates on this @vsemp

vsemp:
I didn't see any inconsistent arguments or completely incorrect descriptions.

There the following potential inconsistencies of rest_api.md with api.py:
- Console code doesn't have documentation
- list_active_extensions from "API Extensions" is in api.py while other "API Extensions" are not

vsemp:
I suggest writing a separate function that checks some common access requirements.

**Here is one example:**

> Authorization requirements:
[...]

PR #902, opened 2017-10-22
Clean up redundant auth check pattern.

### LINKED ISSUE #623: Checking of project parameter in auth calls
Currently, one could pass `None` or a string to the auth methods `have_project_access` and `require_project_access`.

Passing `None` is especially easy, since `network.creator` can be `None` or a `Project`.

I propose that we should check for `None` in these methods, and if we see it, make the output depend on just `self._have_admin()`. We should also update the documentation to reflect this.

Additionally, if it's straightforward, it would be nice to verify that `project` is an object of type `Project`, throwing an exception if it isn't. That would hopefully catch errors where folks passed in a string.

zenhack:
This generalizes the interface for {require,have}_project_access, so
that passing `None` is valid, and equivalent to {require,have}_admin.
This lets us simplify a whole bunch of stuff like this:

    if project is None:
        require_admin()
    else:
        require_project_access(project)

This also fixes #623, as there is already a check requiring the argument
to be an instance of Project.

I seem to recall someone pointing out the redundency recently, but I can't remember who.

### DIFF
--- a/hil/api.py
+++ b/hil/api.py
@@ -148,13 +148,9 @@ def network_grant_project_access(project, network):
     """
     network = _must_find(model.Network, network)
     project = _must_find(model.Project, project)
-    auth_backend = get_auth_backend()
 
     # Must be admin or the owner of the network to add projects
-    if network.owner is None:
-        auth_backend.require_admin()
-    else:
-        auth_backend.require_project_access(network.owner)
+    get_auth_backend().require_project_access(network.owner)
 
     if project in network.access:
         raise errors.DuplicateError(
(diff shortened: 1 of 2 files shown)


naved001:
> I seem to recall someone pointing out the redundency recently, but I can't remember who.

This [comment](https://github.com/CCI-MOC/hil/issues/536#issuecomment-332298998) by @vsemp
With this and #886 we can close #536

zenhack:
Quoting Naved Ansari (2017-10-22 03:21:15)

> This comment by @vsemp

Aha.

naved001:
@SahilTikale, @izhmash

It's a small pr, any of you want to review this?

naved001:
@Izhmash thanks for the late night review haha!

Izhmash:
No prob! although the build is erroring out, hmm.

naved001:
hmm, looks like it can't find a keystone commit that we install. unrelated to this PR; will look into it tomorrow morning. In the meanwhile I restart the build (just to be sure)

jeremyfreudberg:
@naved001 stable/newton branch is/was EOL and very recently pruned from keystone. (the name stable/newton is no longer a valid git reference, although obviously you can still checkout any specific hash you want)

PR #905, opened 2017-10-27
Use setuptools_scm to derive the version.

zenhack:
This way we don't have to remember to update the version when we make a
release; the version will be derived from the git tag (and stored in the
package info if we make an sdist). We actually forgot this when we
tagged v0.2!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,6 @@ def _get_readme():
         return f.read()
 
 setup(name='hil',
-      version='0.2rc2',
       maintainer='Developers of the HIL Project at MOC',
       maintainer_email='hil@lists.massopen.cloud',
       url='https://github.com/CCI-MOC/hil',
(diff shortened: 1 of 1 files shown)


zenhack:
Weird, I get `0.3.dev45+g1709920`.

naved001:
The number followed by 'g' is the commit number which is the same in both cases. Not sure what the rest of the numbers mean.

naved001:
The number followed by 'g' is the commit number which is the same in both cases. Not sure what the rest of the numbers mean.

naved001:
> If you do git tag, what do you get?

channel-proposal
v0.1
v0.1.1

> Based on that, my result is what is expected.

It all makes sense now.

zenhack:
Yeah, that would do it.

naved001:
@SahilTikale @mosayyebzadeh @izhmash can we get second review on this?

naved001:
2 sign offs, so merging.

PR #907, opened 2017-10-27
ssh connection for switches

mosayyebzadeh:


hil/client/node.py
```diff
@@ -100,3 +100,16 @@ def stop_console(self, node):
...
+        """Set metadata. This is used for saving PCR values based on node's UUID label"""
+        url = self.object_url('node', node_name, 'label', label, 'metadata_set')
+        payload = json.dumps({'value': value})
+        return self.check_response(self.httpClient.request('PUT', url, data=payload))
+
+    def metdata_delete(self, node_name, label):
+        """Delete metadata. Using UUID label, PCRs' value is deleted """
+        url = self.object_url('node', node_name, 'label', label, 'metadata_delete')
+        return self.check_response(self.httpClient.request('DELETE', url))
+
```
naved001: metadata stuff should be its own PR.

hil/ext/switches/dell.py
```diff
@@ -94,7 +94,8 @@ def _sendline(self, line):
     @staticmethod
     def connect(switch):
         """connect to the switch, and log in."""
-        console = pexpect.spawn('telnet ' + switch.hostname)
+        console = pexpect.spawn('ssh ' + switch.username + '@' + switch.hostname)
+        #console = pexpect.spawn('telnet ' + switch.hostname)
```
naved001: you should remove the commented out code.

hil/ext/switches/nexus.py
```diff
@@ -126,9 +126,9 @@ def disconnect(self):
     @staticmethod
     def connect(switch):
         """Connect to the switch."""
-        console = pexpect.spawn('telnet ' + switch.hostname)
-        console.expect('login: ')
-        console.sendline(switch.username)
+        console = pexpect.spawn('ssh ' + switch.username + '@' + switch.hostname)
+        #console.expect('login: ')
+        #console.sendline(switch.username)
```
naved001: same thing, commented code.

naved001:
* Closes #46

mosayyebzadeh:
@naved001 Done.

naved001:
@mosayyebzadeh from travis
```
hil/ext/switches/dell.py:97:80: E501 line too long (81 > 79 characters)

hil/ext/switches/nexus.py:129:80: E501 line too long (81 > 79 characters)
```

You have pep8 failures.

docs/network-drivers.md
```diff
@@ -148,6 +150,12 @@ Register ports just like the powerconnect driver. e.g. ``gi1/0/5``.
 
 ### Nexus driver
 
+#### Switch preperatio
```
zenhack: spelling.

docs/network-drivers.md
```diff
@@ -148,6 +150,12 @@ Register ports just like the powerconnect driver. e.g. ``gi1/0/5``.
...
+
+A few commands are necessary to run on the switch before it can be used with HIL.
+
+1. This switch uses ssh for connection. Be sure that ssh is enabled on the switch.
```
zenhack: Given that we just have the one thing here, I might cut out the "A few commands..." bit from this section and the leading `1.`; it can just be one sentance.

naved001:
> @naved001, do you plan to do the n3000 as a separate pr?

Yeah, I'll make a separate pr for it to keep the size of this one down.

And as far as pub key auth goes, console.pexpect wouldn't expect "password" prompt  anymore, it will just login. Will test this with a real switch when I'm in front of a computer.

hil/ext/switches/dell.py
```diff
@@ -94,7 +94,7 @@ def _sendline(self, line):
     @staticmethod
     def connect(switch):
         """connect to the switch, and log in."""
-        console = pexpect.spawn('telnet ' + switch.hostname)
+        console = pexpect.spawn('ssh ' + switch.username + '@' + switch.hostname)
```
naved001: I would add a comment here saying that the dell switch still expects a username even after we specified the username in the spawn command.

naved001:
> > And as far as pub key auth goes, console.pexpect wouldn't expect "password" prompt anymore, it will just login. Will test this with a real switch when I'm in front of a computer.

Yep that's how it was on the nexus switch.  I'll add support for pubkey auth and the dell n3000 in a separate PR.

mosayyebzadeh:
Thanks for the the reviews. All done!

naved001:
looks good to me, just rebase this and I'll approve.

naved001:
You still have pep8 errors. From travis
```
hil/ext/switches/dell.py:94:9: E265 block comment should start with '# '

hil/ext/switches/nexus.py:125:21: E222 multiple spaces after operator
```
It would be a good idea to run pep8 tests locally before you push changes and wait for travis to run it.

naved001:
@mosayyebzadeh just one question. 

http://pexpect.readthedocs.io/en/stable/api/pxssh.html

This uses some different methods for ssh connection. Do you know what are the advantages/disadvantages of using that?

mosayyebzadeh:
I believe it is better to use that. It handles some tricky features related to ssh like first time connection to switch which needs to say "yes" first then send the user name.

I will change the code to use this one.

naved001:
@mosayyebzadeh want to update us on this PR?

mosayyebzadeh:
I checked the code. It answers some rare cases (e.g., the first time connection) however adds to complexity of code. I think our simple way which handles our needs is enough.

zenhack:
Sounds good to me, but in that case we should mention those cases in the docs.

naved001:
Yeah, we should document that we aren't handling the case where a user logs in for the first time; so admins should "test login" to their switch at least once. @mosayyebzadeh

mosayyebzadeh:
Yes, you are right. I will update the document ASAP.

naved001:
You should update the documents [here](https://github.com/CCI-MOC/hil/blob/master/docs/deployment.md)

mosayyebzadeh:
I updated the deployment document. For the public key support, I think we need another pr and therefore I did not mention it in the doc.

zenhack:
Re: first time connection, that's separate from public key auth -- by public key auth, we mean authentication the client with a public key. Regardless of how you authenticate the user, there's still a host key that you're prompted about the first time you connect to a box, e.g.:

```
The authenticity of host '111.222.333.444 (111.222.333.444)' can't be established.
RSA key fingerprint is f3:cf:58:ae:71:0b:c8:04:6f:34:a3:b2:e4:1e:0c:8b.
Are you sure you want to continue connecting (yes/no)? 
```

We should add something to the docs that reminds adminstrators that when deploying they need to do that manually once, as the user the network daemon runs as.

mosayyebzadeh:
I have done that. In deployment document, I have mentioned that the admins should take care of the first time connection to the switches.

zenhack:
I don't see it in the diff; did you forget to add/commit/etc?

mosayyebzadeh:
It is awkward, I pushed it and Naved checked the updates. Anyhow, I updated deployment doc. again.

naved001:
@mosayyebzadeh tests are failing because you have trailing whitespaces in the docs; can you fix those?

mosayyebzadeh:
@zenhack I have updated the files. Is there any other requirements?

zenhack:
@mosayyebzadeh, I just sent you a pr with some changes. Other than that this looks good.

zenhack:
Great, merging.

PR #911, opened 2017-11-03
Save dellnos9 switch

naved001:
Enable saving the switch configuration file to flash memory for the dell s3048-on switch/.
And there are some minor changes in the driver itself, but it's all in different commits. 

Will do the same for the brocade switch in a different PR.

Please scrutinize my PRs real hard, I have tried to be really careful but still miss out on some details. 

Also, will probably want to rebase after other switch related PRs are merged.

### DIFF
--- a/examples/hil.cfg
+++ b/examples/hil.cfg
@@ -134,3 +134,6 @@ save = True
 # Same behaviour as the dell switch. Set `save` to False to stop the switch
 # from writing to flash memory.
 save = True
+
+[hil.ext.switches.dellnos9]
+save = True
--- a/hil/ext/switches/dellnos9.py
+++ b/hil/ext/switches/dellnos9.py
@@ -214,7 +214,7 @@ class DellNOS9(Switch, SwitchSession):
         """
         command = 'interfaces switchport %s %s' % \
             (self.interface_type, interface)
-        response = self._execute(interface, SHOW, command)
+        response = self._execute(SHOW, command)
         return response.text.replace(' ', '')
 
     def _add_vlan_to_trunk(self, interface, vlan):
@@ -230,7 +230,7 @@ class DellNOS9(Switch, SwitchSession):
             self._port_on(interface)
         command = 'interface vlan ' + vlan + '\r\n tagged ' + \
             self.interface_type + ' ' + interface
-        self._execute(interface, CONFIG, command)
+        self._execute(CONFIG, command)
 
     def _remove_vlan_from_trunk(self, interface, vlan):
         """ Remove a vlan from a trunk port.
@@ -240,7 +240,7 @@ class DellNOS9(Switch, SwitchSession):
             vlan: vlan to remove
         """
         command = self._remove_vlan_command(interface, vlan)
-        self._execute(interface, CONFIG, command)
+        self._execute(CONFIG, command)
 
     def _remove_all_vlans_from_trunk(self, interface):
         """ Remove all vlan from a trunk port.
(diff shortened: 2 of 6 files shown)


Izhmash:
We can abstract `save_running_config` as it is a function used by all of the switches.

naved001:
@Izhmash yep, did that. though, it's not in the brocade yet; but should be soon.

hil/ext/switches/dellnos9.py
```diff
@@ -338,9 +344,49 @@ def _is_port_on(self, port):
         assert shutdown in ('false', 'true'), "unexpected state of switchport"
         return shutdown == 'false'
 
+    def save_running_config(self):
+        command = 'write'
+        self._execute(EXEC, command)
+
+    def _get_config(self, config_type):
```
zenhack: So, this method is attached to a couple of other switch drivers, but isn't actually described in any of the interfaces, and appears to only be used by the test suite. and the leading underscore convention suggests that this is some internal-use method, so I was looking for uses elsewhere in the driver. This is kindof horrible; we shouldn't be breaking interface boundaries like this. Can we find a way to make it a bit more clear what this is here for, and that this is a public method for use by the test suite? I'm not super pleased by the way this is structured...

examples/hil.cfg
```diff
@@ -134,3 +134,6 @@ save = True
 # Same behaviour as the dell switch. Set `save` to False to stop the switch
 # from writing to flash memory.
 save = True
+
+[hil.ext.switches.dellnos9]
+save = True
```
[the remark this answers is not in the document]
naved001: >  Is there a reason we would want to have this on for some switches but not others?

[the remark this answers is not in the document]
naved001: >  Is there a reason we would want to have this on for some switches but not others?

If the switches are stacked (aka logical chassis mode), then switches are saving automatically and the calls to `save_running_config` would fail. 
E.g.
```
sw0# copy running-config startup-config 
This operation is not supported in logical chassis mode.
```

Of course, we could handle a scenario like this in code, but I think the burden should be on hil-admins (who will know what switches they have stacked or not) to do this setup correctly because HIL would unnecessarily try and fail in this case.

hil/ext/switches/dellnos9.py
naved001: yeah, breaking interface boundaries is bad. 

I just did what was approved in the past for the other drivers. ( #753, #749)

In any case, I'll address this and the other comment you made.

examples/hil.cfg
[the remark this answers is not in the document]
naved001: Point being, some type of switches could be stacked while other type might not be, and I think admins should use the hil config file to tell HIL about it.

naved001:
@zenhack have a look at the latest commit and let me know how it looks now.

tests/deployment/switch_config.py
```diff
@@ -67,29 +68,28 @@ def configure():
...
 class TestSwitchSavingToFlash(NetworkTest):
-    """ saves the running config to the flash memory. Test is only for the dell
-        switch"""
+    """ saves the running config to the flash memory and tests if it succeeded
+    by comparing the running and startup config files"""
 
     def get_config(self, config_type):
-        """returns the switch configuration file"""
+        """helper method to get the switch config file by calling
+        the switch's get_config method
```
zenhack: Nitpick: get_config is defined on session, not switch.

zenhack:
I'm basically happy with this. The get_config interface is a bit gross - I don't like how much it exposes about how the switches work - but at least it's documented/not breaking interface boundaries anymore.

I'm going to open a separate issue to clean up that interface a bit. Fix the one in-line comment I made and I'll approve.

tests/deployment/switch_config.py
naved001: updated

zenhack:
LGTM

naved001:
@izhmash would you review this?

Izhmash:
This looks all set to me!

PR #916, opened 2017-11-28
Use ssh authentication for connecting to dell n3000 switches

naved001:


naved001:
Few things.

I wanted to get public key authentication to work. But I have a few thoughts.

1.  

```python
alternatives = ['password: ', '>', '#']
        console = pexpect.spawn(
            'ssh ' + switch.username + '@' + switch.hostname)

        outcome = console.expect(alternatives)
        if outcome == 0:
            console.sendline(switch.password)
            outcome = console.expect(alternatives)
        if outcome == 1:
            console.sendline('en')

        logger.debug('Logged in to switch %r', switch)
        prompts = _console.get_prompts(console)
```
if we have pub key authentication enabled for this switch. It wouldn't expect a password and will directly match '>' and then we send 'en' and we are all happy. But in the case of the nexus switch, or the powerconnect switches that we have, there's no need to get into privilege mode; so the switch matches with '#'. And after that we call `_get_prompts` which does `console.pexpect('some regex match for #')` which fails because the previous call to `expect` modifies the `before` and `after` property of `console`. I am thinking of some ways to workaround this issue, I might have to poke the _get_prompts method.

2. `connect` methods for console switches are hardcoded to follow only one path. We should either document this or handle all the possible cases of a switch login.
    1. ssh to switch
    2.  get into privilege mode (by entering 'en' short for enable).
    3. enter the enable mode password (which may not be the same as the regular password)
    4. you arrive at the main prompt.

A switch can be configured to enable/disable step ii and iii.

zenhack:
1. Let's document the current situation re: hardcoded paths.
2. Probably better to use the long-form `enable`, rather than just `en`, as it's a bit more clear. We should do this for driver code in general, though I suspect we aren't in a number of places.
3. My inclination is to defer handling various configurations until a separate pr at least. Though at a glance your existing approach feels reasonable.

naved001:
```
(.venv) naved:~/n3000-ssh/hil$ py.test tests/deployment/*_networks.py
========================================= test session starts =========================================
platform linux2 -- Python 2.7.12, pytest-3.2.5, py-1.5.2, pluggy-0.4.0
rootdir: /home/naved/n3000-ssh/hil, inifile: setup.cfg
plugins: cov-2.5.1, catchlog-1.2.2, forked-0.2, xdist-1.20.1
collected 2 items                                                                                      

===================================== 2 passed in 279.64 seconds ======================================
```
@mosayyebzadeh could you review this?

PR #917, opened 2017-12-04
Leasing script

mosayyebzadeh:


examples/leasing/node_release_script.py
```diff
@@ -0,0 +1,241 @@
...
+    # networking action setup by revert port in the previous step.
+    counter = 2
+    while counter:
+
+        try:
+            hil_client.project.detach(project, node)
+            print('Node `%s` removed from project `%s`' % (node, project))
+            break
+        except FailedAPICallException as ex:
+            if ex.message == 'Node has pending network actions':
```
zenhack: We should probably modify the client library to expose the type of error, rather than client programs to look at the human-readable description.

examples/leasing/node_release_script.py
```diff
@@ -0,0 +1,241 @@
...
+        print('HIL reservation failure: Unable to detach node \
+                `%s` from project `%s`' % (node, project))
+        raise HILClientFailure()
+
+    with open(statusfile, 'r') as status_file, \
+            open('/tmp/tempfile', 'w') as output_file:
+        for line in status_file:
+            words = line.split()
+            if node == words[0]:
+                newline = node + " " + "free_pool " + str(0) + "\n"
```
zenhack: Why not just `node + " free pool 0\n"`?

examples/leasing/node_release_script.py
```diff
@@ -0,0 +1,241 @@
...
+                output_file.write(newline)
+            else:
+                output_file.write(line)
```
zenhack: Rather than having the else clause, you could just over-ride the existing variable:

```python
if node == words[0]:
    line = node + " free_pool 0\n"
output_file.write(line)
```

examples/leasing/leasing.cfg
```diff
@@ -0,0 +1,8 @@
+[hil]
```
zenhack: It would make sense to add comments explaining what each of these fields is.

examples/leasing/leasing.cfg
```diff
@@ -0,0 +1,8 @@
...
+node_list: 0, 1, 2, 3, 4, 5
+url: 127.0.0.1
+threshold: 3
+user_name: USER
+password: PASSWORD
+endpoint: http://127.0.0.1:5000
+status_file: /var/log/leasing
```
zenhack: `/var/log` doesn't really make sense here; `/var/lib` would be better.

examples/leasing/leasing
```diff
@@ -0,0 +1,6 @@
+0 lease_test 1
+1 lease_test 2
+2 free_pool 0
```
zenhack: Would you document the format of this file? Not immedately obvious what each of the columns means.

examples/leasing/node_release_script.py
```diff
@@ -0,0 +1,241 @@
...
+
+    free_node_list = hil_client.node.list('free')
+    # Only these nodes should be updated in
+    # the file(either for project or for time)
+    nodes_to_update_infile = list(Set(
+        non_persistent_list)-Set(free_node_list))
+
+    for node in nodes_to_update_infile:
+        # get the information from status file for node.
+        node_info_from_file = get_project_and_time_for_node(statusfile, node)
```
zenhack: I think I suggested this when you made the initial request for feedback on your design: it would be better to just parse the file once up front, then do all of your calculations on some data structure (maybe something like `{'node_name': {'threshold': 4, 'project': 'my-project'}, ...}`?), then write it back out all at once.

Additionally, when I find myself coming up with variables that are this long, I tend to step back and ask if perhaps I'm doing things that conceptually shouldn't be coupled. If your function name is a full sentence, then you haven't really encapsulated much.

If you use the representation I suggest above, you might end up with something like:

```python
node_data = load_node_data(filename)
...

to_update = ...
for node in to_update:
    old_project = node['project']
    old_time = node['time']
    ...
```

Note that because `nodes_to_update_infile` is only used for the span of 2 statements, you probably don't need such a long name for clarity; the extra info is clear from the surrounding context.

examples/leasing/node_release_script.py
```diff
@@ -0,0 +1,241 @@
...
+        new_time = int(node_info_from_file[1])+1
+        # See if the node is in the same project and if it is outside
+        # free pool for time less than threshold value then update only
+        # time in status file
+        if (project_in_file == project_in_hil and
+                new_time < threshold_time):
+            update_time_for_node(statusfile, node, old_time, new_time)
+        # See if the node is in the same project and if it is outside
+        # free pool for time greater than threshold value then release
+        # the node back to free pool
```
zenhack: I stopped attempting to read these comments part way through, because they scan poorly (biggest problem is lack of punctuation) and the code is clearer than they are anyway. I don't think they're necessary -- they're basically just re-stating the code in English. In-line comments should provide context, explain *why*, not how -- code is better at the how.

examples/leasing/node_release_script.py
naved001: Yeah, I had written this in some other hil client, and at that time I just wanted the client to work rather than dealing with the client library itself. I'll open an issue about it.

examples/leasing/node_release_script.py
```diff
@@ -0,0 +1,241 @@
...
+        # are not matching then we need to update project in status file
+        elif (project_in_file != project_in_hil and
+                project_in_hil is not None):
+            update_project_in_status_file(
+                statusfile, node, project_in_hil,
+                project_in_file)
+
+
+def get_project_and_time_for_node(statusfile, node):
+    """Reads current information from file for every node."""
```
naved001: Kinda confused by the description and what the code seems to be doing. 

1. If this method is for returning info about one single node then:
* you should update the docstring to reflect that. 
* break the `for` loop once you find your match `node == node_status[0]`.

2. If this method is reading information for **every** node. Then you should just iterate over the file, and return a dictionary with `node_label`, `node_project`, and `node_duration`. passing `node` wouldn't be needed.

examples/leasing/node_release_script.py
```diff
@@ -0,0 +1,241 @@
...
+    node_status = ''
+    node_project = ''
+    node_duration = 0
+    with open(statusfile, 'r') as status_file:
+        for line in status_file:
+            node_status = line.split()
+            if node == node_status[0]:
+                node_project = node_status[1]
+                node_duration = node_status[2]
+    status_file.close()
```
naved001: no need to close the file if you open it in a `with` block

examples/leasing/node_release_script.py
```diff
@@ -0,0 +1,241 @@
...
+        for line in status_file:
+            words = line.split()
+            if node == words[0]:
+                words[2] = str(new_time)
+                newline = ' '.join(words)+"\n"
+                output_file.write(newline)
+            else:
+                output_file.write(line)
+    status_file.close()
+    output_file.close()
```
naved001: same thing, you opened your files in a `with` block, so these lines are redundant.

examples/leasing/node_release_script.py
```diff
@@ -0,0 +1,241 @@
...
+    move('/tmp/tempfile', statusfile)
+
+if __name__ == "__main__":
+    try:
+        config = ConfigParser.SafeConfigParser()
+        config.read("/etc/leasing.cfg")
+        node_list = [x.strip()
+                     for x in config.get('hil', 'node_list').split(',')]
+        threshold = int(config.get('hil', 'threshold'))
+        hil_url = config.get('hil', 'url')
```
naved001: unused variable

examples/leasing/node_release_script.py
```diff
@@ -0,0 +1,205 @@
...
+endpoint: the ip address and the port number
+            which hil is running on
+status_file: it should be /var/lib/leasing
+
+******************
+
+Status file format (var/lib/leasing):
+node_name project_name current_time
+
+current time is the number of times that cron jon
```
Izhmash: cron job *

examples/leasing/node_release_script.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+class StatusFileError(Exception):
+    pass
+
+
+def hil_client_connect(endpoint_ip, name, pw):
+
+    hil_http_client = RequestsHTTPClient()
+    """
```
Izhmash: Is this commented out for debug purposes?  It would be helpful to know if the client wasn't able to connect.

Izhmash: I just saw Ian's older comment, if this won't be a likely failure case we can probably just remove it entirely.

mosayyebzadeh: I will remove it, just wanted to be sure that other reviewers have the same opinion.

examples/leasing/node_release_script.py
```diff
@@ -0,0 +1,212 @@
...
+node_list: names of the nodes
+threshold: number of times that cron job is called before
+            releasing the node
+user_name:
+password:
+endpoint: the ip address and the port number
+            which hil is running on
+status_file: it should be /var/lib/leasing
+
+******************
```
naved001: Just leave pointers saying that this script requires a config file and status file, but move the actual description to the config and status file itself. You could follow [this example](https://github.com/CCI-MOC/hil/blob/master/examples/hil.cfg) on how to document those files.

examples/leasing/node_release_script.py
```diff
@@ -0,0 +1,212 @@
...
+
+Status file format (var/lib/leasing):
+node_name project_name current_time
+
+Threshold is the number of times that cron jon
+has been called.
+"""
+
+# from sets import Set
```
naved001: commented out code should be removed

examples/leasing/node_release_script.py
```diff
@@ -0,0 +1,212 @@
...
+import time
+import ConfigParser
+
+from hil.client.client import Client, RequestsHTTPClient
+from hil.client.base import FailedAPICallException
+
+
+class HILClientFailure(Exception):
+    """Exception indicating that the HIL client failed"""
+    pass
```
naved001: pass is not required when you have a docstring. but, to be honest, I don't feel too strongly about it.

examples/leasing/node_release_script.py
```diff
@@ -0,0 +1,212 @@
...
+
+class StatusFileError(Exception):
+    """Exception regarding Status File Format"""
+    pass
+
+
+def hil_client_connect(endpoint_ip, name, pw):
+    """
+    We need a client object
+    """
```
naved001: Suggestion for dosctring:

```"""Returns a HIL client object"""```

and it doesn't have to be multiline since it's really short. See the [PEP 257](https://www.python.org/dev/peps/pep-0257/#id16) guidelines.

File: examples/leasing/node_release_script.py [hunk 19]
```diff
@@ -0,0 +1,212 @@
...
+
+    hil_http_client = RequestsHTTPClient()
+    hil_http_client.auth = (name, pw)
+
+    return Client(endpoint_ip, hil_http_client)
+
+
+def load_node_info(statusfile):
+    """
+    Creating an structure containing all information from the
```
naved001: Creat**e** **a**

examples/leasing/node_release_script.py
mosayyebzadeh: I was thinking about having a README file in the folder.

naved001: that's fine too. But make sure the config/status file descriptions are in those files itself.

examples/leasing/node_release_script.py
zenhack: We've left it off elsewhere; let's be consistent.

mosayyebzadeh: Done

mosayyebzadeh:
Tested on a mock setup

mosayyebzadeh:
@zenhack Can you check this again?

examples/leasing/README
```diff
@@ -0,0 +1,45 @@
...
+are set to 1 minutes, then after 5*1 = 5 minutes, the node will
+be released.
+
+One can change the "assigned to a project" time for all nodes
+(by chanigng threshold in config file) or for one node
+(by changing the line in the status file related to the node.)
+
+Script reads a config file from /etc/leasing.cfg
+which includes the name of the nodes, time threshold,
+admin user name and password, status file and endpoint
```
zenhack: Probably worth calling out explicitly that this script doesn't support other auth backends (e.g. keystone). You can infer it from the description, but better to be clear.

examples/leasing/README
```diff
@@ -0,0 +1,45 @@
...
+which specifies the ip address and port number which hil runs at.
+
+status file includes each node's status, which node is in
+which project, and which node is free or for how long the node
+has been assigned to a project.
+
+*******************
+
+Config file format (/etc/leasing.cfg):
```
zenhack: Probably better not to duplicate this documentation; I'd just say "see the comments in `./leasing.cfg`.

examples/leasing/leasing.cfg
```diff
@@ -0,0 +1,17 @@
+[hil]
+# Config file format (/etc/leasing.cfg):
```
zenhack: I'd recommend putting the comments next to the fields they describe, e.g.

```
# names of the nodes:
node_list: 0, 1, 2, 3, 4, 5
# number of times that cron job should call the leasing
# script before releasing the node:
threshold:  3
...
```

examples/leasing/leasing.cfg
```diff
@@ -0,0 +1,17 @@
...
+# node_list: names of the nodes
+# threshold: number of times that cron job should call the leasing
+# script before releasing the node
+# user_name:
+# password:
+# endpoint: the ip address and the port number
+# which hil is running on
+# status_file: the path to the status file (e.g., /var/lib/leasing)
+
+node_list: 0, 1, 2, 3, 4, 5
```
zenhack: Maybe in the example use non-numeric names (cisco-2, dell-25, etc...). This a bit clearer to me; the numeric names suggest that the field may have deeper semantics.

examples/leasing/leasing.cfg
```diff
@@ -0,0 +1,17 @@
+[hil]
+# Config file format (/etc/leasing.cfg):
+# node_list: names of the nodes
+# threshold: number of times that cron job should call the leasing
+# script before releasing the node
+# user_name:
+# password:
```
zenhack: Should have a description for the username & password fields (should mention specifcially that they're for a HIL project with admin rights).

examples/leasing/node_release_script.py
```diff
@@ -0,0 +1,187 @@
...
+    write it back to the file. In this way, the file will be read
+    and written once.
+    """
+    nodes = {}
+    with open(statusfile, 'r') as status_file:
+        for line in status_file:
+            node_status = line.split()
+            nodes[node_status[0]] = {}
+            nodes[node_status[0]].update({'project': node_status[1],
+                                          'time': node_status[2]})
```
zenhack: These two statements could be collapsed to:

```python
nodes[node-status[0]] = {
    'project': node_status[1],
    'time': node_status[2],
}
```

examples/leasing/node_release_script.py
```diff
@@ -0,0 +1,187 @@
...
+
+    return Client(endpoint_ip, hil_http_client)
+
+
+def load_node_info(statusfile):
+    """
+    Creates a structure containing all information from the
+    file. We will change and update this data structure and then
+    write it back to the file. In this way, the file will be read
+    and written once.
```
zenhack: 1. Describe the data structure.
2. Mention that it's the return value ("creates" doesn't communicate where the data structure ends up; for all the reader knows it could be in some global variable).
3. You probably don't need to mention the bit about reading & writing once; this isn't such a deep thing that you need to talk about alternative designs.

examples/leasing/node_release_script.py
```diff
@@ -0,0 +1,187 @@
...
+            except FailedAPICallException:
+                print('Failed to revert port `%s` on node \
+                        `%s` switch `%s`' % (port, node, switch))
+                raise HILClientFailure()
+
+    time.sleep(5)
+    # tries 2 times to detach the project because there might be a pending
+    # networking action setup (revert port in the previous step).
+    counter = 2
+    while counter:
```
zenhack: Would be a bit cleaner to use a for loop:

```python
for _ in range(2):
    ...
```

... and just use return inside the try, instead of break.

examples/leasing/node_release_script.py
```diff
@@ -0,0 +1,187 @@
...
+    with open(statusfile, 'w') as status_file:
+        for node in sorted(nodes):
+            status_file.write(str(node) + ' ' + str(nodes[node]['project']) +
+                              ' ' + str(nodes[node]['time']) + '\n')
+
+
+def release_nodes(
+        statusfile, non_persistent_list,
+        threshold_time, hil_username, hil_password, hil_endpoint,
+        hil_client=None
```
zenhack: There's no real reason to have the client be optional here; I'd just create it in main and then pass it in, always. Right now the if-statement below is always executed, so the conditional logic just makes things harder to read. If you make the client parameter mandatory, you can also delete the endpoint/username/password parameters.

examples/leasing/node_release_script.py
```diff
@@ -0,0 +1,187 @@
...
+        ):
+
+    """Release nodes : After certain amount of time (threshold),
+    we should return the nodes back to the free pool.
+    Tenants who need more time should ask time extension.
+    """
```
zenhack: This isn't the right spot to describe the point of the script; you do that in the readme. Rather, describe, what *this function* does. Include a description of each of the parameters, the return value (if any), and any side effects.

examples/leasing/node_release_script.py
```diff
@@ -0,0 +1,187 @@
...
+    # Only these nodes should be updated in
+    # the file(either for project or for time)
+    nodes_to_update = list(set(
+        non_persistent_list)-set(free_node_list))
+
+    # All nodes should have information in status file
+    nodes = load_node_info(statusfile)
+
+    for node in nodes_to_update:
+        if nodes[node] is None:
```
zenhack: I don't think this is possible?

examples/leasing/node_release_script.py
```diff
@@ -0,0 +1,187 @@
...
+from hil.client.client import Client, RequestsHTTPClient
+from hil.client.base import FailedAPICallException
+
+
+class HILClientFailure(Exception):
+    """Exception indicating that the HIL client failed"""
+
+
+class StatusFileError(Exception):
+    """Exception regarding Status File Format"""
```
zenhack: You raise these exceptions in a couple places, but never catch them; probably you should add another `except` clause at the bottom of the script that prints a human-friendly error when these get raised.

examples/leasing/node_release_script.py
mosayyebzadeh: This one covers the cases that someone forgets to list "all" the nodes information in the status file.

zenhack: Ah. If you want to check if the node is absent from the dictionary, do `node not in nodes`. As written, this will just raise a `KeyError`.

mosayyebzadeh:
@zenhack Would you please check it again?

examples/leasing/node_release_script.py
```diff
@@ -0,0 +1,191 @@
...
+            print('Node `%s` removed from project `%s`' % (node, project))
+            return
+        except FailedAPICallException as ex:
+            if ex.message == 'Node has pending network actions':
+                time.sleep(2)
+            else:
+                print('HIL reservation failure: Unable to \
+                        detach node `%s` from project `%s`' % (node, project))
+                raise HILClientFailure(ex.message)
+    if counter == 0:
```
naved001: since you have changed it to a `for` loop, the value of counter at this point would be 1. so this check would fail. In fact, we don't need this check at all, if we reach this line it means failure anyway (since we would return earlier on success). Just get rid of this check.

examples/leasing/node_release_script.py
```diff
@@ -0,0 +1,191 @@
...
+            node_list,
+            int(config.get('hil', 'threshold')),
+            hil_client
+            )
+
+    except ConfigParser.NoOptionError, err:
+        print err
+    except StatusFileError():
+        print('Status file is not complete, One or \
+                more nodes infomration is missing')
```
naved001: typo: in**forma**tion

naved001:
ah, pylint is complaining because we aren't using `counter` anywhere. I guess one way to fix this would be to change it `_`. 
```python
for _ in range(2):
    blah blah
````

mosayyebzadeh:
@zenhack Would you please check it again?

zenhack:
I'm satisfied; merging.

PR #922, opened 2017-12-14
Support Public key authentication to connect to switches

naved001:
This adds support for the Dell Powerconnect switch and Cisco Nexus.

Configuring the Dell N3000 to use pubkey was a mess. I did configure it and could login using my key, it wasn't smooth though.
```
naved:~/pubkey-ssh/hil$ ssh switch.ipmi.cluster -l naved

R5-PA-C01-U39# (I PRESSED any key here)
Password:Login of naved failed (then I pressed enter)
R5-PA-C01-U39>
R5-PA-C01-U39>en

R5-PA-C01-U39#   
```

Don't know why it tells me about login failed when I did login correctly? Ideally, I would want to reset the switch config and start with a clean slate, but I can't do it since it's a shared switch.

### DIFF
--- a/hil/ext/switches/_console.py
+++ b/hil/ext/switches/_console.py
@@ -178,3 +178,25 @@ def get_prompts(console):
             'if_prompt': re.escape(cmd_prompt[:-1]) + '\(config\\-if[^)]*\)#',
             'main_prompt': re.escape(cmd_prompt),
         }
+
+
+def login(switch):
+    """login to switch using either a password or key."""
+
+    alternatives = ['User Name:', '[Pp]assword:*', '>', '#']
+    console = pexpect.spawn(
+            'ssh ' + switch.username + '@' + switch.hostname)
+
+    outcome = console.expect(alternatives)
+    if outcome == 0:
+        # some switches, like the dell powerconnect, ask for the username again
+        console.sendline(switch.username)
+        outcome = console.expect(alternatives)
+    if outcome == 1:
+        console.sendline(switch.password)
+        outcome = console.expect(alternatives)
+    if outcome == 2:
+        console.sendline('enable')
+
+    logger.debug('Logged in to switch %r', switch)
+    return console
--- a/hil/ext/switches/_console.py
+++ b/hil/ext/switches/_console.py
@@ -181,7 +181,10 @@ def get_prompts(console):
 
 
 def login(switch):
-    """login to switch using either a password or key."""
+    """Login to switch using either a password or key.
+
+    `switch` must have the attributes, `username` and `password`.
+    """
 
     alternatives = ['User Name:', '[Pp]assword:*', '>', '#']
     console = pexpect.spawn(
--- a/hil/ext/switches/n3000.py
+++ b/hil/ext/switches/n3000.py
@@ -102,7 +102,10 @@ class _DellN3000Session(_BaseSession):
 
     @staticmethod
     def connect(switch):
-        """connect to the switch and log in"""
+        """Connect to the switch and log in.
+
+        Login using public key is untested on this switch.
+        """
         console = _console.login(switch)
 
         # send a new line so that we can "expect" a prompt again if we already
--- a/hil/ext/switches/n3000.py
+++ b/hil/ext/switches/n3000.py
@@ -104,6 +104,10 @@ class _DellN3000Session(_BaseSession):
     def connect(switch):
         """connect to the switch and log in"""
         console = _console.login(switch)
+
+        # send a new line so that we can "expect" a prompt again if we already
+        # matched when logged in using pubkey
+        console.sendline('')
         prompts = _console.get_prompts(console)
 
         # create the dummy vlan for port_revert
--- a/hil/ext/switches/n3000.py
+++ b/hil/ext/switches/n3000.py
@@ -18,7 +18,6 @@ Currently the driver uses telnet to connect to the switch's console; in
 the long term we want to be using SNMP.
 """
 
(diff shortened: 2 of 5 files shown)


zenhack:
@naved, now that #923 is merged, rebasing should fix the tests.

hil/ext/switches/nexus.py
```diff
@@ -124,12 +124,22 @@ def disable_native(self, vlan_id):
...
             'ssh ' + switch.username + '@' + switch.hostname)
 
-        console.expect('Password: ')
-        console.sendline(switch.password)
+        outcome = console.expect(alternatives)
+        if outcome == 0:
+            console.sendline(switch.password)
+            outcome = console.expect(alternatives)
+        if outcome == 1:
+            console.sendline('en')
```
zenhack: It's probably worth factoring this out into a helper function, and using it in both places.

hil/ext/switches/nexus.py
```diff
@@ -124,12 +124,22 @@ def disable_native(self, vlan_id):
...
 
+        logger.debug('Logged in to switch %r', switch)
+
+        # send a new line so that we can "expect" a prompt again if we already
+        # matched when outcome was 2 (pub key auth)
+        console.sendline('')
```
zenhack: Why does this work here and not above?

naved001: I actually don't know, which is why I put the fixme there (showing the output if we just send a newline). I was hoping you could provide some insights into why this doesn't work on the powerconnect (or some way of debugging this).

Sidenote: we send the Dell n3000 a [bunch of commands](https://github.com/CCI-MOC/hil/blob/master/hil/ext/switches/n3000.py#L117-L125) when connecting to it before getting the prompts, and that works too.

zenhack: Not sure. We should at least add some comments mentioning the difference.

hil/ext/switches/nexus.py
naved001: done.

hil/ext/switches/nexus.py
naved001: Sorry but I don't know what's not clear here. I have added the comment here (Line 140, 141) and in the dell file (Line 98-102) too which describes what happens if we just send a newline.

naved001:
I have refactored the n3000 too, but I couldn't test pubkey auth on it for the reasons described in the title. It works normally with password though.

For the other 2 switches, I ran *network deployment tests with both kind of users (password, and pub key).

naved001:
Re: the prompt stuff in the dell switch. 

If we send it some random string.
```
(Pdb) p console.after
'\r\n\r% Unrecognized command\r\nconsole#'
```
our string of interest is separated by new line so we can extract it.

But if we send a new line to the switch, this is how `console.after` looks like
```naved's-made-up-language
(Pdb) p console.after
'\r\n\r\r\x1b[Kconsole#'
```
we can't get the prompt name from this since there our string of interest in not separated by new lines. 
It has x1b and [K in between, which seem like escape sequences for something (text formatting?) and I can't seem to google it.

naved001:
@zenhack I think I should remove the fixme from `dell.py` because to me it doesn't look like we can do anything about it. I'll just leave the comment there explaining what's going on.

hil/ext/switches/n3000.py
```diff
@@ -104,16 +103,13 @@ def _sendline(self, line):
...
-            'ssh ' + switch.username + '@' + switch.hostname)
+        console = _console.login(switch)
 
-        console.expect('password: ')
-        console.sendline(switch.password)
-        console.expect('>')
-        console.sendline('enable')
-
-        logger.debug('Logged in to switch %r', switch)
+        # send a new line so that we can "expect" a prompt again if we already
```
naved001: login using pubkey will most likely work on the n3000 since the `_console.login` is pretty generic. But because I haven't tested it, I am not documenting/claiming that it will work.

hil/ext/switches/_console.py
```diff
@@ -178,3 +178,25 @@ def get_prompts(console):
             'if_prompt': re.escape(cmd_prompt[:-1]) + '\(config\\-if[^)]*\)#',
             'main_prompt': re.escape(cmd_prompt),
         }
+
+
+def login(switch):
+    """login to switch using either a password or key."""
```
zenhack: Should mention the requirement for username and password attributes on the switch; All of the switches we're using with this happen to have that attribute, but there's nothing defined by the Switch class that says they need to.

hil/ext/switches/n3000.py
zenhack: `_console.login` specifies that it does pubkey auth, so a reader of this code might naturally assume that it works unless stated otherwise. Ideally we'd just test it so this could be a moot point, but otherwise I'd at least like a comment mentioning that it's untested.

naved001:
@zenhack updated the comments and docstring. 

@mosayyebzadeh could you give this a second review?

mosayyebzadeh:
Everything seems fine to me. Just change the document and I will approve it.

naved001:
>  Just change the document and I will approve it.

@mosayyebzadeh what do you want changed in the documents?

mosayyebzadeh:
Highlight this part in the document or make it noticeable: "There should be no "enable" password."

naved001:
@mosayyebzadeh done. Also, I noticed that it was in the wrong section; so I put it up on top because it's applicable for all switches.

zenhack:
LGTM, merging.

Issue #910, opened 2017-11-01
Fix the CLI/client-library to print out appropriate error messages


naved001:
More often than not, people get this error when using HIL "Unexpected error: No JSON object could be decoded". I think this error might be coming from the client library. This can be frustrating to deal with at times because the real error is hidden.

naved001:
one example that I found today.

```
[centos@slurm-controller ~]$ hil show_node blah-blah
Error: Node blah-blah does not exist.

[centos@slurm-controller ~]$ hil show_node blah/blah
Unexpected error: No JSON object could be decoded
```

if there's a '/' in node name it freaks out.

zenhack:
Here's what I assume going on in this specific case:

* putting a `/` in the node name is causing it to not match the route for that api call; we had to do something to special on the server side to get it to treat e.g. `gi1/0/2` as a single thing.
* Since the route isn't matching, the server is returning some default error page, probably a 404. Since the body isn't JSON, the client trys to parse it and freaks out.

Things we can do:

* The client should validate its arguments, so that if you pass it the name of a node, it can't be treated as anything else. We should be doing this anyway, as it might help prevent injection attacks where some tool using the client library passes in a user-supplied string to the api call.
* The client should look at the http status code, rather than just choking on the body if it's not JSON.
* The server's default error pages should return JSON, just like the errors raised explicitly by the HIL api calls.

Izhmash:
I'll be investigating this one.

Izhmash:
If we want users to be able to put `/` characters in their node names, we'll need to put `path:` in the API URL template, e.g. `@rest_call('GET', '/node/<path:nodename>', Schema({'nodename': basestring}))`.

Instead, would it be better to check for reserved characters and raise an error if one is detected?  That way we would prevent the 404 from causing `Unexpected error: No JSON object could be decoded`. I think it's reasonable to limit names to letters and numbers.

Izhmash:
In the `show_node` case, we could infer that a node doesn't exist if the user tries to search for a node with a reserved character.  The same would go for switches, projects, etc.

zenhack:
Yeah, my intention was for the client library to check and report an error itself.

Izhmash:
I can either catch all attempts of the user to input reserved URL characters when the URL is first formed, or I can do it on a per-command basis.  For example if the user tries to search for a node with a `/` in it, HIL would report that the node could not be found.

The first solution would be simpler but less informative.

zenhack:
[...]

PR #926, opened 2017-12-15
Improve JSON-less 404 error readability and prevention with client library

Izhmash:
This code aims to stop the `Unexpected error: No JSON object could be decoded` error mentioned in #910:
- The error is caused by 404 pages without JSON hitting `json.loads()`.
- If `json.loads()` fails, print out the response content rather than throwing the error above.
- Since this isn't pretty, the following target 404's:
    - Characters that get passed to the API will be limited to numbers, letters, and `$ - _ . + ! * ' ( ) ,`
    - The client library checks object names per-call for characters that don't match this.
- If the client library detects bad characters, it will print something like: `Error: Projects may not contain: ['/', '&']` to alert the user of the illegal characters that they entered.

 I anticipate that the tests will take a while to write since I am introducing changes to the majority of the client library calls.  Thus, I'm introducing this code now so I can get feedback before writing the tests.

Note: some commands, such as `switch_register`, have not been implemented in the client library yet.   These will get checks for reserved characters when they are implemented.

### DIFF
--- a/hil/client/base.py
+++ b/hil/client/base.py
@@ -4,7 +4,6 @@ from urlparse import urljoin
 import json
 import re
 from hil.errors import BadArgumentError
-import inspect
 
 
 class FailedAPICallException(Exception):
@@ -69,7 +68,7 @@ class ClientBase(object):
                 error_type=e['type'],
                 message=e['msg'],
             )
-        # Catching 404's that do not return JSON
+        # Catching responses that do not return JSON
         except ValueError:
             return response.content
 
@@ -96,7 +95,6 @@ def check_reserved_chars(*outer_args, **outer_kwargs):
     and dynamically report the error by the offending argument(s)"""
     def wrapper(f):
         """Auxiliary wrapper for check_reserved_chars"""
-        argspec = inspect.getargspec(f)
 
         def reserved_wrap(*args, **kwargs):
             """Wrapper that is passed the arguments of the wrapped function"""
--- a/hil/client/base.py
+++ b/hil/client/base.py
@@ -76,12 +76,12 @@ class ClientBase(object):
 
 def _find_reserved(string, slashes_ok=False):
     """Returns a list of illegal characters in a string"""
-    if slashes_ok:
-        p = r"[^A-Za-z0-9 /$_.+!*'(),-]+"
-    else:
-        p = r"[^A-Za-z0-9 $_.+!*'(),-]+"
-    # return all unique chars that exist neither in p nor string
-    return list(set(x for l in re.findall(p, string) for x in l))
+    p = r"[^A-Za-z0-9 /$_.+!*'(),-]"
+    result = re.findall(p, string)
+    if not slashes_ok and '/' in string:
+        result.append('/')
+    # return one of each illegal character in string
+    return list(set(result))
 
 
 def check_reserved(obj_type, obj_string, slashes_ok=False):
--- a/hil/client/base.py
+++ b/hil/client/base.py
@@ -73,11 +73,11 @@ class ClientBase(object):
 
     def find_reserved(self, string):
         """Returns a list of illegal characters in a string"""
-        p = '[^A-Za-z0-9 \$\-\_\.\+\!\*\'\(\)\,]+'
+        p = r"[^A-Za-z0-9 $_.+!*'(),-]+"
         return list(x for l in re.findall(p, string) for x in l)
 
     def find_reserved_w_slash(self, string):
         """Returns a list of illegal characters in a string
         excluding `/` for channels and ports"""
-        p = '[^A-Za-z0-9 \/\$\-\_\.\+\!\*\'\(\)\,]+'
+        p = r"[^A-Za-z0-9 /$_.+!*'(),-]+"
         return list(x for l in re.findall(p, string) for x in l)
--- a/hil/client/base.py
+++ b/hil/client/base.py
@@ -3,6 +3,7 @@
 from urlparse import urljoin
 import json
 import re
+from hil.errors import BadArgumentError
 
 
 class FailedAPICallException(Exception):
@@ -71,13 +72,17 @@ class ClientBase(object):
         except ValueError:
             return response.content
 
-    def find_reserved(self, string):
+    def _find_reserved(self, string, slashes_ok=False):
         """Returns a list of illegal characters in a string"""
-        p = r"[^A-Za-z0-9 $_.+!*'(),-]+"
+        if slashes_ok:
+            p = r"[^A-Za-z0-9 /$_.+!*'(),-]+"
+        else:
+            p = r"[^A-Za-z0-9 $_.+!*'(),-]+"
         return list(x for l in re.findall(p, string) for x in l)
 
-    def find_reserved_w_slash(self, string):
-        """Returns a list of illegal characters in a string
-        excluding `/` for channels and ports"""
-        p = r"[^A-Za-z0-9 /$_.+!*'(),-]+"
-        return list(x for l in re.findall(p, string) for x in l)
+    def check_reserved(self, obj_type, obj_string, slashes_ok=False):
+        """Check for illegal characters and report of their existence"""
+        bad_chars = self._find_reserved(obj_string, slashes_ok)
+        if bool(bad_chars):
+            error = obj_type + " may not contain: " + str(bad_chars)
+            raise BadArgumentError(error)
--- a/hil/client/base.py
+++ b/hil/client/base.py
@@ -74,10 +74,10 @@ class ClientBase(object):
     def find_reserved(self, string):
         """Returns a list of illegal characters in a string"""
         p = '[^A-Za-z0-9 \$\-\_\.\+\!\*\'\(\)\,]+'
-        return set(x for l in re.findall(p, string) for x in l)
+        return list(x for l in re.findall(p, string) for x in l)
 
     def find_reserved_w_slash(self, string):
         """Returns a list of illegal characters in a string
-        including `/` for channels and ports"""
+        excluding `/` for channels and ports"""
         p = '[^A-Za-z0-9 \/\$\-\_\.\+\!\*\'\(\)\,]+'
-        return set(x for l in re.findall(p, string) for x in l)
+        return list(x for l in re.findall(p, string) for x in l)
--- a/hil/client/base.py
+++ b/hil/client/base.py
@@ -66,5 +66,6 @@ class ClientBase(object):
                 error_type=e['type'],
                 message=e['msg'],
             )
+        # Catching 404's that do not return JSON
         except ValueError:
             return response.content
--- a/hil/client/base.py
+++ b/hil/client/base.py
@@ -4,6 +4,7 @@ from urlparse import urljoin
 import json
 import re
 from hil.errors import BadArgumentError
+import inspect
 
 
 class FailedAPICallException(Exception):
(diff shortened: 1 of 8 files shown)


hil/client/network.py
```diff
@@ -24,6 +29,22 @@ def create(self, network, owner, access, net_id):
...
             See docs/networks.md for details.
             """
+            bad_chars = self.find_reserved(network)
+            if bool(bad_chars):
+                raise BadArgumentError("Networks may not contain: %s"
+                                       % bad_chars)
+            bad_chars = self.find_reserved(owner)
+            if bool(bad_chars):
+                raise BadArgumentError("Owner may not contain: %s"
+                                       % bad_chars)
```
zenhack: We could probably do a better job of factoring this out. maybe:

```python
check_reserved('Owner', owner)
```

with a suitable definition of `check_reserved`.

We could go a bit further and implement a decorator so we could do something like:

```python
@check_reserved_chars('network', 'owner', 'access', slashes_ok=['net_id'])
def create(self, network, owner, access, net_id):
    # Implementation unchanged from master.
```

It might also make sense to survey the arguments and see what the proportion of arguments to be checked vs. not checked; depending, we may want to specify arguments *not* to check, rather than the other way around.

hil/client/base.py
```diff
@@ -60,9 +61,23 @@ def check_response(self, response):
...
                 error_type=e['type'],
                 message=e['msg'],
             )
+        # Catching 404's that do not return JSON
+        except ValueError:
+            return response.content
+
+    def find_reserved(self, string):
+        """Returns a list of illegal characters in a string"""
+        p = '[^A-Za-z0-9 \$\-\_\.\+\!\*\'\(\)\,]+'
```
zenhack: 1. I don't think most of these need escaping when inside a character class. The dash can go at the end of the list to avoid it being interpreted as special. Escaping the single quote can be avoided by using double quotes to delmit the string as a whole.
2. Generally speaking, use `r"raw strings"` for regexes, so you don't have to worry about what needs to be escaped in a python string as well.

Izhmash: Ok, I'll work on that.

hil/client/network.py
Izhmash: I like the decorator solution, I'm going to work on that.

naved001: @Izhmash let me know when you do the decorator solution, I'll review this after that.

Izhmash:
Just wanted to note that I'm back working for the new semester so I'll have activity on here soon.

hil/client/base.py
Izhmash: Thanks for the tip, https://github.com/CCI-MOC/hil/pull/926/commits/c1fea672dff1a55aa9504c82e45bde5d5bb90389 addresses this.

hil/client/network.py
Izhmash: @zenhack how would you relate the arguments in `create()` to the types in the decorator? Since the arguments are passed as `*args` the only way I can see to relate them would be by their order which might not be robust enough.  It would work out if the arguments were changed to be keyworded but that would be editing the original function.

xuhang57: I am not sure whether this is right but: what you need is a function that checking the error, for example 

```
def decorator(argument):
    def real_decorator(function):
        def wrapper(*args, **kwargs):
            funny_stuff()
            something_with_argument(argument)
            function(*args, **kwargs)
            more_funny_stuff()
        return wrapper
    return real_decorator
```

> https://stackoverflow.com/questions/5929107/decorators-with-parameters

xuhang57: And 
```
@check_reserved_chars(arguments)
class Network(ClientBase):
```
should work so you don't have to put the decorator in every function.

Izhmash: @xuhang57 thanks for the input.  The code you presented above definitely fits what I had in mind. The issue lies with relating the non-keyworded arguments in the decorator to the arguments in the methods such as `create()` so we can choose what gets checked.

The only way that I could see to relate (for example) the string `'network'` in the decorator to the `network` passed to the method would be by location since network names have no standard in HIL.  That, then, becomes complicated when considering that some of the objects will allow slashes with the `slashes_ok` flag above.

In the meantime I'm going to work on factoring out my original code and present a solution without decorators.

Izhmash:
My latest commit cleans up the methods for checking for reserved characters.

hil/client/base.py
```diff
@@ -71,13 +72,19 @@ def check_response(self, response):
...
-        """Returns a list of illegal characters in a string
-        excluding `/` for channels and ports"""
-        p = r"[^A-Za-z0-9 /$_.+!*'(),-]+"
-        return list(x for l in re.findall(p, string) for x in l)
+        if slashes_ok:
+            p = r"[^A-Za-z0-9 /$_.+!*'(),-]+"
+            return list(x for l in re.findall(p, string) for x in l)
+        else:
+            p = r"[^A-Za-z0-9 $_.+!*'(),-]+"
+            return list(x for l in re.findall(p, string) for x in l)
```
naved001: you could just do a single return after the if-else block

hil/client/base.py
```diff
@@ -71,13 +72,19 @@ def check_response(self, response):
...
+
+    def check_reserved(self, obj_type, obj_string, slashes_ok=False):
+        """Check for illegal characters and report of their existance"""
```
naved001: exist**e**nce

hil/client/base.py
Izhmash: Fixed

hil/client/base.py
Izhmash: Fixed

Izhmash:
The tests for this are in progress now.

hil/client/network.py
zenhack: The inspect module may be of interest:

https://docs.python.org/2/library/inspect.html

from there, you can do something like:

```python
argspec = inspect.getargspec(f)
def wrapper(*args):
    for argname, argval in zip(argspec.args, args):
        if argname not in slashes_ok:
            check(argval)
```

Maybe adding some similar logic for kwargs.

Izhmash:
I've setup decorators/wrappers to check the illegal characters now, so the client functions should remain otherwise unedited.

hil/client/base.py
```diff
@@ -60,9 +63,50 @@ def check_response(self, response):
...
+def _find_reserved(string, slashes_ok=False):
+    """Returns a list of illegal characters in a string"""
+    if slashes_ok:
+        p = r"[^A-Za-z0-9 /$_.+!*'(),-]+"
+    else:
+        p = r"[^A-Za-z0-9 $_.+!*'(),-]+"
+    return list(x for l in re.findall(p, string) for x in l)
+
+
+def check_reserved(obj_type, obj_string, slashes_ok=False):
```
xuhang57: I know this serves as a decorator and you have tested the functions with using this decorator. But do you guys think we need a direct unit test with this function? I don't have much experience. Any idea?

Izhmash: My idea was to send some bad input and check to make sure it triggers the correct error as I did in `tests/unit/client.py`.  I'm not sure if it needs to be tested directly besides that; I'd like some input too.

Izhmash:
I've addressed the requests and added a first pass at the tests.  Did I address your requests correctly @zenhack and @naved001 ?

hil/client/base.py
```diff
@@ -60,9 +62,51 @@ def check_response(self, response):
...
+    and dynamically report the error by the offending argument(s)"""
+    def wrapper(f):
+        """Auxiliary wrapper for check_reserved_chars"""
+
+        def reserved_wrap(*args, **kwargs):
+            """Wrapper that is passed the arguments of the wrapped function"""
+            if 'slashes_ok' in outer_kwargs:
+                slashes_ok = outer_kwargs.get('slashes_ok')
+            else:
+                slashes_ok = []
```
zenhack: This if/else can just be `outer_kwargs.get('slashes_ok', [])`

hil/client/base.py
```diff
@@ -60,9 +62,51 @@ def check_response(self, response):
...
+            for argname, argval in zip(outer_args, args[1:]):
```
zenhack: Took me a second to work out why we were chopping off the first element; probably should add a comment about that.

hil/client/base.py
```diff
@@ -60,9 +62,51 @@ def check_response(self, response):
...
+                if argname not in slashes_ok:
+                    check_reserved(argname, argval)
+                else:
+                    check_reserved(argname, argval, slashes_ok=True)
```
zenhack: Could simplify this to:

```python
check_reserved(argname, argval, slashes_ok=argname in slashes_ok)
```

hil/client/node.py
```diff
@@ -14,6 +15,7 @@ def list(self, is_free):
         url = self.object_url('nodes', is_free)
         return self.check_response(self.httpClient.request('GET', url))
 
+    @check_reserved_chars('node')
     def show(self, node_name):
```
zenhack: You've got a mismatch between the name you're passing to the decorator and the actual name of the parameter. Similar thing in a bunch of places below.

naved001:
I don't have any other comments to make on this at this point, @zenhack has already made some good points.

hil/client/base.py
```diff
@@ -60,9 +62,51 @@ def check_response(self, response):
...
+            return response.content
+
+
+def _find_reserved(string, slashes_ok=False):
+    """Returns a list of illegal characters in a string"""
+    if slashes_ok:
+        p = r"[^A-Za-z0-9 /$_.+!*'(),-]+"
+    else:
+        p = r"[^A-Za-z0-9 $_.+!*'(),-]+"
+    return list(x for l in re.findall(p, string) for x in l)
```
naved001: well, except for this line. There's a lot going on here, would be nice if we split it to make it more readable. But I won't block the PR on this.

hil/client/node.py
Izhmash: I figured there were some places where the variable names wouldn't want to get out. I put 'node' there because otherwise in the error reporting it would say 'node_name' to the user.  Is that preferable?

Izhmash: To settle this I'm just going to use argspec and match the variable names directly.

hil/client/base.py
Izhmash: I might be wrong, but I think this is one of those cases where the list comprehension is actually less complex than if I didn't use a one-liner.  I can add a comment for clarification if that helps.

naved001:
actually, don't worry about it.

Thanks
Naved

Izhmash:
@zenhack I believe I've addressed all of your requests in these two commits

hil/client/base.py
```diff
@@ -79,7 +80,8 @@ def _find_reserved(string, slashes_ok=False):
         p = r"[^A-Za-z0-9 /$_.+!*'(),-]+"
     else:
         p = r"[^A-Za-z0-9 $_.+!*'(),-]+"
-    return list(x for l in re.findall(p, string) for x in l)
+    # return all unique chars that exist neither in p nor string
+    return list(set(x for l in re.findall(p, string) for x in l))
```
zenhack: It took me >1 min to figure out how this worked (I think partially just because there was a lot of seeking involved in parsing the comprehension). What about:

1. Change the patterns to remove the `+`
2. `list(set(re.findall(p, string)))`

Also, I don't like the fact that we have two nearly identical regexes. One alternative:

```python
p = # regex without the /
result = # expression with re.findall
if not slashes_ok and '/' in string:
    result.append('/')

Comment (Izhmash) [reply hil/client/base.py]: @zenhack my code now should reflect your comments.  My logic is slightly different from what you recommended, but works just as it did before.

File: hil/client/base.py [hunk 12]
```diff
@@ -76,12 +76,12 @@ def check_response(self, response):
 
 def _find_reserved(string, slashes_ok=False):
     """Returns a list of illegal characters in a string"""
-    if slashes_ok:
-        p = r"[^A-Za-z0-9 /$_.+!*'(),-]+"
-    else:
-        p = r"[^A-Za-z0-9 $_.+!*'(),-]+"
-    # return all unique chars that exist neither in p nor string
-    return list(set(x for l in re.findall(p, string) for x in l))
+    p = r"[^A-Za-z0-9 /$_.+!*'(),-]"
```
zenhack: This should probably *not* include the slash I think? Otherwise the slashes will end up in the result regardless of the value of `slashes_ok`. Would you also add a test that catches this?

Izhmash: I thought that result shouldn't check slashes by default so I put one in my "allowed characters". Then, if slashes aren't allowed but there is one in the string, I need to add a slash to my list of unique illegal characters, result. My goal is for result to have all the characters that are in string but not in p.

My tests all failed when I did remove the slash from p, but when I put it back everything worked as it should.  Tomorrow I'll double check that and write a test to check for a proper response.

zenhack: Ah, no you're totally right; I had things backwards in my head; forgot that the regex was being inverted, so would match allowed characters rather than banned characters.

zenhack:
LGTM

PR #927, opened 2017-12-18
Ci cleanup

naved001:
Some of the yellow text in the ci was bothering me for a while so decided to clean it up before it becomes failures in the future.

The commit messages have info in them.

One message that is still annoying is this:

```
=============================================================================================== warnings summary ===============================================================================================
None
  pytest-catchlog plugin has been merged into the core, please remove it from your requirements.

-- Docs: http://doc.pytest.org/en/latest/warnings.html

```

I removed pytest-catchlog from the requirements, but then I couldn't import pytest_catchlog. Their [docs](https://pypi.python.org/pypi/pytest-catchlog) don't say anything about it being merged into the core, it still tells you to install it explicitly. Don't know what's up with that.


docs/testing.md
```diff
@@ -42,9 +42,9 @@ To run just a subset of them, specify a particular file or directory:
     py.test tests/unit
     py.test tests/unit/api.py
 
-To run just a pep8 test:
+To run just a pep8/pycodestyle test:
```
mosayyebzadeh: Why not just pycodestyle?

zenhack: Yeah, we should just replace the name. Let's rename the files too (so pep8_tests.sh -> pycodestyle_tests.sh; not sure if there are any others).

ci/keystone/install.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env sh
 set -ex
 
-pip install .[keystone-auth-backend,keystone-cli]
```
zenhack: This referring to the extra in our `setup.py`, so yeah, it's supposed to be `keystone-client`. You could concievably want one and not the other, since the latter is used by our client library & command line tool, while the former is used by the server.

I think we should put this back, fixing the name.

naved001: I goofed up in the commit message for that one. What I meant was, since we are installing keystonemiddleware for our ci tests, it will also install python-keystoneclient. So installing keystone-client is redundant for our ci tests. But it's still left in our setup.py because, like you pointed out, we might need it for client library or cli.

zenhack: Hmm. I guess I don't like leaving it out, since the original expresses intent more clearly; it's kindof a coincidence that it works without it.

naved001:
Note: only the tool name has changed from `pep8` to `pycodestyle` (they had a slightly [heated debate](https://github.com/PyCQA/pycodestyle/issues/466)). 

The document name PEP-8 still exists, and idk if that will be renamed or not.  

I also added one more commit to use `caplog.records` property.

@zenhack @mosayyebzadeh made the changes you guys requested.

zenhack:
LGTM

naved001:
2 sign offs, merging!

Issue #427, opened 2015-05-15
Need async error checking mechanism in API


henn:
Currently, network operations are asynchronous, placed into a queue that  `serve_networks()` processes. Although some errors will be returned immediately to the caller (like permissions issues or invalid parameters), those resulting from the network operations currently have no way of reporting back to the user.

A mechanism for doing so, such as status polling, is needed. This could function by including a new API call and augmenting the `NetworkingAction` model with a status (like PENDING/SUCCESS/ERROR) and an error text field. If we added a completed timestamp, we could even have `serve_networks()` clean up the old entries based on some tunable period.

For this to work, we'll also need the drivers to generate parse-able exceptions. For that, please see #54.

henn:
We're going to address this using the mechanism in https://github.com/CCI-MOC/haas/wiki/Asynchronous-operations

zenhack:
I think instead of having a success status/cleanup timeout, it would make more sense to just delete them and have the client interpret 404 as success -- it will have already gotten receipt of successful enqueuing, so this is the only possibility.

Proposal:

* Operations that create networking actions return some kind of identifier/ticket
* Add a show_networking_action or such that shows the current state -- either pending or error. When the action completes successfully it will be removed, so this will return a 404, as above.
* add a cancel_networking_action that removes a networking action from the queue -- the primary use being to clear out failed actions. But it's just deleting a database object, so I see no reason to restrict it.
* (possibly) add a list_networking_actions that can be either per-project or system wide (the latter being admin-only).

SahilTikale:
I and @zenhack  discussed this couple of weeks back. 
I like the proposal.

zenhack:
Yeah, this is basically exactly what @SahilTikale and I talked about.

xuhang57:
PR merged. Close this now. Please reopen if necessary

Issue #510, opened 2015-12-03
api returns to user when nic is busy


jmcadden:
Consecutive "node_detach_network" requests to the client api can result in a fatal "nic is busy" error. This error seems unnecessary for the user, and the behavior incorrect. Ideally, the api would block until the existing task is finished before continuing. 

```
>$ haas node_detach_network cisco-44 enp130s0f0 net0

>$ haas node_detach_network cisco-44 enp130s0f0 net1
Unexpected status code: 409
Response text:
{"msg": "A networking operation is already active on the nic.", "type": "BlockedError"
```

xuhang57:
PR merged. Close this now. Please reopen if necessary

PR #929, opened 2018-01-07
Query networking actions

### LINKED ISSUE #486: Query pending operations (on nic)
We should be able to query the existence of pending network operations on a
switch. This would be useful for users in general, but is also something that is
important for implementing recursive HaaS correctly.

@gsilvis, @SahilTikale, @okrieg, @lokI8 and I just had a conversation about the
async questions that needed to be resolved, and what we settled on was this:
We'd add this query api call, and then the recursive HaaS switch driver would
invoke the base HaaS's api to do the network operating, and use this call to
poll, not returning until the operating has gone through.

@gsilvis has some misgivings about fudging it to look synchronous (I'll let him
describe them in more detail), but agrees that we can go ahead with this for
now

naved001:
I think it's pretty much done, except migration scripts/tests. But I figured it might be a good idea to get some feedback while I work on that.

* Fix #486 and* #427 since they are duplicate issues.


naved001:
one thing to note is that; if we were to allow queuing multiple actions on a nic later on, we would need to think about things like:
* for how long do we keep completed actions
* or how many completed actions do we keep
* do we keep them in the same NetworkAction table or a new one for completed actions.

I was looking at #510 and that reminded me that @apoorvemohan strongly believes that we should let users queue multiple ops on nics, and that users should deal with the consequences if any of their network actions fail.

zenhack:
I think figuring out multiple queued network actions can safely be treated as a separate issue. I have a couple ideas about the design.

I'll do a review of this when I'm at my computer; on my phone right now.

hil/api.py
```diff
@@ -130,7 +131,8 @@ def project_detach_node(project, node):
     if num_attachments != 0:
         raise errors.BlockedError("Node attached to a network")
     for nic in node.nics:
-        if nic.current_action is not None:
+        if nic.current_action is not None and \
+           nic.current_action.status == 'PENDING':
```
zenhack: We should probably fail if there's an action in an error state too, no?

hil/api.py
```diff
@@ -1099,13 +1115,40 @@ def port_revert(switch, port):
...
+
+# is this only going to be for statuses of Networking actions? What about in
+# the future we have some other calls whose status we want to check?
+@rest_call('GET', '/status/<status_id>', Schema({'status_id': basestring}))
+def get_status(status_id):
+    """Returns the status of the networking action by findind the status_id
+    in the networking actions table
+    """
+    # is auth required here? UUID is obscure enough that nobody can guess
+    # and see someone else's status of their operations. Moreover, it doesn't
```
zenhack: The uuid generation we're using here is *not* cryptographically random, so we should not assume it is unguessable.

I'd rather play it safe and do an authorization check than fuss about whether the information is sensitive. Probably the appropriate thing to check is either admin access or access to the project of the attached node.

naved001: > Probably the appropriate thing to check is either admin access or access to the project of the attached node.

Okay, ill do a check for access to project.

hil/api.py
naved001: If we fail when the last action errored out, how will we ever retry then? I am imagining people retrying a couple of times before notifying admins that their requests are failing; and once fixed they try again.

docs/rest_api.md
```diff
@@ -1310,3 +1330,31 @@ Remove a user from a project.
 Authorization requirements:
 
 * Administrative access.
+
+#### get_status
```
zenhack: It might be nice to (1) provide more information about the action (type, nic, network etc), and (2) make this similar to the (other) `show_*` calls. So this would be something like `show_networking_action`, the URL should probably be e.g. `/networking_action/<label>`. The only difference here is that the label is chosen by HIL rather than the caller.

hil/migrations/versions/76529f0f9e50_add_uuid_and_status_to_networkingaction.py
```diff
@@ -0,0 +1,42 @@
...
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('networking_action', sa.Column('status', sa.String(),
+                  nullable=True))
+    op.add_column('networking_action', sa.Column('uuid', sa.String(),
+                  nullable=True))
+    op.create_index(op.f('ix_networking_action_uuid'), 'networking_action',
+                    ['uuid'], unique=False)
+    op.execute("UPDATE networking_action SET status = 'PENDING', \
+        uuid ='no-uuid'")
```
zenhack: I guess I'd prefer something a bit more explict about why it's not really a uuid. Alternatively, we could go through and generate actual uuids for existing actions. Obviously they wouldn't be usable since no-one would know them, but it avoids having the table in a conceptually odd state untill everything else goes through.

hil/migrations/versions/76529f0f9e50_add_uuid_and_status_to_networkingaction.py
```diff
@@ -0,0 +1,42 @@
...
+# revision identifiers, used by Alembic.
+revision = '76529f0f9e50'
+down_revision = '9089fa811a2b'
+branch_labels = None
+
+# pylint: disable=missing-docstring
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
zenhack: These comments should be removed.

hil/migrations/versions/76529f0f9e50_add_uuid_and_status_to_networkingaction.py
```diff
@@ -0,0 +1,42 @@
...
+    op.alter_column('networking_action', 'uuid', nullable=False)
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_index(op.f('ix_networking_action_uuid'),
+                  table_name='networking_action')
+    op.drop_column('networking_action', 'uuid')
+    op.drop_column('networking_action', 'status')
```
zenhack: Observation: this will do strange things if there are actions in the table whose status is not `'PENDING'`. I'm Ok just reporting an error in that case, but we shouldn't just silently corrupt the information.

hil/model.py
```diff
@@ -587,6 +587,13 @@ class NetworkingAction(db.Model):
 
     id = db.Column(BigIntegerType, primary_key=True)
 
+    # UUID of a networking action. Useful for querying the status of a
+    # networking action.
+    uuid = db.Column(db.String, nullable=False, index=True)
+
+    # status of the operation
```
zenhack: Should probably mention possible values here.

tests/unit/client.py
```diff
@@ -589,7 +598,10 @@ def test_port_revert(self):
                 'node': 'node-01',
                 'nic': 'eth0',
                 'networks': {'vlan/native': 'net-01'}}
-        assert C.port.port_revert('mock-01', 'gi1/0/1') is None
+        response = C.port.port_revert('mock-01', 'gi1/0/1')
+        uuid_pattern = re.compile(
+            "[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}")
```
zenhack: We should probably factor out this regex and put it somewhere in test_common. I think it's used in some of the migration tests as well.

hil/api.py
zenhack: The scenario I'm worried about is something like this:

1. A user submits a newtorking action before going on their lunch break
2. The networking action errors out.
3. Since it's been like 30 min since the first one, they doesn't bother to check the status before submitting the second.
4. Networking actions have now happened out of order.

(4) could be a potential security concern, if e.g. the two actions were a remove of one node and an add of another, when the two nodes are not supposed to have been able to reach one another.

One approach would be to have the networking daemon retry a couple times itself.

hil/deferred.py
```diff
@@ -52,24 +54,38 @@ def modify_port(self, action):
...
     def revert_port(self, action):
         """Apply a revert_port action."""
         session = self.get_session(action.nic.port.owner)
-        session.revert_port(action.nic.port.label)
-        model.NetworkAttachment.query.filter_by(nic=action.nic).delete()
+        try:
+            session.revert_port(action.nic.port.label)
+            model.NetworkingAction.query.filter_by(id=action.id). \
+                update({"status": "DONE"})
+            # model.NetworkingAction.query.filter_by(id=action.id).delete()
```
zenhack: This should be deleted.

naved001:
@zenhack I have addressed some of the comments, I'll fix the the other issues too. 

One thing that I need your help with is the migration tests. I am not exactly sure how to write the migration tests (and the fix the old ones). There's already a test that runs for the networking action table. Like there's a method which says "it generated the sql dump" in the migrations folder? I thought that was generated with the pg_dump command. just little things like that, maybe I'll ping you on irc tomorrow if you are online.

naved001:
@mosayyebzadeh @SahilTikale Can either of you be a second reviewer on this?

tests/unit/api/main.py
```diff
@@ -2376,3 +2385,80 @@ def test_node_power_cycle_force(self):
...
+        response = json.loads(response[0])
+        status_id = response['status_id']
+
+        response = json.loads(api.show_networking_action(status_id))
+        assert response['status'] == 'PENDING'
+        assert response['node'] == 'node-99'
+        assert response['nic'] == '99-eth0'
+        assert response['type'] == 'modify_port'
+        assert response['channel'] == 'vlan/native'
+        assert response['new_network'] == 'hammernet'
```
naved001: ah, it just hit me that I could compare the whole json blob, smh.

zenhack:
re: dump generation, the wording is a bit sloppy; what is meant is that the function was used to create the objects that were in the database when pg_dump was run.

hil/migrations/versions/76529f0f9e50_add_uuid_and_status_to_networkingaction.py
zenhack: Also, I think generating a uuid here would actually fix the migration tests; there's already code in there to normalize things that look like uuids, so as long as we end up with a UUID in this column, it should work.

tests/unit/api/main.py
```diff
@@ -2409,12 +2408,10 @@ def test_show_networking_action_attach(self,):
...
-        assert response['status'] == 'PENDING'
-        assert response['node'] == 'node-99'
-        assert response['nic'] == '99-eth0'
-        assert response['type'] == 'modify_port'
-        assert response['channel'] == 'vlan/native'
-        assert response['new_network'] == 'hammernet'
+        assert response == {'status': 'PENDING', 'node': 'node-99',
+                            'nic': '99-eth0', 'type': 'modify_port',
+                            'channel': 'vlan/native',
+                            'new_network': 'hammernet'}
```
zenhack: This would be more readable with one key-value pair per line.

hil/api.py
```diff
@@ -1131,27 +1131,31 @@ def port_revert(switch, port):
 
     db.session.add(action)
     db.session.commit()
-    return json.dumps({'status_id': unique_id})
+    return json.dumps({'status_id': unique_id}), 202
```
zenhack: 202 doesn't make sense here; it indicates that processing is incomplete, but the request "get the status of ..." is complete.

hil/api.py
```diff
@@ -1131,27 +1131,31 @@ def port_revert(switch, port):
...
     get_auth_backend().require_project_access(project)
 
-    return json.dumps({'status': action.status})
+    action_info = {'status': action.status,
+                   'node': action.nic.owner.label,
+                   'nic': action.nic.label,
+                   'type': action.type,
+                   'channel': action.channel,
+                   'new_network': None if action.new_network is None
+                   else action.new_network.label}
```
zenhack: The formatting of the if expression is a bit hard to read; the else should probably be indented a bit further at a minimum (otherwise it reads as being at the same level in the AST as the dictionary keys, as opposed to beneath them).

I also personally find python's syntax for expression-ifs really hard to read in general, but won't block over it.

hil/test_common.py
```diff
@@ -30,6 +30,10 @@
 import sys
 import os.path
 import logging
+import re
+
+uuid_pattern = re.compile(
+            "[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}")
```
zenhack: We should also use this definition in the migration tests `tests/unit/migrations.py`, which currently define an equivalent regex in-line.

hil/api.py
naved001: > it indicates that processing is incomplete

But that is true here (`revert_port`), isn't it? The other network calls `node_{connect, detach}_network` also return 202.

zenhack:
Quoting Naved Ansari (2018-01-11 10:07:42)

>    But that is true here (revert_port), isn't it? The other network calls
>    node_{connect, detach}_network also return 202.

This refers to the status of the request being made; in the case of
node_{connect, detach}_network, the request is to do the attachment.
Here, it is to report the status.

hil/api.py
naved001: yeah, but I can't place it anywhere otherwise pycodestyl(pep8) complains that it is over-indented.

hil/api.py
naved001: Don't you think it should be on the users to ensure that they check the status of their previous operations before queuing another one? specially because with this API they'll be able to check it. 

In the current version, we say that network operations are guaranteed to happen in-order, but what if there's an error in the switch that doesn't crash the network server, in that case the users will be oblivious that something bad happened. For example, the brocade driver just logs in case of failures.

> One approach would be to have the networking daemon retry a couple times itself.

In case it doesn't go through even on multiple retries, what would our strategy be? Stop the network server so we fail-safe?

hil/migrations/versions/76529f0f9e50_add_uuid_and_status_to_networkingaction.py
naved001: So I did that, and it still doesn't work.
`DETAIL:  Failing row contains (1, null, null, modify_port, 1, 1, vlan/100).`
Something is inserting null for those 2 new columns, although the script is inserting some values (tested this locally).

Another thing, after upgrading I called `hil-admin db downgrade heads`, to see if the downgrade works, but it  actually ran the script for bigInt migration (I think). I am guessing, some changes in downgrade_revision or something somewhere?

If you any other ideas that would be helpful, otherwise ill spend more time digging into the migration stuff.

hil/migrations/versions/76529f0f9e50_add_uuid_and_status_to_networkingaction.py
naved001: alright, so I am just deleting those now, no point in keeping those if we can't query them.

hil/api.py
naved001: Let's say we stop queuing any actions on a nic if the last one failed. We would need some way to force it through (when we know the problem has been fixed). If we expect this to fail rarely, then I guess we could let an admin force it through (by checking for admin privileges).

naved001:
@SahilTikale @mosayyebzadeh ping

naved001:
Hey @Izhmash, is there a reason why you made direct db calls in this [method](https://github.com/CCI-MOC/hil/blob/master/tests/unit/migrations.py#L72) rather than using the API (like how the `create_pending_aactions` does) to make the objects?

naved001:
@zenhack, it's passing the tests now with my most recent commit.

Izhmash:
@naved001 some notes about editing the migration scripts now that #931 has merged:
To update your migration script, delete/move the old script, rebase, initialize the DB, and then generate the new migration script.  Note that @zenhack had to set `branch_labels = None` in #931; you'll have to do the same in his migration script now that it is the latest one.

naved001:
@Izhmash thanks! just did that.

naved001:
@zenhack @Izhmash @SahilTikale ping

tests/unit/api/auth.py
```diff
@@ -761,3 +762,46 @@ def test_wrong_project(self):
...
+                                            'stock_int_pub')
+        self.status_id = json.loads(response[0])
+        self.status_id = self.status_id['status_id']
+
+    def test_show_networking_action_success(self):
+        """Test that project that has access to the node can run
+        show_networking_action"""
+
+        self.auth_backend.set_project(self.manhattan)
+        api.show_networking_action(self.status_id)
```
zenhack: Would make sense to actually check the output.

tests/unit/api/main.py
```diff
@@ -2376,3 +2384,78 @@ def test_node_power_cycle_force(self):
...
+        # make a network call and get the status id from the response.
+        response = api.node_connect_network('node-99', '99-eth0', 'hammernet')
+        response = json.loads(response[0])
+        status_id = response['status_id']
+
+        response = json.loads(api.show_networking_action(status_id))
+        assert response == {'status': 'PENDING', 'node': 'node-99',
+                            'nic': '99-eth0', 'type': 'modify_port',
+                            'channel': 'vlan/native',
+                            'new_network': 'hammernet'}
```
zenhack: Would be easier to read with one key/value pair per line.

tests/unit/client.py
```diff
@@ -766,3 +771,31 @@ def test_extension_list(self):
...
+                )
+        status_id = response['status_id']
+
+        response = C.node.show_networking_action(status_id)
+        assert response['status'] == 'PENDING'
+        assert response['node'] == 'node-01'
+        assert response['nic'] == 'eth0'
+        assert response['new_network'] == 'net-01'
+        assert response['channel'] == 'vlan/native'
+        assert response['type'] == 'modify_port'
```
zenhack: Any reason not to just have one `assert response == {...}`?

tests/unit/deferred.py
```diff
@@ -224,40 +227,55 @@ def test_apply_networking(switch, network, fresh_database):
...
+                                              status='PENDING'))
 
     # this makes the last action invalid for the test switch because the switch
     # raises an error when the networking action is of type revert port.
+    unique_id = str(uuid.uuid4())
     actions[2] = model.NetworkingAction(nic=nic[2],
-                                        new_network=None,
-                                        channel='',
+                                        new_network=None, uuid=unique_id,
+                                        channel='', status='PENDING',
```
zenhack: Probably better to keep it to one argument per line.

tests/unit/client.py
naved001: oh yeah, i just forgot to do it here.

docs/rest_api.md
```diff
@@ -1310,3 +1330,40 @@ Remove a user from a project.
...
+    "new_network": <network-name>
+    "type": <type of networking action>
+    "channel": <network channel>
+}
+
+where:
+* `status` can either be "DONE", "PENDING", or "ERROR".
+* `new_network` can be `null` in case of `node_detach_network` or `revert_port`
+* `type` can be `revert_port` or `modify_port`
+* `channel` could be '' in case of revert_port
```
xuhang57: missing trailing period in L1355 L1356, L1357

hil/api.py
```diff
@@ -1098,14 +1104,46 @@ def port_revert(switch, port):
 
     if port.nic is None:
         raise errors.NotFoundError(port.label + " not attached")
-    if port.nic.current_action:
-        raise errors.BlockedError("Port already has a pending action.")
+    check_pending_action(port.nic)
+    unique_id = str(uuid.uuid4())
```
xuhang57: just curious, why not just `uuid`?  That's not a default keyword right? I am perfectly fine with `unique_id`.

naved001:
```
>>> import uuid
>>> uuid = uuid.uuid4()
>>> print uuid
49185699-3189-4865-b65f-b2529dba3580
>>> another_uuid = uuid.uuid4()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'UUID' object has no attribute 'uuid4'
```
Just so that it doens't pollute the namespace.

naved001:
@zenhack @Izhmash @xuhang57 I think addressed pretty much everything you guys pointed out.

One thing to note: since I am now doing `logger.error` in `deferred.py`, I had to remove the `fail_on_log_warnings` fixture from the corresponding unit test, because in that test I am deliberately raising a switch error and catching it.

hil/api.py
xuhang57: g2k 👍

naved001:
I wanted to append some more tests to `tests/unit/deferred.py`. Basically, I wanted to test adding a valid networking action on a nic with a failed networking action.  In my test I use `local_db.session.add(new_action)`, before closing and committing, to add a new action. But it fails with a `DetachedInstanceError`. 

I read the docs for sqlachemy, but couldn't figure out what's wrong with my approach.

tests/unit/api/main.py
```diff
@@ -2376,3 +2384,80 @@ def test_node_power_cycle_force(self):
...
+        """Sets up common stuff"""
+        switchinit()
+        new_node('node-99')
+        api.node_register_nic('node-99', '99-eth0', 'DE:AD:BE:EF:20:14')
+        api.project_create('anvil-nextgen')
+        api.project_connect_node('anvil-nextgen', 'node-99')
+        network_create_simple('hammernet', 'anvil-nextgen')
+        api.port_connect_nic('sw0', PORTS[2], 'node-99', '99-eth0')
+
+    def test_show_networking_action_attach(self,):
```
zenhack: What's with the comma?

tests/unit/api/main.py
```diff
@@ -2376,3 +2384,80 @@ def test_node_power_cycle_force(self):
         new_node('node-99')
         api.project_connect_node('anvil-nextgen', 'node-99')
         api.node_power_cycle('node-99', True)
+
+
+class TestShowNetworkingAction:
+    """Various tests for the show networking action api"""
+
+    def common(self):
```
zenhack: Better to inherit from unittest.TestCase and just call this setUp (and then not explicitly call it below).

tests/unit/client.py
```diff
@@ -766,3 +771,31 @@ def test_extension_list(self):
...
+                            'type': 'modify_port',
+                            'channel': 'vlan/native',
+                            'new_network': 'net-01'}
+
+        deferred.apply_networking()
+        response = C.node.show_networking_action(status_id)
+        assert response['status'] == 'DONE'
+
+    def test_get_status_fail(self):
+        """(unsuccessful) call to get_status"""
```
zenhack: s/get_status/show_networking_action/. There are a couple of other instances of this too.

tests/unit/migrations.py
```diff
@@ -151,7 +151,10 @@ def create_bigint_db():
         db.session.add(model.NetworkingAction(type='modify_port',
                                               nic=nic1,
                                               new_network=network1,
-                                              channel='vlan/100'))
+                                              channel='vlan/100',
+                                              status='PENDING',
+                                              uuid='60bcb0f3-2ee0-4efa-b6ce-e8'
+                                              '1596c6386e'))
```
zenhack: I kinda don't like this being split across two lines. I think it will let you do:

```python
db.session.add(model.NetworkingAction(
    type='modify_port',
    ...
    uuid=...))
```

naved001:
@zenhack I'll fix the stuff you pointed out. Do you happen to know about the DetachedInstanceError that I'm seeing? To put it simply, I am trying to use `local_db` in the main test to add a new action, but it gives me that error.

zenhack:
@naved001, do you have the code you wrote somewhere for me to look at?

naved001:
Sorry for the delay in response, I have the code somewhere on my work machine. I'll push it tomorrow so you can have a look.

naved001:
@zenhack what I was trying to do earlier was flawed for other reasons.

I have updated `tests/unit/deferred.py` where I use the API to add a networking action on a nic with failed action. Running into some other problems, because I think there were some issues with the original logic of the test. Will push it soon.

naved001:
@zenhack okay done. have a look at the latest commit and the message.

I don't know how 2 actions on the same nic got added (one with status = ERROR and one with status = PENDING); because we are enforcing 1 to 1 relation.

I think tests/unit/deferred.py could be refactored, but I'd prefer to not do it in this PR because it's pretty big as it is.

@xuhang57 @Izhmash i have added some more commits after your approval. Please have a look at it once.

hil/api.py
```diff
@@ -1084,14 +1090,46 @@ def port_revert(switch, port):
...
+
+
+@rest_call('GET', '/networking_action/<status_id>', Schema({
+    'status_id': basestring}))
+def show_networking_action(status_id):
+    """Returns the status of the networking action by finding the status_id
+    in the networking actions table.
+    """
+    action = model.NetworkingAction.query.filter_by(uuid=status_id).first()
+    if action is None:
```
xuhang57: `if not action`? 

Just curious, what would be the best approach for handling vars those have value of `None`

naved001: I think that would work too, I just thought `if action is none` reads like a real sentence.

xuhang57: I just happen to think of that, I have no idea which approach is more pythonic. Personally, I use both approaches

zenhack: The existing precedent in the codebase is to do `if ... is None`. We tend to shy away from using things that aren't stictly booleans as conditions. That's certainly not universal in the Python world though.

zenhack:
LGTM. If others are happy with the state of the new tests, we can merge.

xuhang57:
lgtm

naved001:
alright, merging.

naved001:
yay! finally no. 2 on the hil leaderboards.

PR #930, opened 2018-01-07
Move the note about maintenance service to the right section (under project_detach_node)

naved001:
* Documentation fix.

The note was under `node_detach_network`, It should be under `project_detach_node`.


naved001:
Merging because docs need only 1 approval.

Issue #765, opened 2017-04-26
List public networks as a regular user


henn:
As mentioned in #453, we should have a function that lists public networks that any user can connect to in both the API and CLI.

xuhang57:
@naved001 fell free to assign me to this one.

naved001:
@xuhang57 are you still interested on working on this? otherwise @mosayyebzadeh wants to take this up!

xuhang57:
@naved001 sorry for the delay. Feel free to let @mosayyebzadeh work on this. I will try to do some work if I could.

xuhang57:
@naved001 I can start working on this today. @mosayyebzadeh Let me know if Amin hasn't started yet. ;)

PR #934, opened 2018-01-18
List public networks as a regular user

xuhang57:
## Description

Currently, only the `admin` user can list networks. This patch adds the functionality that allows a regular user to list all public networks. Any feedback and suggestion of code revise is more than welcome.

The definition of public networks:

1. Public networks with which has no project associated.
2. Networks associated with the user’s Projects.
3. Networks associated with the projects that the user is added.

This PR fixes: https://github.com/CCI-MOC/hil/issues/765
## Checklist before merging Pull Requests
- [x] New test(s) included reproducing the verify the feature
- [x] CI
- [x] Documentation
- [ ] Reviewed and approved by at least two other contributor.

### DIFF
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -246,13 +246,16 @@ Returns a JSON dictionary of dictionaries, where the exterior dictionary is inde
 the network name and the value of each key is another dictionary with keys corresponding
 to that network's id and projects
 
+Response contains all networks if having the administrative access. Otherwise, response
+only contains all public networks.
+
 The response must contain the following fields:
 
 * "network", the name of a network
 * "network_id", the id of the network
 * "projects", a list of projects with access to the network or 'None' if network is public
 
-Example Response by an admin user:
+Example Response for an admin user:
 
     {
         "netA": {
@@ -265,7 +268,7 @@ Example Response by an admin user:
         }
     }
 
-Example Response by a regular user:
+Example Response for a regular user:
 
     {
         "netA": {
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -240,7 +240,7 @@ default channel, the VLAN drivers choose `vlan/native`.
 
 `GET /networks`
 
-List all networks.
+List all networks or List all public networks.
 
 Returns a JSON dictionary of dictionaries, where the exterior dictionary is indexed by
 the network name and the value of each key is another dictionary with keys corresponding
@@ -267,7 +267,10 @@ Example Response:
 
 Authorization requirements:
 
-* Administrative access is required
+* Administrative access is required to list all networks
+* No special access is required to list all public networks
+
+
 
 #### list_network_attachments
 
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -240,7 +240,7 @@ default channel, the VLAN drivers choose `vlan/native`.
 
 `GET /networks`
 
-List all networks. (Admin-level operation)
+List all networks or list all public networks
 
 Returns a JSON dictionary of dictionaries, where the exterior dictionary is indexed by
 the network name and the value of each key is another dictionary with keys corresponding
@@ -252,7 +252,7 @@ The response must contain the following fields:
 * "network_id", the id of the network
 * "projects", a list of projects with access to the network or 'None' if network is public
 
-Example Response:
+Example Response by an admin user:
 
     {
         "netA": {
@@ -265,25 +265,7 @@ Example Response:
         }
     }
 
-Authorization requirements:
-
-* Administrative access is required to list all networks
-
-`GET /networks`
-
-List all public networks. (Regular user operation)
-
-Returns a JSON dictionary of dictionaries, where the exterior dictionary is indexed by
-the network name and the value of each key is another dictionary with keys corresponding
-to that network's id and projects
-
-The response must contain the following fields:
-
-* "network", the name of a network
-* "network_id", the id of the network
-* "projects", a list of projects with access to the network or 'None' if network is public
-
-Example Response:
+Example Response by a regular user:
 
     {
         "netA": {
@@ -298,6 +280,7 @@ Example Response:
 
 Authorization requirements:
 
+* Administrative access is required to list all networks
 * No special access is required to list all public networks
 
 #### list_network_attachments
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -240,7 +240,7 @@ default channel, the VLAN drivers choose `vlan/native`.
 
 `GET /networks`
 
-List all networks or list all public networks
+List all networks or list all public networks.
 
 Returns a JSON dictionary of dictionaries, where the exterior dictionary is indexed by
 the network name and the value of each key is another dictionary with keys corresponding
@@ -269,11 +269,11 @@ Example Response by a regular user:
 
     {
         "netA": {
-            "network_id": "101",
+            "network_id": "102",
             "projects": None
         },
         "netB": {
-            "network_id": "102",
+            "network_id": "103",
             "projects": None
         }
     }
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -240,7 +240,7 @@ default channel, the VLAN drivers choose `vlan/native`.
 
 `GET /networks`
 
-List all networks or list all public networks.
+List all networks.
 
 Returns a JSON dictionary of dictionaries, where the exterior dictionary is indexed by
 the network name and the value of each key is another dictionary with keys corresponding
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -240,7 +240,7 @@ default channel, the VLAN drivers choose `vlan/native`.
 
 `GET /networks`
 
-List all networks or List all public networks.
+List all networks. (Admin-level operation)
 
 Returns a JSON dictionary of dictionaries, where the exterior dictionary is indexed by
 the network name and the value of each key is another dictionary with keys corresponding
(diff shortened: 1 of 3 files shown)


hil/api.py
```diff
@@ -690,18 +690,35 @@ def headnode_detach_network(headnode, hnic):
...
+        for n in networks:
+            if n.access:
+                net = {'network_id': n.network_id,
+                       'projects': sorted([p.label for p in n.access])}
+            else:
+                net = {'network_id': n.network_id, 'projects': None}
+            result[n.label] = net
+    # Regular User Operation
+    else:
+        networks = db.session.query(model.Network).all()
```
naved001: You coud do this before your if else starts, so you don't have to write it twice. Same thing with `result ={} `

xuhang57: Great!

xuhang57:
I haven't finished reading all the migrations/fixtures. But I don' think we have a `regular user` created in the migrations for testing purpose. It would be nice if someone can confirm this. Thanks!

zenhack:
I think we should limit the definition of public networks to item (1) in your list. We could maybe add a separate call list_project_networks. I think that would make the implementation much simpler.

I especially don't like trying to detect which projects a user has access to inside the call. We don't do this anywhere else; making the user specify a project would be more consistent, for one.

naved001:
> I especially don't like trying to detect which projects a user has access to inside the call.

Sorry but I don't understand what's wrong with it?  I thought we didn't want HIL to have too many redundant APIs, and I figured that by doing auth checks we can list accessible networks.

xuhang57:
This is a design problem then. If we have a convention, I will surely follow it. If there is not, it is a great time for us to discuss and make one, I believe. 

Personally, I don't have a strong opinion on this. Basically, two choices:

- Minimal Design: Many functionalities per function.
- One functionality per function.

zenhack:
It's a signficant departure from how all of the other APIs work, so that alone I think is a decent argument. But also, the current implementation's code is signficantly more fiddly than I would expect from the solution I propose. (@xuhang57 has some comments re: refactoring the loops, which I do think can be improved, but the implementation is needlessly complicated by the design).

naved001:
Okay, I'll defer to you on this.
So, list_networks will list all networks if run by admins, and only public networks if run by regular users.
Then list_project_networks will list all networks owned by the project or all networks that are accessible by that project?

zenhack:
Let's do accessible networks.

xuhang57:
Great, I just pushed the change based on you guys' suggestions. `list_project_networks` has already been implemented. 

I will work with Naved on the testing for the next step.

hil/api.py
```diff
@@ -690,16 +690,21 @@ def headnode_detach_network(headnode, hnic):
...
+            if n.access:
+                net = {'network_id': n.network_id,
+                       'projects': sorted([p.label for p in n.access])}
+            else:
+                net = {'network_id': n.network_id, 'projects': None}
+            result[n.label] = net
+    # Regular User Operation
+    else:
+        pub_nets = db.session.query(model.Network).filter_by(owner=None).all()
+        if n in pub_nets:
```
naved001: you meant to do `for n in pub_nets`?

xuhang57: good catch 👍

xuhang57:
test added. Is the unit test the only test I need to worry about? (Most likely)

hil/api.py
```diff
@@ -690,18 +690,24 @@ def headnode_detach_network(headnode, hnic):
...
+                net = {'network_id': n.network_id,
+                       'projects': sorted([p.label for p in n.access])}
+            else:
+                net = {'network_id': n.network_id, 'projects': None}
+            result[n.label] = net
+    # Regular User Operation
+    else:
+        pub_nets = db.session.query(model.Network).filter_by(owner=None).all()
+        for n in pub_nets:
+            if not n.access:
```
zenhack: I think what we want to do here is actually `filter_by(access=None)`; owner is about who can delete the network, not access it, so we don't need to filter on that. And it's better to have the database check conditions like this, rather than fetching all the results and then filtering them in python.

We should also move first query in this function inside of the admin-only branch, since we don't use it in the non-admin case.

xuhang57: make sense!

xuhang57:
As for the `docs/rest_api.md`,  please let me know if there is a required format. I try to make as few changes as possible.

naved001:
@xuhang57 Just keep it consistent with how the other APIs are documented.

docs/rest_api.md
```diff
@@ -240,7 +240,7 @@ default channel, the VLAN drivers choose `vlan/native`.
 
 `GET /networks`
 
-List all networks.
+List all networks or List all public networks.
```
zenhack: Second "List" should not be capitalized.

Also, I would expand this such that it actually describes the conditions under which the call does one or the other (i.e. depending on whether the caller is admin).

hil/api.py
```diff
@@ -690,18 +690,23 @@ def headnode_detach_network(headnode, hnic):
...
+            else:
+                net = {'network_id': n.network_id, 'projects': None}
+            result[n.label] = net
+    # Regular User Operation
+    else:
+        pub_nets = db.session.query(model.Network).filter_by(access=None).all()
+        for n in pub_nets:
             net = {'network_id': n.network_id, 'projects': None}
-        result[n.label] = net
+            result[n.label] = net
```
zenhack: Can probably combine these statements: `result[n.label] = {...}`.

xuhang57:
@naved001 @zenhack done.

xuhang57:
@SahilTikale @naved001 @Izhmash @zenhack any feedback would be much appreciated

docs/rest_api.md
```diff
@@ -267,7 +267,38 @@ Example Response:
...
+        },
+        "netB": {
+            "network_id": "102",
+            "projects": None
+        }
+    }
+
+Authorization requirements:
+
+* No special access is required to list all public networks
```
naved001: This looks like it's been documented twice. I would just factor out the common part (like the URL, sample body of the request if any, description etc) and have responses for both cases. And in the end, under authorization requirements have 2 bullet points listing it. You can look at how I documented the [`show_node` API](https://github.com/CCI-MOC/hil/blob/master/docs/rest_api.md#show_node)

docs/rest_api.md
```diff
@@ -240,7 +240,7 @@ default channel, the VLAN drivers choose `vlan/native`.
 
 `GET /networks`
 
-List all networks.
+List all networks or list all public networks
```
xuhang57: missing a period here

docs/rest_api.md
```diff
@@ -265,9 +265,23 @@ Example Response:
         }
     }
 
+Example Response by a regular user:
+
+    {
+        "netA": {
+            "network_id": "101",
```
xuhang57: change the number in the `network_id` to avoid the confusion

xuhang57:
will rebase with the master

xuhang57:
are we using `squash merge`? Maybe we can change the settings and make it as the default merge option.

naved001:
we just do the regular merge.

docs/rest_api.md
```diff
@@ -240,7 +240,7 @@ default channel, the VLAN drivers choose `vlan/native`.
 
 `GET /networks`
 
-List all networks.
+List all networks or list all public networks.
```
zenhack: I still think the way this is structured is a little clumsy; the information about when you get which operation is *sortof* implicit in the authorization requirements, but it's still a little weird in that it's not actually the description of the operation itself, and it doesn't technically say how you get one operation or another, just the access you need.

I'd add a short paragraph that spells out how the decision is made.

docs/rest_api.md
```diff
@@ -252,7 +252,7 @@ The response must contain the following fields:
 * "network_id", the id of the network
 * "projects", a list of projects with access to the network or 'None' if network is public
 
-Example Response:
+Example Response by an admin user:
```
zenhack: Nitpick: s/by/for/. (The response is *by* the api server). Same thing below.

docs/rest_api.md
xuhang57: Sure, make sense, will give it a try

xuhang57:
@zenhack Done.

docs/rest_api.md
```diff
@@ -246,13 +246,16 @@ Returns a JSON dictionary of dictionaries, where the exterior dictionary is inde
 the network name and the value of each key is another dictionary with keys corresponding
 to that network's id and projects
 
+Response contains all networks if having the administrative access. Otherwise, response
```
Izhmash: I'd say something like `Response contains all networks if the user is an admin.` ...

xuhang57:
Do not merge yet since @Izhmash has a PR that enforcing a new format of writing testing scripts. Will pull his code and change this PR accordingly once his PR is merged.

docs/rest_api.md
zenhack: I think this grammar fix is important, but otherwise I'm happy.

xuhang57:
@zenhack done. please merge if you think it is ok. Thanks!

hil/api.py
```diff
@@ -682,18 +682,24 @@ def headnode_detach_network(headnode, hnic):
...
+    if get_auth_backend().have_admin():
+        networks = db.session.query(model.Network).all()
+        for n in networks:
+            if n.access:
+                projects = sorted([p.label for p in n.access])
+                result[n.label] = {'network_id': n.network_id,
+                                   'projects': projects}
+            else:
+                result[n.label] = {'network_id': n.network_id,
+                                   'projects': None}
```
zenhack: So I'm looking at this again, and we could clean this up a lot doing:

```python
if get_auth_backend.have_admin():
    networks = ...
else:
    networks = ...

# Current admin-only block.
```

xuhang57: sure, make sense to me for using the same variable names. I used a different one since originally `networks` is not within the "if" scope.

xuhang57:
@zenhack done.

zenhack:
LGTM, merging.

Issue #559, opened 2016-03-24
Implement VPN termination


gsilvis:
This solves a lot of our use-cases.  What we want is the following API calls:

- Create VPN endpoint
- Delete VPN endpoint
- Attach VPN endpoint to network
- Detach VPN endpoint from network
- Get connection details

or something similar.  ("Create VPN endpoint for network ____" is another reasonable possibility.)

With this, people can run their own headnodes anywhere---for instance, on their personal laptop, or in a pre-existing VM host, such as Openstack.

Just like the headnode service, philosophically, it's better for this to be separate.  However, having this minimal version as an extension in HaaS seems very useful and easy.  Unlike the headnode service, it is easy to make it feature-complete enough for real use.

gsilvis:
Thought:  I believe some switches may support doing VPN.  So we could have drivers that take advantage of that, so all the VPN traffic doesn't need to go through the HaaS master node

henn:
> Thought: I believe some switches may support doing VPN. So we could
> have drivers that take advantage of that, so all the VPN traffic
> doesn't need to go through the HaaS master node
>
>

Great point! We should make the API implementation-agnostic.

Also - for the initial implementation, I had been thinking that we could
use lightweight containers (like docker) to isolate VPNs from each
other, other networks as well as the host.

gsilvis:
I don't think that should be necessary---it should be possible for openvpn to route the packets directly to the relevant bridge, which will only have access to the relevant vlan

zenhack:
@dschatzberg suggested containers too; I'm surprised how quickly you both reached for that. @gsilvis, is right, it's totally unnecessary. You can actually run a separate process per vpn config; the systemd service file on my laptop for openvpn is parametrized over the config file, so you can do e.g. `systemctl start openvpn@configname`. I think the right approach (at least for the openvpn backend, if we're going to make it pluggable), it to generate a config and then start a process per-vpn. If we want to support non-systemd systems we could use [supervisord](http://supervisord.org/). All of the examples I saw when I had a (quick) look had a cert for the client authentication; I don't know what else (if anything) openvpn supports, and I'm not super-familiar with the underlying protocols, but one thing to do if we wanted to use client certs is to have the create call (which will be a put or post) expect the client cert as part of the body of the request.

dschatzberg:
[...]

Issue #548, opened 2016-03-14
Some requirements for the deployment tests are not documented


zenhack:
In particular:

* The headnode names `base-headnode` and `base-headnode-2` are hardcoded in the tests, so these *must* be present in testsuite.cfg
* site-layout.json must have at least 4 nodes, since the tests expect to use this many. The nodes don't actually have to *exist* at present, just the ports on their nics. That might change if we ever add tests for stuff like power-cycling.

//cc @SahilTikale

xuhang57:
@zenhack Is this outdated? Are we still handling the `headnode` in hil?

zenhack:
afaik this is still current. We haven't actually removed the headnode stuff yet; that has to wait until the openvpn stuff is implemented.

PR #936, opened 2018-01-19
Openvpn spec

zenhack:
This is rough, but I a first pass at a design for the openvpn stuff. Probably not done, but hopefully a good starting point for discussion.

### DIFF
--- a/specs/openvpn-support.md
+++ b/specs/openvpn-support.md
@@ -28,7 +28,7 @@ To allow the daemon to invoke this tool, an administrator will add an
 entry to `/etc/sudoers` like:
 
 ```
-hil-vpnd ALL = (root) /usr/libexec/hil-vpn-privop
+hil-vpnd ALL = (root) NOPASSWD: /usr/libexec/hil-vpn-privop
 ```
 
 where `hil-vpnd` is the user which the daemon will run as. For extra
@@ -42,7 +42,7 @@ execution to other users entirely.
 The daemon will be called by HIL, and will provide a very simple
 authentication scheme (probably the same as the AdminToken used by
 obmd). It will not need to be accessible by users, though the openvpn
-process do for obvious reasons.
+processes do for obvious reasons.
 
 ## API
 
--- a/specs/openvpn-support.md
+++ b/specs/openvpn-support.md
@@ -16,7 +16,7 @@ HIL will talk to the daemon to manage vpns
   configuration under the name `<name>`, which attaches to `<vlan-no>`
   and listens on `<port-no>`. It generates a static key and prints it
   to stdout.
-* `hil-vpn-privo start <name>` starts an openvpn process for the config
+* `hil-vpn-privop start <name>` starts an openvpn process for the config
   `<name>`.
 * `hil-vpn-privop stop <name>` stops the openvpn process for the config
   `<name>`.
--- a/specs/openvpn-support.md
+++ b/specs/openvpn-support.md
@@ -1,0 +1,107 @@
+We will implement a new microservice to manage vpns, which will consist
+of two separate programs:
+
+* `hil-vpnd`, which is a daemon exposing a rest api for manipulating
+  vpns
+* `hil-vpn-privop`, a wrapper for the priviledged operations needed by
+   the daemon.
+
+HIL will talk to the daemon to manage vpns
+
+# Privileged operation wrapper
+
+`hil-vpn-privop` is a command line tool providing a narrow interface:
+
+* `hil-vpn-privop create <name> <vlan-no> <port-no>` creates an openvpn
+  configuration under the name `<name>`, which attaches to `<vlan-no>`
+  and listens on `<port-no>`. It generates a static key and prints it
+  to stdout.
+* `hil-vpn-privo start <name>` starts an openvpn process for the config
+  `<name>`.
+* `hil-vpn-privop stop <name>` stops the openvpn process for the config
+  `<name>`.
+* `hil-vpn-privop delete <name>` deletes the config for `<name>`. The
+  vpn must be stopped for this to work.
+* `hil-vpn-privop list` lists the available configs.
+
+To allow the daemon to invoke this tool, an administrator will add an
+entry to `/etc/sudoers` like:
+
+```
+hil-vpnd ALL = (root) /usr/libexec/hil-vpn-privop
+```
+
+where `hil-vpnd` is the user which the daemon will run as. For extra
+security, we will recommend setting the permissions on the file to deny
+execution to other users entirely.
+
+# The daemon
+
+## Authentication
+
+The daemon will be called by HIL, and will provide a very simple
+authentication scheme (probably the same as the AdminToken used by
+obmd). It will not need to be accessible by users, though the openvpn
+process do for obvious reasons.
+
+## API
+
+The daemon provides a REST API for manipulating vpns attached to VLANs.
+it uses `hil-vpn-privop` to perform privileged operations, in order to
+minimize attack surface.
+
+### Create a new vpn
+
+`POST /vpns/new`
+
+Request body
+
+```
+{
+    "vlan": 3231,
+}
+```
+
+Response body:
+
+```json
+{
+    "key": "<pem-encoded key>",
+    "id": "<unique-id>",
+    "port": 9321
+}
+```
+
+Allocates a new vpn attached to the specified vlan. Returns the port
+number on which it is listening, along with a unique identifier and
+a pem encoded key.
+
+### Delete a vpn
+
+`DELETE /vpns/<id>`
+
+where `<id>` is the id from the body of the call that created vpn.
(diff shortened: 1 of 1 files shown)


zenhack:
Quoting Naved Ansari (2018-01-18 21:34:11)

>    Really basic question about how openvpn/vpn works:
>
>    A hil user will run an open vpn client on their system which will
>    connect to the open vpn "configuration" (over the internet, using the
>    ip and port) on the host with access to the vlans. Is my understanding
>    correct?

Correct. In more detail, it creates a virtual NIC on the client's
system, and just wraps the Ethernet frames sent on that nice inside of
UDP packets, plus some crypto.

zenhack:
Quoting Naved Ansari (2018-01-22 16:19:02)

> Also, do you want to mention in this spec what changes will be
> necessary in HIL?

Sure, I can add that.

>    does it create both, server side and client side, config files? if it
>    does make the client side file, where does it keep it or just print it
>    to stdout?

It only creates the server side; the client is responsible for
configuring OpenVPN themselves. HIL will have to provide the client with
necessary information (host/port number/key), and we will provide
documentation on how to configure a client.


>      * This page says we can only have 1 client and 1 server. For
>        scalibility we would want multiple clients for a server.

Things to keep in mind:

* In this context "server" means "openvpn process." I haven't
  investigated what the footprint of openvpn is, but it's still going to
  be *much* less than any VM based solution we provide, and I doubt it
  will be a huge problem from a performance perspective.
* We can attach more than one VPN to a single network, so the
  restriction doesn't really impact how the tool can be used.
* This is mainly for bootstrapping; users with sophisticated needs will
  likely want to install provisioning software on a HIL node.

>      * What version of OpenVPN do we plan to use? From the docs [3]here it
>        looks like we would require 2.0 or greater.

Requiring >=2.0 is fine with me.

zenhack:
Quoting Ian Denhardt (2018-01-22 16:48:49)

> > Also, do you want to mention in this spec what changes will be
> > necessary in HIL?
>
> Sure, I can add that.

Done.

naved001:
> * In this context "server" means "openvpn process." 

Oh now it all makes sense. You are spawning independent OpenVPN process because they will attach to different VLAN (bridge, interface) on the server. In that case we don't even need 2.0, because I mentioned that for having multiple clients per server, which now I understand that we don't need.

specs/openvpn-support.md
```diff
@@ -0,0 +1,154 @@
...
+pool of interfaces (in HIL we create this with the `create_interfaces`
+script). We can probably re-use much of the code.  We'll also need to
+specify a range of ports to use for the vpns.
+
+Creating a vpn will consist of generating a key and configuration file
+in an appropriate location, which will refer to the key and specify a
+tap device corresponding to the appropriate vlan. We can manage the
+life cycle of the openvpn processes via systemd.
+
+# Changes to HIL
```
naved001: I am assuming there will be some database changes that'll be made (to store the port number mapped to a network and project). Don't know if it's worth documenting it here.

specs/openvpn-support.md
```diff
@@ -0,0 +1,154 @@
...
+
+* network_delete will require that no vpns are attached to the network
+* HIL will grow some new api calls:
```
naved001: what about some show calls?
* list_project_VPNs
* show_vpn <vpn name> would list out the network it connects to, and the port on the host.

zenhack: Sounds good to me.

specs/openvpn-support.md
zenhack: Yeah, we'd basically have to add a new table to the database for vpns. I can document if you'd like, but I figured it was fairly obvious.

naved001: okay, I am fine with not documenting it.

naved001:
some discussion here #559 

Also, @pjd-nu emailed me saying that he'll review this over the weekend and that would be nice.

razaaliraza:
I have gone through the specs and the conversation here and have just one really basic question.

> This is mainly for bootstrapping; users with sophisticated needs will
  likely want to install provisioning software on a HIL node.

What do we mean by sophisticated users and how do their requirements change from what can be catered through this micro-service? Can we add more functionality to this micro-service so even sophisticated users can use it or would that be redundant?

zenhack:
Re: more sophisticated use cases: I can imagine wanting to have nodes distribute installation images in a p2p fasion, or wanting redundancy for your management node(s), or just wanting more horsepower than the cheapo vm you rented in the rack across the hall. For some folks, just throwing a pxe server in a vm and hooking it up via vpn may be enough, but for other use cases, you can use the vpn to provision a HIL node, and do more from there.

> Can we add more functionality to this micro-service so even sophisticated users can use it or would that be redundant?

The intention is that if users need more, they can use the vpn to install other provisioning tools inside the HIL networks. It's not that they "can't use it," but will probably use it to bootstrap their own provisioning system. HIL is very much a "mechanism, not policy" design, and we're really aiming to provide the bare minimum. So it's less of a question of "can we" than "should we."

razaaliraza:
That clarifies the need for the bare minimum. this ties into HIL's philosophy. Thanks!

naved001:
one quick question; are you planning to write the daemon in Go or python? I would prefer python since more people could give helpful reviews, but if Go has any advantages in this case, then I'd say go for it.

zenhack:
It might make synchronization a bit easier; I haven't though through in detail what back-end operations we'd need to synchronize, but Go is nice for that kind of stuff (this is why I chose it for obmd). I'd thought Python might make some sense in this case because maybe we'd be able to lift some of the headnode related code from HIL, but I had a look and there's not really much relevant logic. Re: more reviewers, point, but I think long term if we're going to be maintaining one codebase in Go, it's probably worth some other people getting some practice.

That said, there's the disclaimer that I'm probably going to be a bit biased towards Go rather than Python in general just because I *like* it better.

naved001:
okay, good enough for me.

Let's wait for a couple days more if @pjd-nu reviews this, otherwise we'll go ahead with the plan.
@razaaliraza does this have your approval? if so, could you approve it/give it a +1

pjd-nu:
Sorry for the delay.

Each OpenVPN config needs to listen on a separate <IP,port> pair; i.e. there's no way to multiplex different connections. (unlike e.g. ssh) The current design has the following properties:

* The IP address is implicit - vpnd is unable to specify it, and the return value from vpn-privop doesn't indicate it. This seems like it may be a bit of an issue, because now vpnd has to know the external address, which will no doubt break if someone changes that address.

* Port numbers are managed (allocated) by vpnd. That doesn't seem to be a problem.

* .ovpn files are created by vpnd. Again doesn't seem to be a problem. Note that the .ovpn files from the bridged ethernet config I had for a while have several certificates in them as well as the private key - someone needs to think about whether this information will be available to unprivileged vpnd - if it's either (1) protected, or (2) generated along with the config then it needs to be returned by vpn-privop.

zenhack:
Quoting Peter Desnoyers (2018-01-31 11:33:06)

>    Each OpenVPN config needs to listen on a separate <IP,port> pair; i.e.
>    there's no way to multiplex different connections. (unlike e.g. ssh)

Correct.

>    The current design has the following properties:

>      * The IP address is implicit - vpnd is unable to specify it, and the
>        return value from vpn-privop doesn't indicate it. This seems like
>        it may be a bit of an issue, because now vpnd has to know the
>        external address, which will no doubt break if someone changes that
>        address.

Good call; we probably want to return the ip address in the response
body from vpnd.

One way to handle this would be to store the IP to listen on in a
(global) config file, which then gets propagated to the generated
openvpn configs and also returned by vpnd.

>      * Port numbers are managed (allocated) by vpnd. That doesn't seem to
>        be a problem.

Yep.

>      * .ovpn files are created by vpnd. Again doesn't seem to be a
>        problem. Note that the .ovpn files from the bridged ethernet config
>        I had for a while have several certificates in them as well as the
>        private key - someone needs to think about whether this information
>        will be available to unprivileged vpnd - if it's either (1)
>        protected, or (2) generated along with the config then it needs to
>        be returned by vpn-privop.

The notion was to use the static key setup. This was linked from the
"Openvpn details" section of the document:

    https://openvpn.net/index.php/open-source/documentation/miscellaneous/78-static-key-mini-howto.html

This means not having to deal with PKI or even asymmetric crypto, just the one
shared secret.

The top of the proposal mentions that the openvpn config and key are
generated by vpn-privop. The key is printed to stdout, making it
available to vpnd.

xuhang57:
looping this issue: https://github.com/CCI-MOC/hil/issues/548

naved001:
@pjd-nu does this have your approval or is there some change that you would like to see? Thanks!

naved001:
@shwsun @SahilTikale @knikolla @pjd-nu @Izhmash  anyone wants to review and approve this?

naved001:
@radonm

Issue #924, opened 2017-12-15
Unimplemented api calls in client library


Izhmash:
Neither Node's nor Switch's `register()` functions are implemented in the client library.   Since every call should be accessible in the client library, this should be fixed.

zenhack:
Yeah, I think there are a number of calls that aren't there. They're not hard to add. It would be good to make a list.

naved001:
1. `node_register()`
2. `switch_register()`
3. `node_set_bootdev()`
69. `metadata_set()`
420. `metadata_delete()`
69. `list_network_attachments()`

`start_console()` and `stop_console()` use the client library, but `show_console()` doesn't. That's weird.
The headnode stuff doesn't use any client library calls, but we don't need to add support for it either.

cc: @SahilTikale

zenhack:
The console stuff probably isn't worth bothering with until the obmd stuff lands (or maybe we should do that at the same time).

I agree, we can omit the headnode stuff. Though it would be trivial to add if someone were going to just slog through and add all of these in one pass.

I edited your comment to turn it into a checklist; hope that's cool.

xuhang57:
Will look into this. Trying to figure out how to use the CLI for these operations first.

zenhack:
Cool, I'l assign you. I've also changed the title of the issue to reflect the current thinking.

naved001:
Complete!

PR #939, opened 2018-01-23
API for node_register

xuhang57:
This patch adds the API function for `node_register`. The `cli` code change accordingly.

Currently, `obm_types = ["ipmi", "mock"]` is a hardcoded variable. A discussion is needed on whether or not to keep the hardcoded variable.

Any idea/suggestion/opinion/ is more than welcome.

Consider this is a WIP.

This relates to the issue: https://github.com/CCI-MOC/hil/issues/924

### DIFF
--- a/hil/client/node.py
+++ b/hil/client/node.py
@@ -47,8 +47,7 @@ class Node(ClientBase):
                        "user": args[1], "password": args[2]
                        }
         else:
-            raise UnknownSubtypeError("Unknown subtype provided.
-                                      Please check your local settings")
+            raise UnknownSubtypeError("Unknown subtype provided.")
 
         url = self.object_url('node', node)
         payload = json.dumps({"obm": obminfo})
--- a/hil/client/node.py
+++ b/hil/client/node.py
@@ -32,7 +32,27 @@ class Node(ClientBase):
         # and currently active drivers for HIL
         # obm_api = "http://schema.massopencloud.org/haas/v0/obm/"
         # obm_types = ["ipmi", "mock"]
-        raise NotImplementedError
+        obm_api = "http://schema.massopencloud.org/haas/v0/obm/"
+        obm_types = ["ipmi", "mock"]
+        if subtype in obm_types:
+            if len(args) == 3:
+                obminfo = {"type": obm_api + subtype, "host": args[0],
+                           "user": args[1], "password": args[2]
+                           }
+            else:
+                sys.stderr.write('ERROR: subtype ' + subtype +
+                                 ' requires exactly 3 arguments\n')
+                sys.stderr.write('<hostname> <ipmi-username> <ipmi-password>\n')
+                return
+        else:
+            sys.stderr.write('ERROR: Wrong OBM subtype supplied\n')
+            sys.stderr.write('Supported OBM sub-types: ipmi, mock\n')
+            return
+        url = self.object_url('node', node)
+        payload = json.dumps({"obm": obminfo})
+	return self.check_response(
+                self.httpClient.request('PUT', url, data=payload)
+                )
 
     def delete(self, node_name):
         """Deletes the node from database. """
--- a/hil/client/node.py
+++ b/hil/client/node.py
@@ -43,6 +43,10 @@ class Node(ClientBase):
                 obminfo = {"type": obm_api + subtype, "host": args[0],
                            "user": args[1], "password": args[2]
                            }
+        else:
+            raise Exception('ERROR: subtype ' + subtype +
+                            ' requires exactly 3 arguments\n')
+
         url = self.object_url('node', node)
         payload = json.dumps({"obm": obminfo})
         return self.check_response(
--- a/hil/client/node.py
+++ b/hil/client/node.py
@@ -1,10 +1,8 @@
 """Client support for node related api calls."""
 import json
 from hil.client.base import ClientBase
-from hil.errors import BadArgumentError
-from hil.errors import UnknownSubtypeError
 from hil.client.base import check_reserved_chars
-
+from hil.errors import BadArgumentError, UnknownSubtypeError
 
 class Node(ClientBase):
     """Consists of calls to query and manipulate node related
--- a/hil/client/node.py
+++ b/hil/client/node.py
@@ -1,6 +1,8 @@
 """Client support for node related api calls."""
 import json
 from hil.client.base import ClientBase
+from hil.errors import BadArgumentError
+from hil.errors import InactiveSubtypeError
 
 
 class Node(ClientBase):
@@ -32,18 +34,19 @@ class Node(ClientBase):
         # and currently active drivers for HIL
         # obm_api = "http://schema.massopencloud.org/haas/v0/obm/"
         # obm_types = ["ipmi", "mock"]
+        if (len(args) != 3):
+            raise BadArgumentError
+
         obm_api = "http://schema.massopencloud.org/haas/v0/obm/"
         # This is a temp fix. obmd will let node_register no longer
         # need a type so a new design will be required.
         obm_types = ["ipmi", "mock"]
         if subtype in obm_types:
-            try:
-                obminfo = {"type": obm_api + subtype, "host": args[0],
-                           "user": args[1], "password": args[2]
-                           }
-            except Exception:
-                raise Exception('ERROR: subtype ' + subtype +
-                                ' requires exactly 3 arguments')
+            obminfo = {"type": obm_api + subtype, "host": args[0],
+                       "user": args[1], "password": args[2]
+                       }
+        else:
+            raise InactiveSubtypeError
 
         url = self.object_url('node', node)
         payload = json.dumps({"obm": obminfo})
--- a/hil/client/node.py
+++ b/hil/client/node.py
@@ -34,12 +34,10 @@ class Node(ClientBase):
         # obm_types = ["ipmi", "mock"]
         obm_api = "http://schema.massopencloud.org/haas/v0/obm/"
         ext_url = self.object_url('active_extensions')
-        ext_resp = self.check_response(self.httpClient.request("GET", ext_url))
-        obm_types = []
-        for resp in ext_resp:
-            if "hil.ext.obm" in resp:
-                obm_types.append(resp.split(".")[-1])
-
+        # This is a temp fix. obmd will let node_register no longer
+        # need a type so a new design will be required.
+        obm_types = ["ipmi", "mock"]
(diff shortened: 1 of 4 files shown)


xuhang57:
one of the ideas to improve the code is to store `obm_api` as a global variable somewhere so we can use `self.obm_api` to get the uri (url?). I am still new to the code base. I guess we can probably put it in the base.py under `hil/client/`? Any idea would be much appreciated.

xuhang57:
@SahilTikale @naved001 @Izhmash @zenhack  any suggestion/discussion would be much appreciated. I don't know whether I am on the right track and is there any improvements I could make? Thanks!

hil/client/node.py
```diff
@@ -32,7 +32,24 @@ def register(self, node, subtype, *args):
...
-        raise NotImplementedError
+        obm_api = "http://schema.massopencloud.org/haas/v0/obm/"
+        ext_url = self.object_url('active_extensions')
+        ext_resp = self.check_response(self.httpClient.request("GET", ext_url))
+        obm_types = []
+        for resp in ext_resp:
+            if "hil.ext.obm" in resp:
+                obm_types.append(resp.split(".")[-1])
+
+        if subtype in obm_types:
```
naved001: you haven't handled the `else` case here. What if it's not in the subtype, we would still make the API call with nothing in the payload.

xuhang57: Currently, I got `Error: Request arguments are not valid for this request`.  So should I add some error handling here?

naved001: I think we could raise an exception when we detect that `subtype is not in obm_types`. Client library raises the `FailedAPICallException` with some helpful message.

zenhack:
Worth noting, once the obmd stuff lands, node_register won't take a type
at all, because the differences between OBMs will be handled by obmd. So
I wouldn't fuss about it right now.

This of course pushes the question about how to deal with this to
whatever software we end up using to talk to obmd, but that doesn't go
in HIL's node_register.

xuhang57:
So the `obmd` would only affect the `node_register` right? As for other items in the list such as switch_register, we should be able to proceed?

As for the `node_register`, do we need a temporary solution for the client api? If not, what would be the timeline for having the `obmd` in place? Otherwise, hardcode the type and leave a note along with the change would be a good solution.

Thanks for the inputs guys.

naved001:
@zenhack 
> Worth noting, once the obmd stuff lands, node_register won't take a type at all, because the differences between OBMs will be handled by obmd. So I wouldn't fuss about it right now. 

Good point. Do you think it's still worth moving things into the client library for now since the work has already been done? Like how @xuhang57 suggested that we hardcode the type and leave a note that that call will be changed when obmd is in.

> This of course pushes the question about how to deal with this to whatever software we end up using to talk to obmd, but that doesn't go in HIL's node_register.

I guess, we could write a basic CLI to talk to obmd and handle these cases for obmd's register call in the cli itself. Since that will be admin only operations and not really user facing.

@xuhang57 
> So the obmd would only affect the node_register right? As for other items in the list such as switch_register, we should be able to proceed?

Yes and yes. And I think my point about handling subtypes and payload still stands.

hil/client/node.py
zenhack: `FailedAPICallException` is a bit weird here since it's not in response to an error from the server. It might make more sense to raise a different error, though I don't feel super strongly.

zenhack:
Right, the obmd stuff only affects node_register. I'm fine just hard-coding stuff in the meantime.

xuhang57:
@zenhack @naved001 Currently, I am querying the obm type from user's settings rather than hardcoding them. How's that? If fine, code review would be much appreciated!

hil/client/node.py
naved001: oh right, in that case you could define a new error in base.py and raise that.

naved001:
@zenhack @xuhang57 I am not still not sure why we need the list_active_extensions API. Wish I had spoken earlier about it when we merged it, but I am having doubts now.

xuhang57:
ok, before I make some changes on the `raise exception`, I am inclined to hard code the obm types after discussing with Naved. @SahilTikale has asked about this and the reason why we are coding is since the api server will reject the invalid type s anyways, there is no need to check the active extension in the client library again.

Will update the patch accordingly.

naved001:
Just to add what @xuhang57 said:

My idea is that, even if we get the active OBM types (or in case of switch_register, the active switch types), we still need to know, for each active backend, how to make the request body.

zenhack:
Yeah, and furthermore it seems unlikely that someone using `*_register` would not know a priori how HIL was configured -- these calls are admin-only after all. and if they do need to be able to get the info dynamically, they can just respond to the failures. I kindof agree that `list_active_extensions` seems unnecessary. I think I vaguely remember feeling that way at the time, but didn't have the energy to argue with folks. Do we know of any code actually using it? I'm fine removing it entirely (if we want to be cautious we could mark it as deprecated for a while).

xuhang57:
I don't see anywhere in the code base has used that function except for the tests. And yeah, should mark it as deprecated for 2 releases?

naved001:
My vote is to remove that call entirely. @SahilTikale I remember you opening the issue for having that API, do you have any strong argument for keeping it?

xuhang57:
Will ping Sahil during the weekly meeting.

hil/client/node.py
```diff
@@ -32,7 +32,26 @@ def register(self, node, subtype, *args):
...
+        # need a type so a new design will be required.
+        obm_types = ["ipmi", "mock"]
+        
+        if subtype in obm_types:
+            if len(args) == 3:
+                obminfo = {"type": obm_api + subtype, "host": args[0],
+                           "user": args[1], "password": args[2]
+                           }
+        else:
+            raise Exception('ERROR: subtype ' + subtype +
```
xuhang57: I am not sure whether it is worth to write a new exception in the base.py for this function. 
1. After having the `obmd`, are we still checking the length of args?
2. Would this function be the only function that uses the new exception? 

@zenhack @naved001 thoughts?

naved001: Switch register could also use this. And when we have some other network backend then the call to create networks might require different number of arguments (not sure on this though).

xuhang57:
@zenhack @naved001  Done.

hil/client/node.py
zenhack: Probably best not to just raise `Exception`; this makes it hard to just handle the relevant exception. I would make another class.

hil/client/node.py
```diff
@@ -32,7 +32,24 @@ def register(self, node, subtype, *args):
         # and currently active drivers for HIL
         # obm_api = "http://schema.massopencloud.org/haas/v0/obm/"
         # obm_types = ["ipmi", "mock"]
-        raise NotImplementedError
+        obm_api = "http://schema.massopencloud.org/haas/v0/obm/"
+        # This is a temp fix. obmd will let node_register no longer
+        # need a type so a new design will be required.
+        obm_types = ["ipmi", "mock"]
+        if subtype in obm_types:
+            try:
```
zenhack: It would be clearer to just check the length up front:

```python
if len(args) != 3:
    raise ...
```

hil/client/node.py
```diff
@@ -32,7 +32,24 @@ def register(self, node, subtype, *args):
...
+                obminfo = {"type": obm_api + subtype, "host": args[0],
+                           "user": args[1], "password": args[2]
+                           }
+            except Exception:
+                raise Exception('ERROR: subtype ' + subtype +
+                                ' requires exactly 3 arguments')
+
+        url = self.object_url('node', node)
+        payload = json.dumps({"obm": obminfo})
```
zenhack: Right now if the user passes a bad subtype, we'll get to this line and raise a `NameError`; we should probably signal an error up front in this case, rather than falling through to here.

xuhang57: sure, basically an exception handling/validation for subtype before it passes to the api

xuhang57:
@naved001 @zenhack done.

hil/client/node.py
```diff
@@ -32,7 +34,25 @@ def register(self, node, subtype, *args):
         # and currently active drivers for HIL
         # obm_api = "http://schema.massopencloud.org/haas/v0/obm/"
         # obm_types = ["ipmi", "mock"]
-        raise NotImplementedError
+        if (len(args) != 3):
+            raise BadArgumentError
```
naved001: I think you should tell why it's bad. ("3 Arguments needed. Supplied = len(args)")

tests/unit/client.py
```diff
@@ -295,6 +295,11 @@ def test_list_nodes_all(self):
                 u'node-06', u'node-07', u'node-08', u'node-09'
                 ]
 
+    def test_node_register(self):
+        """(successful) to node_register"""
```
naved001: I know you copied it from other tests, but I think it reads better as "successful *call* to node_register"

Might also wanna add 2 negative tests, one with invalid subtype, and another with incorrect number of arguments.

hil/client/node.py
xuhang57: sure!

tests/unit/client.py
```diff
@@ -296,9 +297,16 @@ def test_list_nodes_all(self):
                 ]
 
     def test_node_register(self):
-        """(successful) to node_register"""
+        """(successful) to call node_register"""
```
naved001: (successful) **call to** node_register

not the other way around. But since you are testing unsuccessful calls in this test as well, you could just update the docstring to "Test node_register"  or "Test various calls to node_register", whatever you like.

xuhang57: cannot think :( . Thanks!

hil/client/node.py
```diff
@@ -32,18 +34,19 @@ def register(self, node, subtype, *args):
...
-                           "user": args[1], "password": args[2]
-                           }
-            except Exception:
-                raise Exception('ERROR: subtype ' + subtype +
-                                ' requires exactly 3 arguments')
+            obminfo = {"type": obm_api + subtype, "host": args[0],
+                       "user": args[1], "password": args[2]
+                       }
+        else:
+            raise InactiveSubtypeError
```
zenhack: Should provide a message to go along with this, i.e. `InactiveSubTypeError('...')`. This is true for errors in general.

also, Maybe `Unknown` instead of `Inactive`? that makes more sense to me.

xuhang57:
@zenhack @naved001 done.

xuhang57:
@naved001 please merge it if you think it is good to go ;)

Issue #795, opened 2017-06-09
create() in haas/client/user.py accepts a string rather than bool for admin privilege


Izhmash:
This should be cleaned up to keep the client library as clean and streamlined as possible.

Suggestion: `create()` should take a boolean rather than a string for `is_admin`.  The CLI rather than the client library should then handle the strings ('admin', 'regular').  The approach would be similar to `set_admin()`.

xuhang57:
PR merged so close this now

PR #941, opened 2018-01-26
create() in haas/client/user.py accepts a string rather than bool for admin privilege

xuhang57:
This patch fixes this: https://github.com/CCI-MOC/hil/issues/795

### DIFF
--- a/hil/client/user.py
+++ b/hil/client/user.py
@@ -13,21 +13,17 @@ class User(ClientBase):
     manipulate users related objects and relations.
     """
 
-    def create(self, username, password, privilege):
+    def create(self, username, password, is_admin):
         """Create a user <username> with password <password>.
 
-        <privilege> may by either "admin" or "regular",
+        <is_admin> may by either true or false,
         and determines whether a user is authorized for
         administrative privileges.
         """
         url = self.object_url('/auth/basic/user', username)
 
-        if privilege not in('admin', 'regular'):
-            raise ValueError(
-                "invalid privilege type: must be either  'admin' or 'regular'."
-                )
         payload = json.dumps({
-                'password': password, 'is_admin': privilege == 'admin',
+                'password': password, 'is_admin': is_admin,
                 })
         return self.check_response(
                 self.httpClient.request("PUT", url, data=payload)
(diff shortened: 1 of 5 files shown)


hil/client/user.py
```diff
@@ -13,21 +13,17 @@ class User(ClientBase):
     manipulate users related objects and relations.
     """
 
-    def create(self, username, password, privilege):
+    def create(self, username, password, is_admin):
         """Create a user <username> with password <password>.
 
-        <privilege> may by either "admin" or "regular",
+        <is_admin> may by either true or false,
```
naved001: might just say `<is_admin> is a boolean`

hil/ext/auth/database.py
```diff
@@ -63,9 +63,9 @@ def set_password(self, password):
 @rest_call('PUT', '/auth/basic/user/<user>', schema=Schema({
     'user': basestring,
     'password': basestring,
-    Optional('is_admin'): bool,
+    'is_admin': bool,
```
naved001: this changes the API. somebody could still directly make the API call where we want this to be optional.

hil/client/user.py
xuhang57: 👍

hil/ext/auth/database.py
xuhang57: sure, will change accordingly.

tests/unit/client.py
```diff
@@ -602,75 +602,75 @@ class Test_user:
 
     def test_user_create(self):
         """ Test user creation. """
-        assert C.user.create('billy', 'pass1234', 'admin') is None
-        assert C.user.create('bobby', 'pass1234', 'regular') is None
+        assert C.user.create('billy', 'pass1234', True) is None
+        assert C.user.create('bobby', 'pass1234', False) is None
```
zenhack: It would probably be a bit more readable to pass this argument by name, e.g `is_admin=True`, now that the value is less self-describing.

xuhang57: good point, will add that in the next commit.

xuhang57:
@zenhack @naved001 done.

zenhack:
LGTM, merged.

PR #942, opened 2018-01-26
Add a subcommand on the hil tool that report the version itself

xuhang57:
This patch relates to https://github.com/CCI-MOC/hil/issues/389
And this implementation is based on the first option on @zenhack 's list.

### DIFF
--- a/hil/cli.py
+++ b/hil/cli.py
@@ -234,8 +234,7 @@ def do_delete(url):
 @cmd
 def version():
     """Check hil version"""
-    sys.stdout.write(VERSION)
-
+    sys.stdout.write("HIL version: %s\n" % VERSION)
 
 @cmd
 def serve(port):
--- a/hil/cli.py
+++ b/hil/cli.py
@@ -236,6 +236,7 @@ def version():
     """Check hil version"""
     sys.stdout.write(VERSION)
 
+
 @cmd
 def serve(port):
     """Run a development api server. Don't use this in production."""
--- a/hil/cli.py
+++ b/hil/cli.py
@@ -26,6 +26,8 @@ import urllib
 import schema
 import logging
 
+import pkg_resources
+
 from functools import wraps
 
 from hil.client.client import Client, RequestsHTTPClient, KeystoneHTTPClient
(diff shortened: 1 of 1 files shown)


xuhang57:
This would return: 
```
0.3.dev163+ga715d2e(.venv) ubuntu@lucas-hil:~/hil$ vim hil/cli.py
```

Any suggestion on what exactly the return value should look like? 

And on github tag:https://github.com/CCI-MOC/hil/releases
There is only a `0.2` release.

hil/cli.py
```diff
@@ -37,7 +39,7 @@
 usage_dict = {}
 MIN_PORT_NUMBER = 1
 MAX_PORT_NUMBER = 2**16 - 1
-
+VERSION = pkg_resources.require('hil')[0].version
```
naved001: any reason why you are getting version here and not in the function itself?

xuhang57: I use Ian's code as an example:https://github.com/zenhack/simp_le/blob/0fc118de6c0fcad0e7a23a8e788a1e00995a1f5a/simp_le.py#L67

And I have seen people coding this way in other places. I guess it is a convention? But I understand we only use this once in the entire code base. Maybe using just the function makes sense?

naved001: Don't feel too strongly about this. If it's a convention which might enable some reuse in the future, I am okay with it.

xuhang57: hmm, @zenhack might know better here. I am not sure whether this is kinda a convention but it seems like so to me :)

zenhack: I don't have a strong opinion either. The code I linked to is in a project I inherited; I wasn't around for the decision on that one.

Re: output format, maybe just something like:

```
HIL version ${version_number}
```

You sound make sure to add a newline, so it doesn't end up on the same line as your prompt.

xuhang57: Sure, was planning to add that once I have an idea about how the output should look like. Should've mentioned that earlier. Thanks.

xuhang57:
```
(.venv) ubuntu@lucas-hil:~/hil$ hil version
HIL version: 0.3.dev163+ga715d2e
```
Updated cli response formt. @zenhack @naved001

naved001:
Any need to rebase since I think we modified cli.py in another PR which we merged (though in different places)?

naved001:
There's a rebase and merge button too? Anybody used it before?

zenhack:
Would prefer if you merged master back into your branch (don't care rebase vs. merge commit). I'd prefer not to do it on GitHub because that means we're pushing something to master that is not the exact same tree that travis is running against.

zenhack:
...but I don't know that it's worth fussing about for something so small.

naved001:
Alright, I'll merge.

Issue #903, opened 2017-10-25
Document our client library


naved001:
Right now there are no documentation for the HIL client library and we should add that soon.

@tpd001 suggested that we could have:

> 
> 1. A summary README.md at least specifying the main interfaces and how to use them, or an update to the readthedocs page which covers the client library.  Some examples would be helpful.   Any debug controls should be described.
> 
> 2.  A requirements.txt file

cc: @SahilTikale

naved001:
@razaaliraza I am assigning this to you. This seems like a good way to start learning more about the code and I'll be available to help/review.

xuhang57:
I could look into this. A discussion with @naved001 would be handy.

xuhang57:
@zenhack @naved001 so an outline for the `README.md` here, do we need to cover anything else?:

### HIL Python Client API
#### How to get started?
```
import requests
import os
from hil.client.client import Client, RequestsHTTPClient

ep = os.environ.get('HIL_ENDPOINT')
basic_username = os.getenv('HIL_USERNAME')
basic_password = os.getenv('HIL_PASSWORD')

http_client = RequestsHTTPClient()
http_client.auth = (basic_username, basic_password)
C = Client(ep, http_client)
print C.project.create("test-project")

```

naved001:
* I don't think you don't need to import requests, the client library does that for you.

What you have provided otherwise is a decent sample to show how to use the client object. You could also link the [leasing script](https://github.com/CCI-MOC/hil/blob/master/examples/leasing/node_release_script.py) as a more comprehensive example.

* And to be able to import the hil client modules, I just do a `pip install git+https://github.com/cci-moc/hil` which puts those modules in the path. Advanced users could pick the files themselves. Would be nice to mention this. I don't think there's need to document every other call.

xuhang57:
PR merged. W now have the doc on our client library. We need to add more examples in the doc though.

Issue #677, opened 2016-11-02
Current doc need some cleanup


shwsun:
I have spotted some markdown syntax error in our docs. For example, in [INSTALL.rst](https://github.com/CCI-MOC/hil/blob/master/docs/INSTALL.rst) command should always has a `$` before it, and the markdown writing style is not uniformed.

naved001:
I can fix it. Besides putting a $ before a command, what other syntax errors are there?

naved001:
I have made some changes you can see [here](https://github.com/CCI-MOC/hil/compare/master...naved001:cleanup_677)
What other markup errors do you see?

shwsun:
@naved001 apparently some places in our doc is not right. What I want for this issue is to take a good look at serval important file first and address the ill-formatted part first.

naved001:
I am not sure what changes to make apart from the ones that I've already made. Could you be more specific about the issues in the docs.

naved001:
Just revisited `INSTALL.rst` it mostly looks fine to me. 

>  command should always has a $ before it

I actually think that's a bad idea because most of the time I want to copy and paste whole code blocks, putting the dollar signs before every command would make it painful to copy and paste it.

@shwsun 

I am in the favor of closing this issue because I don't see any glaring issues with the documentation formatting.

shwsun:
By just look at the INSTALL.rst, I can spot a few places that are confusing. For example, we dont have a clear separation of things we want the user to type in the terminal and things that should goes into a file . Everything is fine with us simply because we worked on this so long and can just ask around. 

I don't think it is a high priority job but I dont think we should close it.

xuhang57:
here too: https://github.com/CCI-MOC/hil/blob/master/docs/INSTALL-devel.rst

And `TESTING` in `INSTALL.rst`.

zenhack:
I'd like to actually come up with a list of actionable items for this issue; Already mentioned:

- [ ] we dont have a clear separation of things we want the user to type in the terminal and things that should goes into a file

As it stands this issue is too open ended.

xuhang57:
Maybe we should group all the documentation issues and find a sprint or two to solve them all together. Setting up the guidelines for the documentation would come first of course.

zenhack:
[...]

Issue #648, opened 2016-09-01
Document how to make a HIL server handle more requests


henn:
There are certain things one can do to handle more requests that we should document. We've hit these while running various benchmarks that are trying to do things in parallel. These include:

- [ ] Increasing the `processes` attribute in `WSGIDaemonProcess` to something like 128
- [ ] (maybe) increasing the number of apache listeners to accommodate the extra WSGI Daemons.
- [ ] Increasing postgres's max connection count by modifying `max_connections` in `/var/lib/pgsql/data/postgresql.conf`

xuhang57:
@naved001  do we have any documentation about multi-process at all so far?

Issue #411, opened 2015-04-08
Better documentation for HIL consumers/end-users


henn:
Currently, our [end-user/consumer documentation](https://github.com/CCI-MOC/haas/blob/master/docs/USING.rst) does not have any examples.

It would be good to improve this before our next release, including perhaps some scripts in the examples/ directory.

zenhack:
The file you link to expressly refers to `examples/puppet_headnode`; we do have one example, which works. It is rather slapped together though.

henn:
Ok - then we can mark this as easy :)

The documentation should be tied into the README

zenhack:
I'd like to see a concrete list of actionable tasks for this issue; as it stands exactly what needs to be done is somewhat vague.

henn:
Essentially, I was thinking of something that would show the basic operations of HaaS. Perhaps:
- [ ] Creates a project
- [ ] Allocates 2 nodes
- [ ] Creates 2 networks
- [ ] Attaches those networks to the nodes
- [ ] Creates a headnode that is attached to the PXE interfaces
- [ ] Resets the nodes
- [ ] (Bonus points for PXE'ing an OS onto the nodes from the same script)
- [ ] Tears everything down.

zenhack:
Okay, I think that's a good checklist of things to have in an example somewhere. @gsilvis, @lokI8, thoughts?

xuhang57:
@zenhack @naved001 So here we have a good list of examples we should come up with. This relates to: https://github.com/CCI-MOC/hil/issues/903, right? 

Basically, we will have both `README.md` and `Use HIL as a client` as the docs for HIL python client API documentation.

naved001:
you are right, this is just how to use the HIL cli to do something useful. We can probably skip the headnode related call, and pxe booting an OS (that's not really what HIL does).

xuhang57:
Thanks for confirming. So a updated list would be:

- [ ] Creates a project
- [ ] Allocates 2 nodes
- [ ] Creates 2 networks
- [ ] Attaches those networks to the nodes
- [ ] Resets the nodes

zenhack:
Yeah, looks good to me. When we add the openvpn stuff we should also add examples for that (plus links to configuring the openvpn client).

xuhang57:
PR merged. Closing this.

PR #951, opened 2018-02-01
DOCS MATTER!

xuhang57:
Trying to fix ALL docs issue. Any effort on reviewing ongoing changes would be much appreciated.

Starting from:

- [x] https://github.com/CCI-MOC/hil/issues/677#issuecomment-327842422

- [x] https://github.com/CCI-MOC/hil/issues/903

<del>- [ ] https://github.com/CCI-MOC/hil/issues/648</del>

- [x] https://github.com/CCI-MOC/hil/issues/411

- [x] https://github.com/CCI-MOC/hil/issues/312


xuhang57:
For https://github.com/CCI-MOC/hil/issues/648, I don't know anything about `multi-processes in HIL`. Any idea how could I proceed? Or should someone else do it?

AUTHORS
```diff
@@ -1,11 +1,18 @@
...
 Ian Denhardt <ian@zenhack.net>
 Igibek Koishybayev <koishybayev.igibek@gmail.com>
 Jay Hennessey <henn@users.sourceforge.net>
+(Jethro) Shuwen Sun <http://orcid.org/0000-0001-6811-8310>
 Jonathan Bell <jbell@bu.edu>
+Kristi Nikolla <knikolla@massopen.cloud>
+Kyle Hogan <klhogan@bu.edu>
 Logan Bernard <bernard.logan4@gmail.com>
+Lucas Xu <xuh@massopen.cloud>
+Naved Ansari <naved001@massopen.cloud>
```
naved001: I don't have that email address. you could put my gmail or bu email, it's naved001@gmail.com/bu.edu

docs/INSTALL.rst
```diff
@@ -16,11 +16,11 @@ Prerequisite Software
 HIL requires a number of packages available through the CentOS RPM
 repositories, as well as the EPEL repository. EPEL can be enabled via::
 
-    yum install epel-release
+    $ yum install epel-release
```
naved001: oh boy, you went through all the trouble of adding the $ sign everywhere. Sorry I wasn't paying attention that you were taking this up. But if you think about it putting the dollar sign kinda makes it harder to copy and paste it. Earlier, I could triple click and copy the whole line and paste it into a terminal, but with this I would have to remove the dollar sign each time. :(

Don't make this change yet, maybe I copy paste wrong and what you are doing is right. I want to hear from other reviewers or whoever suggested putting the dolla sign there.

docs/INSTALL.rst
```diff
@@ -70,24 +70,24 @@ Copy the following lines in file ``hil_env``
...
-    sudo su -
-    cd /root
-    git clone https://github.com/CCI-MOC/hil
-    cd hil
-    sudo python setup.py install
+    $ sudo su -
+    $ cd /root
+    $ git clone https://github.com/CCI-MOC/hil
+    $ cd hil
+    $ sudo python setup.py install
```
naved001: if you are root at this point, you don't need to sudo.

docs/INSTALL.rst
naved001: > Don't make this change yet,

I mean, don't undo your changes yet.

xuhang57: hah, thought we had a conclusion on that issue.
@shwsun ping~

docs/INSTALL.rst
xuhang57: true. But how could you even `cd /root` if you a not a root user? hmm.

AUTHORS
xuhang57: free like everybody is not using their regular email addresses here. You sure you want to use the gmail? Why not get a massopen.cloud email from Rado just saying ;)

docs/INSTALL.rst
shwsun: I don't quite understand the debate. It is convention to have a `$` denote that it is a command rather something that you should put into a file. We need to distinguish things should be in the config file from commands we want users to run. 

@xuhang57 thank you for your time to clean the HIL doc. While you are at it, could you fix [this](https://github.com/xuhang57/hil/blob/62b5246436c4f5a539c3efd30892966bb6ab044c/docs/INSTALL.rst#user-need-to-choose-appropriate-values-for-their-environment) too? It annoys me a while.

xuhang57: sure, "user needS" and `export` before each line? That's all right?

docs/INSTALL.rst
xuhang57: nvm. `sudo su-`

docs/INSTALL.rst
naved001: > I don't quite understand the debate. It is convention to have a $ denote that it is a command rather something that you should put into a file.

I guess I am not a fan of that convention then, for the reasons I mentioned in my comment.

> We need to distinguish things should be in the config file from commands we want users to run.

I don't know, but I feel it's apparent from the context and that the fact that we mention on top of the block of lines if it goes into a file.

In any case, I'll wait to hear what others have to say; but now I am inclined to follow the convention just because that's the convention.

shwsun: @Izhmash @zenhack @SahilTikale @knikolla  anyone want to weigh in?

zenhack: No strong opinion, but my inclination is to include the `$`, as these are meant to be instructions, not a script -- we should be optimizing for understanding, not copy & paste. The `$` is conventional and provides a useful cue to the reader, even if we do mention above that it's a shell command.

docs/INSTALL.rst
zenhack: Related to the convention discussion above, A common convention is to display root prompts with `#` rather than `$`, which I think is a good idea if we go with the prompt indicators.

AUTHORS
zenhack: Not sure I follow the point about "everybody is not using their regular email addresses here". I only see three massopen.cloud addresses, and all of them are introduced by this patch.

AUTHORS
```diff
@@ -1,11 +1,18 @@
 Abishek Raju <abishek.raju@gmail.com>
+Amin Mosayyebzadeh <mosayyeb@bu.edu>
 Andrew Mohn <AndrewMohn@gmail.com>
 George Silvis, III <george.iii.silvis@gmail.com>
+Ian Ballou ianballou67@gmail.com
```
zenhack: Missing brackets.

docs/USING.rst
```diff
@@ -111,19 +111,94 @@ For nodes using IPMI use the following api call:
...
+::
+
+
+    $ hil project_create mockproject01
+
+4) Register a node in HIL
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+
+Eg> Node name: mocknode01
```
zenhack: This doesn't render particularly well; all of the fields get wrapped on to one line. Also the `Eg>` notation is weird to me.

hil/client/README.md
```diff
@@ -0,0 +1,24 @@
+### HIL Python Client API
```
zenhack: * Start with level 1 headings (one `#`)
* This should go under docs somewhere.
* Probably should call it "client library"
* This could use some introductory verbiage; You're kindof just plopping some code down without any overall context. Also, we don't describe the design of the library at all; it would be good to briefly describe the important objects and sketch its structure (docstrings can serve as comprehensive reference docs, but we should give an overview).

docs/USING.rst
xuhang57: Hmm, I am trying to follow the previous examples. And we actually have two different formats in those two existing examples. I am gonna reformat them then and try to come up with one clean format.

AUTHORS
xuhang57: Sorry for the confusion. I meant many people are using the email addresses that they don't check on a regular basis. And IMHO, hardcoded email addresses is very bad since it is very easy to be scraped and receive spams.

hil/client/README.md
xuhang57: > This should go under docs somewhere.
Personally, I don't like putting this under `/docs` since there are too many docs there already and people probably won't read it. So I put it in the client directory. But if it is the best practice then I will move it there. Maybe creating a new folder called `/client`?

Point 1 & 3 sound good to me.

As for the point 4, I thought we have the overview description when the client library first got introduced to the code base. Will discuss with @naved001 and try to come up with an overview. The idea sounds good to me, but I don't have a plan on how to write one atm.

docs/USING.rst
zenhack: I hadn't noticed it was existing notation, but yeah, let's improve upon it.

hil/client/README.md
zenhack: Re: `/docs`, this is where all of our high-level docs have gone. Maybe we should update the README at the root of the repo to provide a more thorough index, if we want stuff to be easier to find.

I'm looking at the client library as something that people who don't hack on HIL itself might hypothetically use; having it buried inside of the source code doesn't make much sense. It might be more reasonable if this were an internal-use-only module.

Re (4): if we have overview docs, I don't know where they are.

xuhang57:
@zenhack pushed a brief description of the client library. Please let me know if you think it needs to be improved. 👍 

@shwsun @naved001 done.

docs/INSTALL.rst
Izhmash: I am used to seeing `$` when reading documentation.  It helps when you need to quickly scan through a wall of text to piece out what needs to be run in the terminal.

docs/client-library.md
```diff
@@ -0,0 +1,28 @@
+# HIL Client Library
+
+##Description
```
Izhmash: You need a space here between the `##` and `Description`

docs/client-library.md
```diff
@@ -0,0 +1,28 @@
...
+The HIL API Client Library for Python is designed for Python client-application developers. It offers simple access to HIL APIs.
+
+## How to Install HIL Modules?
+```
+pip install git+https://github.com/cci-moc/hil
+```
+
```
Izhmash: We might want to document which client library commands are available if we don't want users to have to look through the source.

xuhang57:
@Izhmash done 👍

xuhang57:
@zenhack @naved001 done.

docs/USING.rst
```diff
@@ -75,10 +75,11 @@ Included herewith are some examples about
...
 
-Eg> Switch name: mockswitch01
-     Host name:  switchhost01
-     User name:  switchuser01
-     Password:   password1234
+   - **Switch name:** mockswitch01
+   - **Switch type:** mock
+   - **Host name:**   switchhost01
+   - **User name:**   switchuser01
+   - **Password:**    password1234
```
zenhack: This is currently being rendered as a block comment; you should un-indent it. There are a couple of other instances of this problem elsewhere.

xuhang57: Was doing that on purpose since that's what one of the examples did previously. Hmm, I think you ar right and it doesn't make sense to block comment them.

xuhang57:
@zenhack done.

zenhack:
Great, merging.

Issue #928, opened 2018-01-02
obmd migration strategy


zenhack:
A spec has been merged describing the changes needed to HIL to integrate with OBMd. One question still floating around in my mind is how we're going to go about migrating existing deployments; Doing this naively would require removing all nodes and re-adding them. It would be nice to find a less disruptive way to deal with this.

Open to input, and in particular insight into how much disruption is acceptable.

naved001:
I guess we would get rid of all obm information from the `node` table (so a migration script for that) leaving only node label and a pointer to it's obmd counterpart, and then another script to add a table for obmd (another migration script for that). The latter might be different if we are going to have a different database altogether. It would be clearer if we first decide how the database is going to look like.

> how much disruption is acceptable.

AFAIK, hil is only deployed on our own clusters that I am managing, so at least I would be fine with some disruption.

@SahilTikale since you are back, would appreciate your thoughts on the discussions here.
@razaaliraza @mosayyebzadeh feel free to contribute to the discussion on the threads here.

zenhack:
We could write a couple helper scripts specifically for this use case:

* One that extracts the node info from the HIL database, and creates the nodes in OBMd.
* One that adds the obmd connection info to the nodes, and possibly deletes the ipmi connection info.

We could also do the database migration in two steps:

1. Add the obm uri & admin token fields to the node model. For now, allow them to be null and don't use them for anything. This is a small patch that can go in now without any other work.
2. The *next* migration script will come with release notes that notify the adminstrator that they must (1) upgrade to the previous version first (i.e. they can't "skip" it), and (2) run the above scripts to set up the data. Then, the migration script sets the obmd connection info fields as not-nullable, and deletes the obm/ipmi tables.

The current design assumes obmd has a separate database.

zenhack:
@naved001, thoughts on the above?

naved001:
okay, so the order in which somebody would run these is:

1. Use helper-script (1) to extract the info from HIL DB and create nodes in OBMd.
2. Then migration-script (1) will add the obm uri and admin token fields, and allow those to be null.
3. Then helper-script(2) will feed the OBMd connection info into the HIL DB. How will this one work? I am imagining updating the rows in the DB directly. (because we can't remove nodes, and re-add them; they might not be free).
4. The final migration-script(2) will delete the old obm tables/columns, and make the OBMd stuff not nullable.

Is that correct?

zenhack:
[...]

PR #953, opened 2018-02-07
obmd data migration script

zenhack:
Per #928.

//cc @naved001

### DIFF
--- a/hil/commands/migrate_ipmi_info.py
+++ b/hil/commands/migrate_ipmi_info.py
@@ -1,0 +1,103 @@
+"""Helper command for extracting IPMI info from the database.
+
+This is part of our obmd migration strategy; the idea is that this
+information is then uploaded to obmd.
+"""
+import sys
+import json
+
+import requests
+
+from flask_script import Command, Option
+
+from hil import server
+from hil.flaskapp import app
+from hil import model
+
+
+class MigrateIpmiInfo(Command):
+    """Migrate ipmi info to obmd"""
+
+    option_list = (
+        Option('--obmd-base-url', dest='obmd_base_url',
+               help='Base url for the obmd api'),
+        Option('--obmd-admin-token', dest='obmd_admin_token',
+               help='Admin token for obmd'),
+    )
+
+    def run(self, obmd_base_url, obmd_admin_token):
+        server.init()
+        with app.app_context():
+            info = db_extract_ipmi_info()
+            obmd_upload_ipmi_info(obmd_base_url, obmd_admin_token, info)
+            db_add_obmd_info(obmd_base_url, obmd_admin_token)
+
+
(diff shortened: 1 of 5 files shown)


hil/commands/migrate_ipmi_info.py
```diff
@@ -0,0 +1,131 @@
+"""Helper commands for moving IPMI info from the database to obmd.
+
+This is part of our obmd migration strategy; see the discussion at:
```
naved001: forgot to link the discussion? #928

hil/commands/migrate_ipmi_info.py
```diff
@@ -0,0 +1,131 @@
...
+        # XXX: apparently we've incorrectly specified the relationship between
+        # Obms and nodes, such that the node attribute returns a list, even
+        # though there can only ever be one node. If we were keeping pluggable
+        # obm support for longer, We'd just fix this, but since we're removing
+        # the functionality soon it's easier to just do this and not have to
+        # worry about what other code we might disturb:
+        node = obm.node[0]
+
+        if obm.type != ipmi_api_name:
+            sys.exit(("Node %s{label} has an obm of unspported "
```
naved001: why not skip it rather than exiting?

xuhang57:
Not familiar with obmd. But overall, Igtm.

hil/commands/migrate_ipmi_info.py
zenhack: Given that we plan on ripping out tge driver mechanism entirely, that would mean potentially losing nodes; this is something an admin ought to know about.

hil/commands/migrate_ipmi_info.py
zenhack: Yep, good spot. Will fix.

Izhmash:
How is the rest of the migration strategy going to be implemented? From issue #928 I see:

>  4. The final migration-script(2) will delete the old obm tables/columns, and make the OBMd stuff not nullable. 

Will this cleanup happen in another file?

zenhack:
@Izhmash, yeah, that would be a migration script added in `hil/migrations/versions/`

zenhack:
...perhaps we need to plan to tag a version between this pr and the one that introduces the migration script? The strategy involves an intermediate upgrade, so we should be able to refer to the specific intermediate version.

Izhmash:
@zenhack I think that sounds like a good idea, yeah. Would definitely help relieve user confusion. What would be the new version number?

zenhack:
@Izhmash, 0.3 is next on the list. We've got a [milestone][1] for that; maybe we should add the second part of this to the 0.4 milestone, and then start focusing on triaging the 0.3 issues so we can get that out the door.

[1]: https://github.com/CCI-MOC/hil/milestone/9

PR #954, opened 2018-02-12
Add list_network_attachments to the client library

xuhang57:
This patch adds the client API function for `list_network_attachments`. The CLI code change accordingly.

This relates to the issue: #924

### DIFF
--- a/hil/client/network.py
+++ b/hil/client/network.py
@@ -21,10 +21,10 @@ class Network(ClientBase):
             url = self.object_url('network', network, 'attachments')
             if project == "all":
                 return self.check_response(self.httpClient.request("GET", url))
-            else:
-                params = {'project': project}
-                return self.check_response(
-                        self.httpClient.request("GET", url, params=params))
+
+            params = {'project': project}
+            return self.check_response(
+                    self.httpClient.request("GET", url, params=params))
 
         @check_reserved_chars()
         def show(self, network):
--- a/hil/client/network.py
+++ b/hil/client/network.py
@@ -16,6 +16,17 @@ class Network(ClientBase):
             return self.check_response(self.httpClient.request("GET", url))
 
         @check_reserved_chars()
+        def list_network_attachments(self, network, project):
+            """Lists nodes connected to a network"""
+            url = self.object_url('network', network, 'attachments')
+            if project == "all":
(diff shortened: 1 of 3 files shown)


hil/client/network.py
```diff
@@ -15,6 +15,17 @@ def list(self):
...
+        @check_reserved_chars()
+        def list_network_attachments(self, network, project):
+            """Lists nodes connected to a network"""
+            url = self.object_url('network', network, 'attachments')
+            if project == "all":
+                return self.check_response(self.httpClient.request("GET", url))
+            else:
+                params = {'project': project}
+                return self.check_response(
+                        self.httpClient.request("GET", url, params=params))
```
naved001: @zenhack what's your opinion on handling a guard block? For example, when I see a code like this, I tend to get rid of the `else:` since that would require less indentation and 1 less line. Just curious what the best practice is.

tests/unit/client.py
```diff
@@ -837,6 +837,11 @@ def test_network_list(self):
                 u'net-05': {u'network_id': u'1005', u'projects': [u'proj-02']}
                 }
 
+    def test_list_network_attachments(self):
+        """ Test lsit of network attachments """
```
naved001: typo: l**is**t

hil/client/network.py
xuhang57: good question. curious too.

xuhang57:
since this patch only adds a function in the client library and we don't have a docs to keep track of all functions we have in client library, I will not make any docs changes along with this PR. 

Done.

hil/client/network.py
zenhack: Yeah, that's my general inclination.

tests/unit/client.py
```diff
@@ -837,6 +837,11 @@ def test_network_list(self):
                 u'net-05': {u'network_id': u'1005', u'projects': [u'proj-02']}
                 }
 
+    def test_list_network_attachments(self):
```
zenhack: Can we have a test for when the result is non empty?

xuhang57: added!

xuhang57:
@naved001 @zenhack Done.

Issue #516, opened 2016-01-22
cli tests erroneously report 0% test coverage.


zenhack:
This happens because we actually launch the cli as a separate process, but it would nonetheless be nice to get the stats on that.

xuhang57:
do we have any code coverage tool to test the code coverage? I don't find it in the ci I believe. Maybe we can use https://coveralls.io/?

zenhack:
We have pytest-cov set up for code coverage, though I don't think it runs automatically in CI. That would be a nice addition (feel free to open a separate issue), but what this issue is about is that the way the coverage tool works, it doesn't track coverage from forked processes, so when we call the command line tool from inside of a test, it doesn't notice that we've executed code inside the cli tool.

xuhang57:
so we need someone to send a PR for enabling `pytest-cov` in CI right?

zenhack:
Hm, the relevant bits in setup.cfg are actually in there commented out. Commit 14d6be62 seems to be the patch that removed it. The commit message suggests it was causing failures? @SahilTikale, that commit has your name on it, do you remember the details?

The other thing is, it would still be nice to integrate with GitHub somehow, so we can get the results without going and looking at the travis log.

xuhang57:
coverall has a bot can post the coverage report as a comment in PR. See this example:https://github.com/CCI-MOC/giji-backend/pull/31

zenhack:
I'm fine with adding coveralls to the CI.

xuhang57:
@zenhack I don't have the access to enable it. I believe only the admins could. Maybe one of the admins could add it.

zenhack:
I don't think I have admin access to this repository either. @naved001?

naved001:
both of you are admins now. Sorry I've been busy with some infrastructure stuff, will get back to HIL soon enough.

xuhang57:
hmm, it turns out being an admin on the repo is not good enough. You need to be an organization admin to enable it from the `coveralls.io`. For some reasons, I cannot search coverall from the settings.

ping @pns005 @knikolla @okrieg @pjd-nu

Issue #535, opened 2016-02-23
CLI code needs test coverage


henn:
Currently, we only test `haas serve` and `haas serve_networks`.

We need to test more. This should be done after #431.

henn:
This should also include:

- [ ] Testing the various `HTTPClient` methods (currently in cli.py).
- [ ] For testing KeystoneHTTPClient, it might make sense to modify `tests/integration/keystone.py`.
- [ ] Don't forget to include testing GET parameters, per #655

xuhang57:
Given @naved001  is working on the new CLI, this issue should be triaged after the new CLI is merged.

PR #957, opened 2018-02-13
Add code coverage with coveralls.io

xuhang57:
Relates to:
https://github.com/CCI-MOC/hil/issues/535
https://github.com/CCI-MOC/hil/issues/516


naved001:
our coverage is 10.028 %? that seems inaccurate.

xuhang57:
I believe a reason why is that it is testing all branches. Still trying to figure out the root cause.

xuhang57:
So, I don't know why we had disabled the code coverage tests before.

But it turns out, If I enable it, in Travis (not through the setup.cfg), lots of tests are failing.

Now, it would be great if someone could help me with:

1. why enable the coverage test in setup.cfg? 
2. why these tests are failing, after enabling the coverage tests?

Thanks

xuhang57:
Also, with having `cli.py` and `test/unit/cli.py` together (and some other files), It gets an error about `import mismatch`.
See here:https://travis-ci.org/CCI-MOC/hil/jobs/341489660#L3171

Changing `./cli.py` to `/test_cli.py`

xuhang57:
So now it is 68%. For unit tests.

xuhang57:
screen shot 2018-02-15 at 11 49 31

ci/run_unit_tests.sh
```diff
@@ -8,7 +8,7 @@ if [ $DB = sqlite ]; then
     extra_flags='-n auto'
 fi
 
-py.test $extra_flags \
+py.test --cov=hil $extra_flags \
```
zenhack: This is going to miss a bunch of stuff, since we have other scripts that run integration tests with e.g. keystone, apache, obmd.

xuhang57: true, but for some reason, when I add the `--cov=hil` to all the `py.test`, the coverage report returns a 0.0%. I suspect it is expecting us to have a one `py.test` to run all the tests, at one time. Still investigating.

xuhang57:
@zenhack done. I forgot to use `--cov-append` before

zenhack:
Ok, this looks mostly reasonable to me. Can we preview the results somewhere? you have a screenshot above.

xuhang57:
@zenhack  if you click the `Details` on the CI test, you will see two tests, click on `coverage/coveralls`, you can see the details about the coverage: e.g., why we have a 43% coverage.

screen shot 2018-02-18 at 20 27 46

zenhack:
Looks like the biggest sources of lacking coverage are:

* The hardware drivers; no surprise, since we don't run the tests for real hardware in travis, because we can't.
* hil.cli; this is #516 at work.
* hil.test_common. I suspect the code we're missing is stuff that would otherwise be excercised by the hadware driver tests. If there are actually any "real" gaps in coverage there it means we have code we're not even using...

I'm okay saying fixing #516 is out of scope for this pr, and figuring out a solution separately.

Is it possible to blacklist individual modules for test coverage? It would be nice to have it just not include the hardware drivers in it's accounting, since it's just noise.

xuhang57:
Will look into that and get back to you later.

xuhang57:
@zenhack , If we separate the tests for hardware drivers from other tests, I think it is very easy to not testing the coverage. Looking into that now.

Also, as for #516, it is saying the test coverage is 0.0%. However, I believe in the current test coverage report, the test coverage is about ~40%. I am not sure whether #516 is still relevant or not.

zenhack:
Coveralls does seem to report the coverage of that module correctly:

https://coveralls.io/builds/15613016/source?filename=hil/cli.py

So we can probably close #516 once this is merged.

.coveragerc
```diff
@@ -0,0 +1,4 @@
+[run]
+omit = 
+    hil/ext/obm/*
+    hil/ext/switches/*
```
zenhack: Ideally we would still cover the mock drivers.

naved001:
> @naved001 ping. So we should decide what not to test. Any suggestions?

`hil/ext/obm/mock.py`
`hil/ext/switches/mock.py` 

These are the mock drivers we could cover. The CI runs the deployment tests with mock drivers, so I am sure calls to the mock switch will be made.

All the other drivers can be ignored.

xuhang57:
@naved001 @zenhack  We do not include the test coverage for deployment tests. Should we cover that also?

zenhack:
Yeah, we should add it to that one too.

xuhang57:
So I took a look at the unit tests for obm and switches. I couldn't find any tests with `mock drivers`. If I am correct, then there is no point for covering them. We should send another PR with adding tests and enable the coverage.

if I am not correct, could you guys please let me know where the tests are? @zenhack @naved001 

Thanks!

zenhack:
I'm happy.

.coveragerc
```diff
@@ -0,0 +1,13 @@
+[run]
+omit =
+    hil/ext/obm/impi.py
+    hil/ext/switches/brocade.py
+    hil/ext/switches/_console.py
+    hil/ext/switches/edllnos9.py
```
jeremyfreudberg: Typo... And I still don't understand why these files are skipped

setup.py
```diff
@@ -98,6 +98,8 @@ def _get_readme():
                 'pycodestyle>=2.3.1',
                 'pylint>=1.6.0,<2.0',
                 'requests_mock>=1.0.0,<2.0',
+                'coverage==4.5.1',
```
jeremyfreudberg: Why such a specific version pin

.coveragerc
zenhack: Because we don't use these in travis, because they are drivers for hardware that travis doesn't have access to. So the coverage will always be 0%, and it's just noise.

setup.py
xuhang57: I picked the latest version atm. 

https://coverage.readthedocs.io/en/coverage-4.5.1/changes.html

.coveragerc
jeremyfreudberg: Sorry, I still don't understand. For two (interconnected) reasons, actually, I'm confused.

1. There seem to be files in tests/unit/ which correspond to these skipped drivers. So are there not unit tests?
2. If there really aren't unit tests for those files, then don't you deserve that 0% anyway?

I think I must be missing something obvious here...

setup.py
jeremyfreudberg: So if we had picked up this work next month instead of today then the pin would be different.... That's a bit too serendipitous for me

xuhang57: True. I have the same feeling for `pycodestyle`.  Maybe I should use `coverage>=4.5.1` 

But I don't know whether we should keep pursuing the latest version of a package. Since the latest version is not really guaranteed bug-free and it might have conflicts with other packages.

.coveragerc
zenhack: @jeremyfreudberg, we do have tests, but we don't run them in travis because they require physical hardware that isn't accessible from travis -- they are hardware drivers after all. We try to be good about running them manually when we change stuff pertinent to the hardware.

setup.py
zenhack: I agree pinning this to a specific version is not a good idea. See the comment above this in setup.py for general policy; let's do `>=4.5.1,<5.0.0`.

.coveragerc
zenhack: (There are a couple tests that mention some of the drivers under `tests/unit/`, but those are migrations and stuff for the client library; they don't really poke the functionality of those modules. The relevent tests are under `deployment/`).

jeremyfreudberg: Fair enough. I think it's okay to omit these from the coverage calculation.

(I did the quickest of pass-throughs over the skipped driver files... there's nothing there that couldn't have a unit test if we actually wanted to. So like I said earlier, we could say that the lack of tests does mean 0% coverage. But really a lot of the unit tests that could end up being written would just be the `assert_called_with` over and over, which is arguably not useful... so I'm fine with the way things are now, at least in the context of this PR.)

.coveragerc
```diff
@@ -0,0 +1,13 @@
+[run]
+omit =
+    hil/ext/obm/impi.py
+    hil/ext/switches/brocade.py
+    hil/ext/switches/_console.py
+    hil/ext/switches/dllnos9.py
```
naved001: you missed the e. d**e**llnos9.py

naved001:
@jeremyfreudberg are you happy with this PR now?  ❌ ➡️ ✔️

PR #958, opened 2018-02-14
 Require setting a native vlan before adding any trunk vlans on the brocade switch (and correctly parse range of vlans)

naved001:
* Moved that fix from the dellnos9 driver to switches.common and then used it
in brocade.

* Fixes #728 

It's a work in progress because `tests/deployment/vlan_networks.py` is failing. Gotta dig into that and push a fix asap.

Issue #709, opened 2017-01-25
Document port initial states


henn:
It's not clear to me what state a port is expected to be in when it is registered with HIL.

For example, I just registered a Dell powerconnect port with HIL, ensured that it had no VLANs configured and was in access mode and connected it through HIL to a network. The result was that *every* VLAN ended up trunked to it.

#683 will help with this, though we'll still need to do the analysis for each driver to figure out what the initial state should be before networks are added to a port.

naved001:
> ensured that it had no VLANs configured and was in access mode and connected it through HIL to a network. 

Afaik, in the powerconnects, when a new port is put into trunk mode it has access to every VLAN by default.

We should put that in `network-drivers.md` or do what I suggested here  #761

zenhack:
Yeah, we should document this. I left a comment in #761 about that solution, but either way it seems like we could just add a change to the docs now, and worry about automating things later.

xuhang57:
PR merged. Closing this.

PR #959, opened 2018-02-14
Document the initial port state

naved001:
* and some general cleanup.

While writing this doc, I realized that for the brocade switch, we should probably turn off the switchport until a native vlan is set, otherwise it will be sitting on the default vlan of 1. Will do that in #958 

* Issue #709

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -92,7 +92,7 @@ usual confirmation dialog for new hosts:
 
 ...which the HIL drivers do not handle.
 
-* Please make sure that there are no unwanted vlans on a swichport when it is
+* Please make sure that there are no unwanted VLANs on a switchport when it is
 registered in HIL. There are more details about this in the description for
 supported switches.
 
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -80,19 +80,10 @@ even the root user is untrusted.
 
 ### Switch configuration
 
-In order to deploy HIL, at least one switch supported by one of HIL's
-drivers is required.
+* In order to deploy HIL, at least one switch supported by one of HIL's
+drivers is required. See the list of [supported switches](network-drivers.html)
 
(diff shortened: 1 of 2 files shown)


docs/deployment.md
```diff
@@ -101,6 +92,10 @@ usual confirmation dialog for new hosts:
 
 ...which the HIL drivers do not handle.
 
+* Please make sure that there are no unwanted vlans on a swichport when it is
```
xuhang57: *switchport? and maybe *VLANS? Just want it to be consistent.

docs/network-drivers.md
```diff
@@ -216,6 +223,30 @@ same format accepted by the underlying switch, in this case (e.g.)
...
+1. Make sure the REST server is enabled on the switch.
+
+```
+sw0# config
+sw0(config)# rbridge-id 1
+sw0(config-rbridge-id-1)# no http server shutdown
+```
+
+2. For every switchport that is registered in HIL, make sure that there are no
+vlans attached to that port. Toggling switchport command usually does the trick.
```
xuhang57: same as above. *VLANS?

naved001:
@xuhang57 done

docs/network-drivers.md
```diff
@@ -75,8 +75,6 @@ HIL:
 
 * ``hil.ext.switches.dell``, which provides a driver for the Dell
   Powerconnect 5500 series switches.
-* ``hil.ext.switches.n3000``, which provides a driver for the Dell N3000
```
zenhack: Why are we removing this?

naved001: idk how but it ended up there twice, so I am just removing the copy.

zenhack: Ah, makes sense (the other one didn't show in the diff, so I was confused...)

PR #961, opened 2018-02-15
Remove pytest-catchlog from dependences.

zenhack:
We've been getting warnings about it having been merged into pytest
core, e.g:

https://travis-ci.org/CCI-MOC/hil/jobs/341455006#L2725

This also bumps the minimum version for pytest in our setup.py, since
the change was made in 3.3.0:

* https://docs.pytest.org/en/latest/changelog.html#id56
* https://github.com/pytest-dev/pytest/pull/2794

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -92,8 +92,7 @@ setup(name='hil',
                         ],
       extras_require={
           'tests': [
-                'pytest>=3.0.0,<4.0',
-                'pytest-catchlog>=1.2.2,<2.0',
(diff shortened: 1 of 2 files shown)


naved001:
I tried doing this in #927 and when I removed pytest-catchlog from the requirements, it suddenly couldn't import it. So don't know what's up with that.

zenhack:
@naved001, found the problem, see the last commit. D'oh.

Issue #948, opened 2018-01-31
Update switch drivers to handle errors


naved001:
#929 sets up a way for handling errors raised by switches. 

* For brocade and dellnos9 switches, their `_make_request` function log errors. I think we can just modify it to raise a switch error which the networking server will handle.

* For our pexpect drivers, I don't know if we can do much. Maybe catch some pexpect timeout error (when connecting to the switch) and raise SwitchError (or subclass it). That should do it.

@SahilTikale tagging you on this since you are working on two switch drivers right now and you may want to keep this in mind.

PR #963, opened 2018-02-27
Update Switch Drivers to Handle Errors

xuhang57:
Since #929 introduces an approach of handling switch errors, this PR updates the existing switch drivers.

Relates to: https://github.com/CCI-MOC/hil/issues/948

### DIFF
--- a/hil/ext/switches/brocade.py
+++ b/hil/ext/switches/brocade.py
@@ -302,6 +302,7 @@ class Brocade(Switch, SwitchSession):
         r = requests.request(method, url, data=data, auth=self._auth)
         if r.status_code >= 400 and \
            r.status_code not in acceptable_error_codes:
-            logger.error('Bad Request to switch. Reason: %s', r.reason)
-            raise SwitchError('Bad Request to switch. Reason: %s', r.reason)
+            logger.error('Bad Request to switch. 
+                          Response: %s and Reason: %s' % (r.text, r.reason))
+            raise SwitchError('Bad Request to switch. Response: %s and Reason: %s' % (r.text, r.reason))
         return r
--- a/hil/ext/switches/brocade.py
+++ b/hil/ext/switches/brocade.py
@@ -299,8 +299,9 @@ class Brocade(Switch, SwitchSession):
 
     def _make_request(self, method, url, data=None,
                       acceptable_error_codes=()):
-        try:
-            r = requests.request(method, url, data=data, auth=self._auth)
-        except SwitchError:
+        r = requests.request(method, url, data=data, auth=self._auth)
+        if r.status_code >= 400 and \
+           r.status_code not in acceptable_error_codes:
             logger.error('Bad Request to switch. Response: %s', r.text)
+            raise SwitchError('Bad Request to switch')
         return r
--- a/hil/ext/switches/brocade.py
+++ b/hil/ext/switches/brocade.py
@@ -14,6 +14,7 @@ from hil.migrations import paths
 from hil.model import db, Switch, SwitchSession
 from hil.errors import BadArgumentError
 from hil.model import BigIntegerType
+from hil.errors import SwitchError
 
 paths[__name__] = join(dirname(__file__), 'migrations', 'brocade')
 
@@ -298,8 +299,8 @@ class Brocade(Switch, SwitchSession):
 
     def _make_request(self, method, url, data=None,
                       acceptable_error_codes=()):
-        r = requests.request(method, url, data=data, auth=self._auth)
-        if r.status_code >= 400 and \
-           r.status_code not in acceptable_error_codes:
-            logger.error('Bad Request to switch. Response: %s', r.text)
+        try:
+            r = requests.request(method, url, data=data, auth=self._auth)
+        except SwitchError:
+            logger.error('Bad Request to switch. Response: %s', r.text)    
         return r
--- a/hil/ext/switches/brocade.py
+++ b/hil/ext/switches/brocade.py
@@ -302,5 +302,5 @@ class Brocade(Switch, SwitchSession):
         try:
             r = requests.request(method, url, data=data, auth=self._auth)
         except SwitchError:
-            logger.error('Bad Request to switch. Response: %s', r.text)    
+            logger.error('Bad Request to switch. Response: %s', r.text)
         return r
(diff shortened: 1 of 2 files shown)


hil/ext/switches/brocade.py
```diff
@@ -298,8 +299,8 @@ def _construct_tag(name):
 
     def _make_request(self, method, url, data=None,
                       acceptable_error_codes=()):
-        r = requests.request(method, url, data=data, auth=self._auth)
-        if r.status_code >= 400 and \
-           r.status_code not in acceptable_error_codes:
-            logger.error('Bad Request to switch. Response: %s', r.text)
+        try:
+            r = requests.request(method, url, data=data, auth=self._auth)
+        except SwitchError:
```
xuhang57: @naved001 ping. Not sure whether this is the correct approach of handling the error. If not, would you mind telling me what would be the correct way?

For now, I am not sure whether `requests.request()` here will raise a `SwitchError`. Is there a way to test this? 

Thanks!

hil/ext/switches/brocade.py
```diff
@@ -298,8 +299,8 @@ def _construct_tag(name):
 
     def _make_request(self, method, url, data=None,
                       acceptable_error_codes=()):
-        r = requests.request(method, url, data=data, auth=self._auth)
-        if r.status_code >= 400 and \
-           r.status_code not in acceptable_error_codes:
+        try:
+            r = requests.request(method, url, data=data, auth=self._auth)
+        except SwitchError:
```
naved001: this won't work because `request` won't raise a switcherror. 
what we should do is
 ```python
if r.status_code >= 400 and r.status_code blah blah:
    logger.error('Bad request')
    raise SwitchError('BadRequest')
```
we raise the error here, and the networking daemon will catch it and handle it.

**Comment by xuhang57** [comment]
hmm, I don't know whether I should add some test cases to test the lines I just added. Since we are not testing the coverage on actual drivers, which makes sense that I am adding a few lines that are not covered. 

What do you guys think? If I do need to add the tests, any idea on what kind of tests I should have would be really helpful!

**Comment by naved001** [comment]
> If I do need to add the tests, any idea on what kind of tests I should have would be really helpful!

Just spitballing here:

* Add one more node to the site-layout.json file which is connected to a non-existent switchport. 
* Then, let's say in `tests/deployment/vlan_networks.py`, add another test which simply connects a network to that switchport. This should fail and then you can catch the exception, and verify related things.

But I don't know what are we testing here. The mechanism is already tested by `tests/unit/deferred.py` using the mock switch. 

what do you think @zenhack

**Comment by zenhack** [comment]
I don't feel terribly strongly about adding tests for this. It might be nice to tweak the tests so we can run them against any driver instead of just `mock`, but that's a bigger change that I don't think it's worth holding this up over. @naved001 it might be good to run the deployment tests against the brocade switch if you've got the bandwidth for that.

I'm confused as to why coveralls is complaining; we marked the brocade driver as omit. it's also displaying paths in the virtualenv, which is weird. We've got some kind of configuration problem I think.

GitHub is telling me that I can still hit merge if I want to, so maybe we can just deal with coveralls config problem separately.

**Comment by naved001** [comment]
> @naved001 it might be good to run the deployment tests against the brocade switch if you've got the bandwidth for that.

Yeah I can do that, but to be honest it doesn't really change much, if an error is raised it will also log it as an error and the tests will fail since we have the `fail on log warnings`. I'll run it anyway.

> GitHub is telling me that I can still hit merge if I want to, so maybe we can just deal with coveralls config problem separately.

Yeah, coveralls is optional right now (unlike Travis, which is required). So we can ignore it as per our convenience.

**Comment by xuhang57** [comment]
But, yes, I am also very confused on why the coverage drops. I don't have a clue yet. 

The thing is, you cannot drop the overage percentage when you have something that has already been 0.0%. 

F = file

F1 (0.0%) + F2 + F3 + F4
-------------------------    = 38.4%
                 Files

**Comment by zenhack** [comment]
@xuhang57, remember, you're making the uncovered file bigger, so that is going to drop the percentage of the overall code base. My question is more "why isn't coveralls ignoring this file?"

**Comment by xuhang57** [comment]
@zenhack Make sense. 

I am basically following this: http://coverage.readthedocs.io/en/coverage-4.2/source.html#source

Would be nice if there is a second person that can read this doc a bit and make sure it suits what we need.

Currently, coveralls basically set certain files to be 0.0% rather than ignoring them I believe.

**Comment by naved001** [comment]
@zenhack Deployment tests passed with a caveat (unrelated to this PR):

for `vlan_networks.py` deployment tests if I specify a range of vlans in `testsuite.cfg` then the switch returns a [range of vlans](https://github.com/CCI-MOC/hil/pull/958#issuecomment-368948978) for which I have pushed a fixed in #958. The simple workaround was to specify non contiguous VLANs.

File: hil/ext/switches/brocade.py [hunk 3]
```diff
@@ -302,4 +303,5 @@ def _make_request(self, method, url, data=None,
         if r.status_code >= 400 and \
            r.status_code not in acceptable_error_codes:
             logger.error('Bad Request to switch. Response: %s', r.text)
+            raise SwitchError('Bad Request to switch')
```
naved001: I called `_get_native_vlan` on a non existing port to raise the error. 
And this was the result:
```
(Pdb) p r.status_code
404
(Pdb) p r.text
u''
(Pdb) p r.reason
'Not Found'
(Pdb) dir(r)
['__attrs__', '__bool__', '__class__', '__delattr__', '__dict__', '__doc__', '__enter__', '__exit__', '__format__', '__getattribute__', '__getstate__', '__hash__', '__init__', '__iter__', '__module__', '__new__', '__nonzero__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__setstate__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', '_content', '_content_consumed', '_next', 'apparent_encoding', 'close', 'connection', 'content', 'cookies', 'elapsed', 'encoding', 'headers', 'history', 'is_permanent_redirect', 'is_redirect', 'iter_content', 'iter_lines', 'json', 'links', 'next', 'ok', 'raise_for_status', 'raw', 'reason', 'request', 'status_code', 'text', 'url']
```
So I think it might be a good idea to log `r.reason` in addition to `r.text` so that we have enough information to work with when debugging. I think `method` and `url` might be useful too.

xuhang57: sounds great! I will use `r.reason`

zenhack:
LGTM. I'm going to merge, even though coveralls is unhappy (we've discussed the issue enough I think).

PR #964, opened 2018-02-27
WIP: Save to flash on brocade switch

naved001:
* In response to issue #899 

* Please don't judge me for using pexpect for getting the startup-config file in the brocade driver. That was  the only way to do it, I figured. 

* The method `get_config()` of brocade is kinda gross right now and needs some work, but I am opening this PR to get feedback about the general approach.

* `tests/deployment/switch_config.py` passed for this.

### DIFF
--- a/hil/ext/switches/brocade.py
+++ b/hil/ext/switches/brocade.py
@@ -16,7 +16,8 @@ from hil.model import db, Switch, SwitchSession
 from hil.errors import BadArgumentError
 from hil.model import BigIntegerType
 from hil.errors import SwitchError
-from hil.ext.switches.common import check_native_networks, parse_vlans
+from hil.ext.switches.common import check_native_networks, parse_vlans, \
+ should_save
 
 
 paths[__name__] = join(dirname(__file__), 'migrations', 'brocade')
@@ -96,11 +97,15 @@ class Brocade(Switch, SwitchSession):
             else:
                 assert new_network == vlan_id
                 self._add_vlan_to_trunk(interface, vlan_id)
+        if should_save(self):
+            self.save_running_config()
 
     def revert_port(self, port):
         self._remove_all_vlans_from_trunk(port)
         if self._get_native_vlan(port) is not None:
             self._remove_native_vlan(port)
+        if should_save(self):
+            self.save_running_config()
 
     def get_port_networks(self, ports):
         """Get port configurations of the switch.
@@ -294,12 +299,46 @@ class Brocade(Switch, SwitchSession):
         if self._check_bna_config_status(session_id):
             logger.debug('Saved running config to startup config')
         else:
-            print "failed"
             logger.debug('Configuration not saved in reasonable time')
 
     def get_config(self, config_type):
-        url = self.hostname + '/rest/config/' + config_type
-        response = self._make_request('GET', url)
+        # This is the only method that uses pexpect to connect to the switch
+        # because the REST API does not return the startup config file.
+        # This method is only used by the test suite, the rest of the switch
+        # functionality does not care about this method.
+
+        # this is how we get the running config file using the API
+        # I couldn't find a way to get the startup config file similarly.
+        # url = self.hostname + '/rest/config/' + config_type
+        # response = self._make_request('GET', url)
+
+        from hil.ext.switches._console import login, get_prompts
+
+        # is this safe to assume that there will be http in front of the
+        # hostname? also create a deep copy rather than modifying the original
+        # object
+        original_hostname = self.hostname
+        self.hostname = self.hostname.replace('http://', '')
+        console = login(self)
+        self.hostname = original_hostname
+
+        console.sendline('')
+        prompts = get_prompts(console)
+
+        console.sendline('terminal length 0')
+        console.expect(r'[\r\n]+.+# ')
+
+        # dunno why, but without asking twice, I didn't get the desired
+        # response. I'll debug this
+        console.sendline('show ' + config_type + '-config')
+        console.expect(r'[\r\n]+.+# ')
+        console.sendline('show ' + config_type + '-config')
+        console.expect(r'[\r\n]+.+# ')
+
+        config = console.after
+        console.sendline('terminal length 40')
+        console.sendline('exit')
+        return config
 
     def _check_bna_config_status(self, id):
         """ This function checks if the bna_config_command was successful by
@@ -324,7 +363,6 @@ class Brocade(Switch, SwitchSession):
             root = etree.fromstring(response.text)
             status = root.find(tag).text
             if status == 'completed':
-                print RETRIES
                 return True
             else:
                 RETRIES -= 1
@@ -374,4 +412,3 @@ class Brocade(Switch, SwitchSession):
                               'Response: %s and '
                               'Reason: %s', r.text, r.reason)
         return r
-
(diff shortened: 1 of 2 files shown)


hil/ext/switches/brocade.py
```diff
@@ -265,6 +271,89 @@ def _disable_native_tag(self, interface):
...
+        # is this safe to assume that there will be http in front of the
+        # hostname? also create a deep copy rather than modifying the original
+        # object
+        original_hostname = self.hostname
+        self.hostname = self.hostname.replace('http://', '')
+        console = login(self)
+        self.hostname = original_hostname
+
+        console.sendline('')
+        prompts = get_prompts(console)
```
naved001: why am I getting the prompts when I don't even use them

hil/ext/switches/brocade.py
```diff
@@ -265,6 +271,89 @@ def _disable_native_tag(self, interface):
...
+        # dunno why, but without asking twice, I didn't get the desired
+        # response. I'll debug this
+        console.sendline('show ' + config_type + '-config')
+        console.expect(r'[\r\n]+.+# ')
+        console.sendline('show ' + config_type + '-config')
+        console.expect(r'[\r\n]+.+# ')
+
+        config = console.after
+        console.sendline('terminal length 40')
+        console.sendline('exit')
```
naved001: should expect EOF here.

hil/ext/switches/brocade.py
```diff
@@ -265,6 +271,89 @@ def _disable_native_tag(self, interface):
...
+        Returns a boolean indicating success/failure.
+        """
+
+        url = self.hostname + '/rest/' +  \
+            'operational-state/bna-config-cmd-status'
+        body = '<bna-config-cmd-status><session-id>%s</session-id> \
+                </bna-config-cmd-status>' % id
+        tag = '{urn:brocade.com:mgmt:brocade-ras}status'
+
+        # Check status 6 times and sleep for 2 seconds before checking again.
```
naved001: this means, that every `modify_port` and `revert_port` operation will be blocked for up to 12 seconds. Should I be calling this method with a new thread, @zenhack ?

hil/ext/switches/brocade.py
xuhang57: maybe you don't want to save `get_prompts(console)` as a variable called `prompts` here. Since you are not using it later on in the code.

tests/deployment/switch_config.py
```diff
@@ -53,17 +53,6 @@ def configure():
                                      'site_layout')
 
 
-@pytest.fixture
-def is_brocade():
-    """open the site-layout file to see if we have a brocade switch"""
-
-    with open('site-layout.json') as layout_data:
-        layout = json.load(layout_data)
```
xuhang57: so the entire file is not using `json`, you could get rid of the `import json`

xuhang57:
Current, I cannot offer a good review. Catching up with how switches have been used in hil generally.

tests/deployment/switch_config.py
naved001: right

hil/ext/switches/brocade.py
naved001: yeah, ill remove that.

naved001:
@zenhack hey Ian, let me know if you are okay with this approach before I spend more time ironing this. A more careful review can be done when you have more time on hand. Thanks!

hil/ext/switches/brocade.py
```diff
@@ -280,6 +286,89 @@ def _disable_native_tag(self, interface):
...
+        # functionality does not care about this method.
+
+        # this is how we get the running config file using the API
+        # I couldn't find a way to get the startup config file similarly.
+        # url = self.hostname + '/rest/config/' + config_type
+        # response = self._make_request('GET', url)
+
+        from hil.ext.switches._console import login, get_prompts
+
+        # is this safe to assume that there will be http in front of the
```
naved001: it's a terrible assumption

xuhang57: chrome will soon not support address without SSL.

naved001: oh didnt know that. this switch doesn't have a web interface anyway.

xuhang57: well, maybe I used the wrong word. It still supports but marks 'insecure' 🤦‍♂️ 
http://www.eweek.com/security/google-chrome-to-label-sites-using-http-as-insecure-starting-in-july

zenhack: Chrome is irrelevant here anyway, since we're not accessing this through a browser.

zenhack:
I'm okay with using the console interface to get the config if there's no way to do that via the http api. I'll do a proper review Monday sometime, sorry it's taken me a while to get to it.

hil/ext/switches/brocade.py
zenhack: How long does it typically take? I think the answer depends.

hil/ext/switches/brocade.py
```diff
@@ -280,6 +286,89 @@ def _disable_native_tag(self, interface):
...
+        response = self._make_request('POST', url, data=body)
+
+        # check if it succesfully completed and log it
+        tag = '{urn:brocade.com:mgmt:brocade-ras}session-id'
+        root = etree.fromstring(response.text)
+        session_id = root.find(tag).text
+        if self._check_bna_config_status(session_id):
+            logger.debug('Saved running config to startup config')
+        else:
+            logger.debug('Configuration not saved in reasonable time')
```
zenhack: I think this should be logged at error level; if the config isn't saving correctly that's something a sysadmin should respond to.

hil/ext/switches/brocade.py
zenhack: It would make more sense to just parse the URL properly and extract the host; there are perfectly usable modules in the standard library for this.

hil/ext/switches/brocade.py
naved001: It takes 8 seconds typically. To avoid complicated code, I could just not check this at all.

zenhack:
That is a bit much. I don't like the idea of not checking it though.

My worry with just launching a thread is that we have no backpressure
then; theoretically requests could just keep piling up.

We could have one thread processing a queue:

    https://docs.python.org/2/library/queue.html

..which we set a reasonable bound on, so we do have some backpressure,
though there will be no delay in the common case.

But I also don't like the complexity of that; it seems like it would be
hard to test. Additionally, I'm a bit bothered by the fact that saving
to flash is currently best-effort; we should think through the
implications of a failure + subsequent reset, and make sure we design
this in a way that doesn't open us up to vulnerabilities.

naved001:
one fix to slightly alleviate this problem is to call `save_config` when we disconnect from the switch rather than every time we do a `revert_port` or `modify_port`, that way we only save after we perform a whole bunch of operations. I think we should do this in other switches too. Will think about your other concerns too.

naved001:
Ran this by @okrieg today. 

He feels that we shouldn't be saving the state to switch at all if we can't always guarantee that. We don't know how to handle the scenario if the save fails.

 So what he suggested is that, when HIL is deployed on a switch, the startup-config should have all VLANs disabled on HIL controlled switchports (I remember @zenhack suggested this somewhere) so in case of power failures we fail safe. In addition to that, we provide an admin only API that restores the switch to a state what HIL thinks it should be in (that had been discussed before too). 

I think these changes would obviate the need for switches having to save at all. Let's mull over it a little before we decide what to move forward with.

@xuhang57 please add if I missed something during the meeting.

naved001:
I am gonna close this then.

Issue #544, opened 2016-03-09
_assert_absent and friends should have better names


zenhack:
Problems:
1. They're not assertions; assertions indicate a bug in the code, whereas these can (and will) fire under very normal circumstances.
2. The leading underscore is a convention that indicates private, but since we've started moving certain API calls (e.g. `user_create`) outside of `haas.api`, they're now used by external modules. I did this because there didn't seem to be much point in changing the implementation, but we ought to just explicitly expose them.

Proposal:
- `_assert_absent` becomes `absent_or_conflict`
- `_assert_absent_n` becomes `absent_child_or_conflict`
- `_must_find` becomes `get_or_404`
- `_must_find` becomes `get_child_or_404`

Just a first pass; open to suggestions.

//cc @henn

naved001:
> They're not assertions; assertions indicate a bug in the code

zenhack:
@naved001, it's not so much an English thing as a convention specific to a programming context.

I don't like `ensure`, as that generally implies changing the state if it's incorrect, rather than just aborting. `require` would be better. But the names I suggest above have the advantages of (a) being explicit about what happens if the object is not absent, and (b) having symmetry with the latter two names.

xuhang57:
How about changing it to `声称_absent`?

just kidding.

I agree that `assert` should use in the test files rather than in the actual code. However, I am not sure about changing the `_must_find` since it is pretty common in Python?

### After searching

`must_find`:
https://github.com/search?l=Python&q=must_find&type=Code&utf8=%E2%9C%93
`get_or_404`:
https://github.com/search?l=Python&q=get_or_404&type=Code&utf8=%E2%9C%93

NVM, `get_or_404_` is more common.

naved001:
> How about changing it to 声称_absent?

LOL.

> get_or_404_ is more common.

yeah, I saw it in SQLAlchemy too. So let's just go with what Ian suggested initially and close this one.

PR #965, opened 2018-03-01
_assert_absent and friends are having better names now

xuhang57:
Please see: https://github.com/CCI-MOC/hil/issues/544


naved001:
I'll just wait for @zenhack since he opened the original issue about it.

zenhack:
LGTM, merging.

Issue #45, opened 2014-04-15
Configuration should be checked on startup.


zenhack:
Right now if a config file option isn't set properly, the haas will just fail at runtime, when it tries to use that option. We should be validating the configuration on startup.

xuhang57:
congrats on our 4 years old issue 🥇

xuhang57:
here is the thought:

Should we look into automation on configuring HIL with different types of backend through Ansible?

If so, we can configure ansible to validating the configuration and throw proper error messages or even give a verbose indication on how to proceed. 

Thoughts? @zenhack @naved001 @Izhmash

zenhack:
I am antagonistic to offloading this on to the automation tool. Right now setting up HIL manually is something mere mortals can do; I want to keep it that way.

xuhang57:
What we could have is to:

1. Automation on HIL deployment
2. Mere mortals approach: Manually HIL deployment

So this issue is under option 2 and we should discuss on how we can validate the configure file in HIL.

Any suggestions?

zenhack:
One rough design: first pull things from the configparser object into a dictionary, and then use the schema library to validate it. Each extension could provide provide additional schema for its own section.

Izhmash:
So then core HIL and each of the extensions would have one schema that defines everything that can go into the config file?  I can give this one a go, that idea makes sense to me.

zenhack:
@Izhmash, yeah, basically. One way you could make it a bit more strict:

HIL core would define a schema describing the sections it defines itself, something like:

```python
core_schema = {
    'auth': {
        Optional('log_level'): bool,
    },
    'devel': {
        Optional('dry_run'): bool,
    },
    ...
    'extensions': {str: ''},
}
```

Then, on import, each extension, could add it's own partial schema to the core:

```python
# in hil/ext/auth/keystone.py
core_schema[__name__] = {
    'username': str,
    'password': str,
    ...
}
```

Finally, the core would validate against the full schema once all the extensions are loaded and have had a chance to modify it to their liking.

Izhmash:
Ok, that sounds good to me.

zenhack:
Closing, now thatn #966 is merged.

PR #966, opened 2018-03-01
Config precheck

Izhmash:
Addresses #45:

- Config file is now checked when the HIL server is started
- Sections that are not critical to HIL are optional
- Option types are checked as well, such as strings vs bools
- Tests that didn't include all of the required HIL.cfg sections/options have been updated

Note: I am going to start working on testing, but I wanted to get this out for review first since I'll be out next week.


xuhang57:
Officially the first PR which increases the coverage 💟 🍸 🤙

hil/config.py
```diff
@@ -8,12 +8,50 @@
...
+    },
+    'client': {
+        Optional('endpoint'): str,
+    },
+    'database': {
+        'uri': str,
+    },
+    Optional('devel'): {
+        Optional('dry_run'):
+            lambda s: string_is_bool(s),
```
zenhack: The lambda is superfluous here (and elsewhere); you can just do `Optional('dry_run'): string_is_bool`. You'll have to move the definition of that function up though, so it's defined before core_schema.

hil/ext/switches/n3000.py
```diff
@@ -39,13 +44,13 @@ class DellN3000(Switch):
...
-        schema.Schema({
+        Schema({
             'username': basestring,
             'hostname': basestring,
             'password': basestring,
-            'dummy_vlan': schema.And(schema.Use(int),
-                                     lambda v: 0 < v and v <= 4093,
-                                     schema.Use(str)),
+            'dummy_vlan': And(Use(int),
+                              lambda v: 0 < v and v <= 4093,
```
zenhack: While you're at it, this python supports a shorthand, so this can be `0 < v <= 4093`.

Also, is there a reason for 4093? The others are 4096, which makes more sense.

hil/cli.py
```diff
@@ -853,5 +854,7 @@ def main():
             sys.exit('Error: %s\n' % e.message)
         except BadArgumentError as e:
             sys.exit('Error: %s\n' % e.message)
+        except SchemaError as e:
+            sys.exit('Config error: %s\n' % e.message)
```
zenhack: I might be slighly more explicit here; `Error in config file`.

hil/ext/switches/n3000.py
naved001: yeah, the Dell N3000 switch only showed that it's from 1 to 4093. 
```
R5-PA-C01-U39#show vlan id ?

<1-4093>                 Enter VLAN ID.

R5-PA-C01-U39#show vlan id 4094
                           ^
Value is out of range. The valid range is 1 to 4093.
```

zenhack: Weird. Would you mind adding a comment to that effect?

naved001: In [network-drivers](https://github.com/CCI-MOC/hil/blob/master/docs/network-drivers.md#switch-preparation-1) it's mentioned that we enable VLAN 2-4093, but it doesn't mention the reason.  @Izhmash, can you put a comment in the code while you are at it and update the docs to make it more clear?

hil/ext/switches/nexus.py
```diff
@@ -38,13 +44,13 @@ class Nexus(Switch):
...
-        schema.Schema({
+        Schema({
             'username': basestring,
             'hostname': basestring,
             'password': basestring,
-            'dummy_vlan': schema.And(schema.Use(int),
-                                     lambda v: 0 <= v and v <= 4096,
-                                     schema.Use(str)),
+            'dummy_vlan': And(Use(int),
+                              lambda v: 0 <= v and v <= 4096,
```
naved001: while we are talking about vlan ranges, I am sure 0 shouldn't be included in this range. 
```
haas-loaner# show vlan id ?
  <1-3967,4048-4093>  VLAN ID 1-4094 or range(s): 1-5, 10 or 2-5,7-19
>
haas-loaner# show vlan id 0

```

Izhmash:
I believe I addressed the comments above.  @zenhack for more in-depth config checks, I've added checks for malformed web/db urls, a check for the VLAN pool list, and a check for log_level.  Naved and I discussed checking the headnode config options, and we decided they can be left as `str` for now since they will be removed in the future.

I considered checking `log_dir`, however a bad directory crashes HIL before before the config is even loaded. I'll add an issue about that.

Izhmash:
Tests will be on the way as soon as we lock in which config checks should be there.

hil/config.py
```diff
@@ -89,3 +170,4 @@ def setup(filename='hil.cfg'):
     load(filename)
     configure_logging()
     load_extensions()
+    validate_config()
```
naved001: move this before `configure_logging()` and this PR itself can take care of #976 
This [line](https://github.com/CCI-MOC/hil/blob/master/hil/config.py#L62) is probably what's causing trouble.

hil/config.py
```diff
@@ -8,13 +8,85 @@
...
+
+def is_log_level(option):
+    return option.lower() in ['debug', 'info', 'warn', 'warning', 'error',
+                      'critical', 'fatal']
+
+def string_has_vlans(option):
+    """Check if a string is a valid list of VLANs"""
+    for r in option.split(","):
+        r = r.strip().split("-")
+        if not all(s.isdigit() for s in r):
```
zenhack: 1. you can actually just call `isdigit` once:

```python
>>> '423'.isdigit()
True
```

2. Why not actually check the range?

```python
s.isdigit() and 0 < int(s) <= 4096
```

Or, for that matter, just lift the relevant bit of the schema from the dummy_vlan field.

hil/config.py
```diff
@@ -8,13 +8,85 @@
...
 import sys
+from urlparse import urlparse
 
 cfg = ConfigParser.RawConfigParser()
 cfg.optionxform = str
 
 
+def string_is_bool(option):
+    """Check if a string matches ConfigParser's definition of a bool"""
+    return option.lower() in ['true', 'yes', 'on', '1',
```
zenhack: You could be making better use of the library; `Use`, `And`, and `Or` are handy:

```python
return And(Use(option.lower), Or('true', 'yes', 'on', ...))
```

This can be applied in a number of other places.

hil/config.py
```diff
@@ -8,13 +8,85 @@
...
+    """Check if a string is a valid web URL"""
+    url = urlparse(option)
+    if url.scheme == '' or url.netloc == '':
+        return False
+    return True
+
+def string_is_db_uri(option):
+    """Check if a string is a valid DB URI"""
+    url = urlparse(option)
+    if not ('postgres' in url.scheme or 'sqlite' in url.scheme):
```
zenhack: I don't think we need to use `in` here; they should be exactly equal.

Alternatively, you could do `url.scheme in ('postgres', 'sqlite')`.

hil/config.py
zenhack: We should probably check this by catching the IOError and reporting the issue; conceptually doing it here is weird, because the problem is an absent or non-writable directory, not a malformed confg file. Slightly more pragmatically, there's a race condition where if the directory is deleted/permissions changed between the check and the use, you could still get the error.

hil/ext/auth/keystone.py
```diff
@@ -5,14 +5,23 @@
...
+from hil.config import cfg, core_schema
 from hil.model import Project
 from hil import auth, rest
 import logging
 import sys
 
 logger = rest.ContextLogger(logging.getLogger(__name__), {})
 
+core_schema[__name__] = {
+    'auth_url': str,
```
zenhack: Could probably check for a valid URL here.

Would be worth tightening the other fields here as well, where possible.

hil/config.py
Izhmash: @zenhack I agree that conceptually it's strange to do it here as well. For the scope of this PR, I'm thinking about moving the config validation up and then only checking for a malformed path.  The IO check can be done in another small PR.

hil/config.py
Izhmash: I'll check the range here, good for a first safety check

hil/cli.py
```diff
@@ -853,5 +854,7 @@ def main():
             sys.exit('Error: %s\n' % e.message)
         except BadArgumentError as e:
             sys.exit('Error: %s\n' % e.message)
+        except SchemaError as e:
+            sys.exit('Error in config file: %s\n' % e.message)
```
xuhang57: As we discussed during the weekly meeting. Probably we should just return the `e`? Since `e.message` might be `None?

hil/config.py
```diff
@@ -8,13 +8,85 @@
...
+def string_is_bool(option):
+    """Check if a string matches ConfigParser's definition of a bool"""
+    return option.lower() in ['true', 'yes', 'on', '1',
+                              'false', 'no', 'off', '0']
+
+def string_is_web_url(option):
+    """Check if a string is a valid web URL"""
+    url = urlparse(option)
+    if url.scheme == '' or url.netloc == '':
+        return False
```
xuhang57: Curious whether we could just `return url.scheme != '' and url.netloc != '' `.

I know you may wanna return the `False` explicitly?

hil/config.py
Izhmash: For some reason, having this function like this instead just places whatever is in `option` into my validation schema, causing any config string to pass.

Izhmash: It works however if I put the `And(...)` directly into the schema

hil/config.py
Izhmash: Sure, that works fine.

hil/ext/auth/keystone.py
Izhmash: I believe the only field I can check here is the `auth_url`.  I considered `auth_protocol` at least, but the authentication plugins available can differ depending on what's installed.

hil/config.py
zenhack: Ah, you're right, I was thinking you were checking each character.

hil/config.py
zenhack: Sounds good.

hil/config.py
Izhmash: @zenhack how partial are you to using more of these library functions? I've been hacking at it for a while, and returning in this style causes the validation to not happen. From debugging it looks like there's an issue with returning the validation from another function, since it does work when hard-coding into the Schema dictionary.

To use more of the library functions I could instead remove some of the functions (like `string_is_bool`) and write the validation code right into the schema.

Izhmash:
Tests are in progress now.

hil/config.py
```diff
@@ -8,13 +8,94 @@
...
+    url = urlparse(option)
+    return not (url.scheme == '' or url.netloc == '')
+
+
+def string_is_db_uri(option):
+    """Check if a string is a valid DB URI"""
+    url = urlparse(option)
+    if url.scheme not in ('postgresql', 'sqlite'):
+        return False
+    if url.path == '':
```
naved001: why not `or` this with the previous statement?
```python
if url.scheme not in ('postgresql', 'sqlite') or url.path == '':
```
or maybe directly return it.
```python
def string_is_db_uri(option):
    return url.scheme in ('postgresql', 'sqlite') and url.path != ''
```

Additionally, `'sqlite://'` is valid and gives an [in-memory database](http://docs.sqlalchemy.org/en/latest/dialects/sqlite.html#connect-strings), so `url.path` will be `''`.  But I don't mind making it mandatory for developers to specify `:memory:` in their tests (and that only makes sense, if your config checker is called before our tests that use in memory db).

Also, like in the case of logging directory where we want to make sure that the directory exists, we might wanna check that the database specified in the config file is actually reachable. This could just be an additional check if folks agree.
A simple function like this could check it:
```python
def is_db_responsive(option):
    try:
        db = create_engine(option)
        db.connect()
        db.dispose()
        return True
    except:
        return False
```

zenhack: Might be worth doing a check on start-up, but I think it should be separate from validating the config file.

Izhmash: @naved001 I agree that we should check for an existing directory, but I also agree with @zenhack that we should do it outside of the config validation.  The config validation should really just be checking config strings.  This would fix #976.

I also see what you mean about the in-mem DBs too. I'm just going to check the scheme then, no need to limit users that much.

hil/config.py
Izhmash: Oh, that's what I was missing. That definitely works, but it comes down now to how ugly the error messages look:

Current code error example:
```
Error in config file: string_is_bool('Frue') should evaluate to True
```

New suggestion error example: 
```
Error in config file: Or('true', 'yes', 'on', '1', 'false', 'no', 'off', '0') did not validate 'Frue'
'0' does not match 'Frue'
```
I'm not 100% sure why the errors look different now, perhaps because it's a separate validation happening before the main dictionary validation.  I think the 2nd error option might expose a little too much code to the user.

hil/config.py
Izhmash: I can just do a `os.path.isdir("/home/el")`

Izhmash:
@naved001 I addressed your questions in this commit I believe.

hil/config.py
zenhack: I could go either way re: the error message. They're both exposing implementation detail a bit, but the later has the virtue that it actually gives the user a good hint as to what the correct settnigs are.

Izhmash:
I've updated the functions to better use the schema library.  I tried to not use lambdas, but I went with them since it seems to be a standard in the schema docs.

I also fixed up the `IOError` checking. In testing, `EACCES` was raised rather than `EPERM` for directory permissions.  In case of an unexpected error, I re-raise it so the CLI can deal with it.

tests/unit/config.py
```diff
@@ -25,3 +27,161 @@ def test_load_extension():
...
+    opts = ['true', 'yes', 'on', '1', 'false', 'no', 'off', '0']
+    assert all(config.string_is_bool(s) for s in opts)
+
+
+def test_bool_validation_bad_cases():
+    """Test invalid bools."""
+    opts = ['frue', 'nes', 'yOn', '2', 'Salse', 'Go', 'bff', '3']
+    for s in opts:
+        with pytest.raises(SchemaError):
+            config.string_is_bool(opts)
```
zenhack: Think this should `config.string_is_bool(s)`. Yay for linters catching real bugs.

Izhmash: Ah so that's where it was! I was having a hard time figuring that out last week,

Izhmash:
Everything should be all set now.

PR #967, opened 2018-03-01
Bug Fix: Omit coverage test in virtualenv

xuhang57:
so all files within the `virtaulenv` should not be considered when we are running the coverage test.


zenhack:
NaN% -- hah.

Issue #700, opened 2017-01-12
Update documentation to require python 2.7


henn:
2.6 is ancient, and AFAIK we have no CentOS 6 machines any more.

Any objections?

It would be good to start going to python 3, but let's go one step at a time.

zenhack:
No objections here.

For posterity: the conversation in #680 provides some context; basically @kylehogan submitted code that requires 2.7+ and I mentioned that we should be explicit about what to support.

re: the python 3 thing, another intermediate step would just be writing portable code, which isn't actually that hard. The `six` library smooths over most of the pain points. And we can add 3.x to our travis matrix; I believe there's even a "fail ok" option we could use in the interim, while stuff isn't fully portable still. I don't really care about python 3 myself, but I'll happily review PRs to improve portability.

shwsun:
I agree with removing obsolete things and improving portability

henn:
The [README](https://github.com/CCI-MOC/hil/blob/master/README.rst) says that we require python 2.7, so that portion is solved, but I like the thoughts from @zenhack for slowly transitioning to python 3, so let's make this issue about that.

xuhang57:
`six` ? 

python 2to3?

zenhack:
@xuhang57, the trouble with 2to3 is that it leaves your code in a state where it *only* works with python 3, which is not a leap I'm willing to make just yet; there are a lot of systems out there where python 2 is still the default.

re six: what is your question exactly?

Like I mentioned above, step 1 is probably to add python 3 to the build matrix with a "fail ok" setting, so we can see what's working and what's not. From there we can start fixing stuff.

xuhang57:
Oh, sorry, I was just writing all packages I know that might be helpful for translating python2 to 3.

And what you said about step 1 makes sense to me. But should we create a new branch or even a repo to get started testing? I feel like it makes sense to give it a try and see what we need for the process as early as we could.

zenhack:
@xuhang57, I think someone should submit a PR that adds an item to our build matrix to run the tests against python 3, and flags them allowed to fail. We're doing this for obmd for the (unreleased) master branch of Go, see:

https://github.com/CCI-MOC/obmd/blob/master/.travis.yml#L9-L10

We'll review and merge that PR, which will no doubt have a bunch of failures on python 3. From there we can start submitting PRs fixing the various tests.

zenhack:
Just pushed this to a branch:

[...]

PR #968, opened 2018-03-05
Add Python 3 to the build matrix.

zenhack:
There is predicatbly a failure, but I suspect it's something actually related to Python 3 now, so I'm submitting this as a pr.

Relates to: https://github.com/CCI-MOC/hil/issues/700

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: python
 python:
   - "2.7"
+  - "3.6"
+matrix:
+  allow_failures:
+    - python: "3.6"
 
 virtualenv:
   system_site_packages: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,10 @@
 language: python
 python:
   - "2.7"
-  - "3.6"
+  - "3.4"
 matrix:
   allow_failures:
-    - python: "3.6"
+    - python: "3.4"
 
 virtualenv:
   system_site_packages: true


xuhang57:
I might not know what I am talking about. But fwiw, normally, python 3 is using pip3 rather than pip2 to manage packages so does that mean we need to update that in travis for our python 3 build?

naved001:
Didn't make a difference when I was using pip3 vs pip for installing in my python3 virtualenv. it just couldn't import `importlib._bootstrap` in both cases (the current failure in our travis build).

naved001:
Also, in a virtualenv with python3 as default, pip seems to refer to pip3.

```
(.venv) naved:~/hil/.venv/bin$ python --version
Python 3.5.2
(.venv) naved:~/hil/.venv/bin$ pip --version
pip 9.0.1 from /home/naved/hil/.venv/lib/python3.5/site-packages (python 3.5)
(.venv) naved:~/hil/.venv/bin$ pip3 --version
pip 9.0.1 from /home/naved/hil/.venv/lib/python3.5/site-packages (python 3.5)
(.venv) naved:~/hil/.venv/bin$ pip2 --version
pip 9.0.1 from /usr/local/lib/python2.7/dist-packages (python 2.7)
```

xuhang57:
Is there a way to let travis to show all builds on this page?

.travis.yml
```diff
@@ -1,6 +1,10 @@
 language: python
 python:
   - "2.7"
+  - "3.4"
+matrix:
+  allow_failures:
+    - python: "3.4"
```
xuhang57: I know you are using `3.4` because `3.6` is failing. But since `3.4` is failing as well, should we just use `3.6`?

xuhang57:
https://devguide.python.org/#status-of-python-branches

Given what it says in the official guide, we should start with `3.6` since `3.4` will be deprecated even before `2.7`

zenhack:
Quoting Lucas H. Xu (2018-03-05 16:38:23)

>    I know you are using 3.4 because 3.6 is failing. But since 3.4 is
>    failing as well, should we just use 3.6?

We expect failures from any 3.x variant, but the 3.6 failures look like
they're not even getting as far as *installing* python 3.6. I suspect
the current failure is an actual portability bug, which is a bit better.

xuhang57:
@zenhack Thanks for explaining it. Should we install python 3.6 in the Travis container then?

My point is more like, instead of supporting `3.4` now, which it will be deprecated before 2020, we should look into 3.6 which we don't have to bother ourselves to do a version bump later since we are just about to set thing ups. Admittedly, I am not against using `3.4`

zenhack:
The 3.x series isn't terribly variable, and 3.4 takes less set-up now.
By the time it's actually unsupported by the PSF, we can probably
upgrade to another version in travis just by bumping the number in our
config, whereas right now it's a (small) bit of work. If you want to do
that work, I'm happy to accept a patch.

Quoting Lucas H. Xu (2018-03-05 17:15:37)

xuhang57:
I suggest we could merge this and close the issue #700 . Maybe open a new issue that is about supporting python 3. Or we could just keep using the old issue. 

Thanks Ian!

PR #969, opened 2018-03-07
CLI using Click

naved001:
* Issue #431 
* This is ~~a rough draft/almost~~ done (~~waiting on the switch_register client lib call~~).
* ~~So far~~ I have added all commands ~~for node (adding more in new commits)~~

* My plan is to put all commands in this PR. That will make the PR huge, but since most of the pattern will be repeated so it might not be as difficult to review. 

* ~~In another PR I'll move the `serve` and `serve_networks` command to `hil-admin` and the entry_points mechanism~~. Edit: Moved `serve_networks`, `serve`, and `create_admin_user` to `hil-admin`, and also updated the entry point mechanism.

* Finally, right now we just dump the output of the api calls (see show_node), in some other PR we'll clean that output to make it human readable.

Collaborative effort with @xuhang57


xuhang57:
Thanks and just a quick question. Should we call it `newcli`? I thought we are replacing the old `cli`.

naved001:
you are right, ill delete the old file and rename this to cli.

naved001:
looking at the diff is terrible right now because the whole file has changed.

hil/cli.py
```diff
@@ -1,81 +1,27 @@
...
-logger = logging.getLogger(__name__)
-command_dict = {}
-usage_dict = {}
-MIN_PORT_NUMBER = 1
-MAX_PORT_NUMBER = 2**16 - 1
 VERSION = pkg_resources.require('hil')[0].version
 
 # An instance of HTTPClient, which will be used to make the request.
 http_client = None
 C = None
```
xuhang57: I would say we should try out best to avoid the "simple-letter-variable-name" situation since you are currently working on building a new CLI. Unless there is a convention on global variable name on Python that I am not aware of at the moment ;)

hil/cli.py
```diff
@@ -151,94 +97,54 @@ def setup_http_client():
...
-#        `data` - the body of the request (for PUT, POST and DELETE)
-#        `params` - query parameters (for GET)
+@cli.group()
+def project():
+    """Commands related to project"""
 
-def do_put(url, data={}):
-    """do a put request and check the response."""
-    check_status_code(http_client.request('PUT', url, data=json.dumps(data)))
+@cli.group()
```
xuhang57: it is kinda hard to see so please forgive me if I said something crazy. 

Are we having two `project` under `cli.group()` here?

hil/cli.py
```diff
@@ -151,94 +97,54 @@ def setup_http_client():
     C = Client(ep, http_client)
```
xuhang57: I would say we should try out best to avoid the "simple-letter-variable-name" situation since you are currently working on building a new CLI. Unless there is a convention on global variable name on Python that I am not aware of at the moment ;)

hil/cli.py
```diff
@@ -151,94 +97,54 @@ def setup_http_client():
...
-
-@cmd
-def version():
-    """Check hil version"""
-    sys.stdout.write("HIL version: %s\n" % VERSION)
-
-
-@cmd
+@cli.command(name='serve', help='Run a development HIL server')
+@click.argument('port', type=click.Choice(str(i) for i in range(1,2**16)),
```
xuhang57: Since here we are enforcing a range of port numbers, does it make sense that we are allowing the users to use port numbers under `1000`?

hil/cli.py
```diff
@@ -247,17 +153,17 @@ def serve(port):
...
-    from hil import api, rest
+    from hil import api
     server.init()
     migrations.check_db_schema()
     server.stop_orphan_consoles()
     rest.serve(port, debug=debug)
 
 
-@cmd
+@cli.command(name='serve_networks', help='Run the networking server')
```
xuhang57: do we want to enforce a range of port numbers here as well?

hil/cli.py
```diff
@@ -288,570 +194,194 @@ def serve_networks():
...
 
-@cmd
-def node_delete(node):
-    """Delete <node>"""
-    C.node.delete(node)
+###########################
+# node connect/detach network implementations. Could also be expanded to do
+# node connect/detach project if need be.
 
+@node.group(name='connect', help='Connect node to an <object>')
```
xuhang57: Thinking we should make the `help` message consistent with each other. so maybe it is a good idea to change it to:

`Connect a node to an <object>`

hil/cli.py
```diff
@@ -288,570 +194,194 @@ def serve_networks():
...
+def node_connect_network(node, network, nic, channel):
+    """Connect <node> to <network> on given <nic> and <channel>"""
+    print C.node.connect_network(node, nic, network, channel)
 
 
-@cmd
-def node_power_off(node):
-    """Power off <node>"""
-    C.node.power_off(node)
+@node.group(name='detach', help='Detach node from an <object>')
```
xuhang57: same as above, maybe Detach `a` node

hil/cli.py
```diff
@@ -288,570 +194,194 @@ def serve_networks():
...
+def node_detach():
+    """Connect node an <object>"""
 
 
-@cmd
-def node_set_bootdev(node, dev):
-    """
-    Sets <node> to boot from <dev> persistently
+@node_detach.command(name='network', help="Detach node from a network")
```
xuhang57: here too

hil/cli.py
```diff
@@ -288,570 +194,194 @@ def serve_networks():
...
-
-
-@cmd
-def list_project_nodes(project):
-    """List all nodes attached to a <project>"""
-    q = C.project.nodes_in(project)
-    sys.stdout.write('Nodes allocated to %s:  ' % project + " ".join(q) + '\n')
-
+###########################
+# I am doing the dash thing with the console commands. Let me know what's
```
xuhang57: +1 from me. Unless a better way of handling this. We should use a `-` to connect a two-word action.

hil/cli.py
naved001: cool, will do that. I just copy pasted this from the old cli for now. But yeah, might as well clean it up while we are at it.

hil/cli.py
naved001: no, there's just one? I defined all the top level subcommands on top.

hil/cli.py
xuhang57: exactly. 👍

hil/cli.py
xuhang57: lol, is this confused face new? I like it ahha

hil/cli.py
naved001: this process doesn't listen on a port. It's a simple python program reading from the database.

hil/cli.py
naved001: see how I did nic.

xuhang57:
To summarize, 

- `hil node connect network <node> <network> <nic> <channel>`
vs.
- `hil node network connect <node> <network> <nic> <channel>`

I am fine with either way. We should make a decision and proceed.

Just one minor thing bugs me that:
`hil project node xxx` 
`hil node project xxx`
could be a bit confused on some occasions.

hil/cli.py
naved001: I wouldn't worry about this too much because this is meant for development, so I would let developers choose whatever they want.

hil/cli.py
```diff
@@ -288,570 +190,352 @@ def serve_networks():
             pass
         sleep(sleep_time)
 
+#################################
+### NODE SUBCOMMANDS GO HERE! ###
+#################################
```
zenhack: Big comments like these that basically make sections in the file suggest to me that maybe we should just split up the file into multiple modules.

hil/cli.py
```diff
@@ -288,570 +190,352 @@ def serve_networks():
...
-@cmd
-def port_revert(switch, port):
-    """Detach a <port> on a <switch> from all attached networks."""
-    print C.port.port_revert(switch, port)
+####################################
+### PROJECT SUBCOMMANDS GO HERE! ###
+####################################
 
+###########################
+# I am invoking the project list and project list <something> commands here
```
zenhack: I kinda don't like this approach. I'd prefer to keep the groups and the commands separate. It feelds weird to have list the networks of a given project to be a sub-command of listing all the projects -- they don't feel related enough.

naved001: this is what I think we can do:

`hil project list`  will list all projects
then we have the `hil project network` group
`hil project network list` will list all project networks
`hil project network grant-access` will grant project access to network
`hil project network revoke-access` will do the opposite

we can use the words add/remove in place of grant-access/revoke-access respectively.

Now while we are thinking about this. The way I have implemented connecting a project to a node is like this:
There's a group for project_connect, and then there's only thing that can connect to it i.e. nodes. so it reads like `hil project connect node`. Same thing for detach. This is inconsistent with what I am proposing. So we can change that to:

`hil project node list`
`hil project node add`
`hil project node remove`
and our network commands will read like
`hil project network list`
`hil project network add` (to grant access)
`hil project network delete` (to revoke-access)
`hil network attachments list` (note: hil network attachments then becomes a group).

hil/cli.py
naved001: good point. I'll make a folder for cli and make multiple modules. Will probably help with reviewing too.

hil/cli.py
```diff
@@ -288,570 +190,352 @@ def serve_networks():
...
-@cmd
-def list_active_extensions():
-    """List active extensions by type. """
-    all_extensions = C.extensions.list_active()
-    if not all_extensions:
-        print "No active extensions"
-    else:
-        for ext in all_extensions:
-            print ext
+@network.group(name='grant-access')
```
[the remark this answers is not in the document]
naved001: yeah, the only reason I made this group was just so that my command reads like `hil project grant-access network`

naved001:
okay, now that I have split it up into different modules, I am going to address reviewers comments.

hil/cli.py
xuhang57: I don't have a strong opinion on this. I think this would work well. But I do believe that once we have made a decision, we should write something like an HIL community CLI guideline. I don't mind helping write or we could have it along with this PR 👍 @naved001 @zenhack @jeremy @Izhmash

naved001:
I have added all commands now except for `show_network_action`. Any ideas how should it read?

hil/cli.py
zenhack: Sounds like a plan.

zenhack:
Why not just `hil network-action show`?

naved001:
> hil network-action show

cool, will do that. I'll put it in `cli/misc.py`

xuhang57:
> I have added all commands now except for show_network_action. Any ideas how should it read?

My thought on that is, this should be like `hil version`, which is a standalone command. 

So we may group them together? Or we should put in the `hil network`?

for example, `hil network show-action`?

xuhang57:
`hil network-action show` works too.

so if we have more than one more for objects or actions, we should use `-` to connect them.

naved001:
no, networking-action command has little to do with the nework object. The version command would go in `misc.py` but Click had a nice way of implementing it by putting the decorator on top of my cli group.

naved001:
@zenhack @xuhang57 I think this is ready for review.

Note:

The command for granting projects access to a network is in the network module, not project because that's how our api is network/<network>/project/<project>. If you guys agree, I can move it to project module.

naved001:
I am just tempted to move the `serve`, `serve_networks` and `create_admin_user` commands to `hil-admin` script in this PR.

naved001:
`hil-admin runserver -p PORT` already exists, then why did we implement `hil serve PORT` ?

hil/cli/cli.py
```diff
@@ -0,0 +1,35 @@
...
+
+def main():
+    """CLI entry point"""
+    ensure_not_root()
+    try:
+        cli()
+    except FailedAPICallException as e:
+        sys.exit('Error: %s\n' % e.message)
+    except Exception as e:
+        sys.exit('Unexpected error: %s\n' % e.message)
```
naved001: delete this.

hil/cli/network.py
```diff
@@ -0,0 +1,81 @@
...
+@click.group()
+def network():
+    """Commands related to network"""
+
+
+@network.command(name='create', short_help='Create a new network')
+@click.argument('network')
+@click.argument('owner')
+@click.option('--access')
+@click.option('--net_id')
```
naved001: @jeremyfreudberg thinks that the optional arguments should be shown without typing`--help` explicitly.

xuhang57:
if `hil help` is more intuitive than `hil --help`, we could figure out a way that maps `hil help` to `hil --help`.

For reference, `git help` and `git --help` both bring up the usage of `git` CLI.

Also, if we want to make the hil CLI more intuitive, we should add more `docstring`, which gives the users more information about the CLI and we should have a good guide for the users to read. And that should suffice.

jeremyfreudberg:
@xuhang57 

for example, something like `hil help network create` and `hil network create --help` feel nearly equally valid, and I too have seen other CLIs which support both. (But keep in mind we can't actually use the string `help` as a positional argument in a command... for example `hil network create help` would be ambiguous... is `help` the name of some parameter (like network name) or the actual call for help.)

xuhang57:
sure, and I tried it out using `git`:

```
(.venv) ➜ ubuntu@hil-dev  ~/naved/hil git:(click-cli) git fetch help
fatal: 'help' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

and using `hil`:

```
(.venv) ➜ ubuntu@hil-dev  ~/naved/hil git:(click-cli) hil network help
Usage: hil network [OPTIONS] COMMAND [ARGS]...

Error: No such command "help".
```
I would say `click` handles that ok?

xuhang57:
@SahilTikale as Sahil points out, we should make the CLI more intuitive. But I don't think we have a critical disadvantage of using `click` and we should be working together to make the CLI more institutive with using `click`.

naved001:
whoever reviews this, please try running it too to get the feel of the CLI.

hil/cli/network.py
zenhack: Let's standardize on hypinating words everywhere; right now we've got some places that are hyphinated and some that have underscores.

examples/dbinit.py
```diff
@@ -35,14 +35,14 @@ def hil(*args):
...
     ipmi_ip = "10.0.0." + str(node + 1)
 
     nic_port = "gi1/0/%d" % (node)
     nic_name = 'nic1'
-    hil('node_register', node, "mock", ipmi_ip, ipmi_user, ipmi_pass)
-    hil('node_register_nic', node, nic_name, 'FillThisInLater')
-    hil('port_register', switch, nic_port)
-    hil('port_connect_nic', switch, nic_port, node, nic_name)
+    hil('node register', node, "mock", ipmi_ip, ipmi_user, ipmi_pass)
+    hil('node nic register', node, nic_name, 'FillThisInLater')
```
zenhack: These should be separate arguments now:

```python
hil('node', 'nic', 'register', node, nic_name, 'FillThisInLater')
```

hil/cli/misc.py
```diff
@@ -0,0 +1,109 @@
+"""Miscellaneous commands go here"""
+import click
+from hil.cli.client_setup import setup_http_client
+
+C = setup_http_client()
+
+
+@click.command(name='serve', help='Run a development HIL server')
+@click.argument('port', type=click.IntRange(0, 2**16-1),
```
zenhack: I think it would be better for this to be an option `--port`, rather than a positional argument.

hil/cli/misc.py
```diff
@@ -0,0 +1,109 @@
+"""Miscellaneous commands go here"""
+import click
+from hil.cli.client_setup import setup_http_client
+
+C = setup_http_client()
+
+
+@click.command(name='serve', help='Run a development HIL server')
```
zenhack: While we're at it, why not rename this to `dev-server`? I think it might be a bit more clear.

hil/cli/network.py
```diff
@@ -0,0 +1,81 @@
...
+
+    If <project> is `None`, lists all attachments for <network>
+    """
+    print C.network.list_network_attachments(network, project)
+
+
+@network_project.command(name='add')
+@click.argument('network')
+@click.argument('project')
+def network_grant_project_access(project, network):
```
zenhack: I really think this would scan better as e.g. `hil network grant-access --project manhattan --network foo`. There's not really a natural order to the arguments here, and having project be a sub-category of network for network & project calls feels clumsy to me.

naved001: so then, `grant-access` becomes a command with project and network as optional arguments?

I was thinking of moving this call to `project` subcommands. Just like how we have, `hil project node add/remove` we can do `hil project network add/remove` if that makes sense.

I am fine doing either.

naved001:
> It feels really awkward to have the server-side commands in here. I think at some point (maybe now, but I could be convinced either way)

I'll do that in the next commit along with the changes you suggested for those commands.

hil/cli/network.py
zenhack: Hm, I was hoping there was a way to have flags that were mandatory, but it doesn't seem like there is. (it obviously doesn't make sense for project/network to be optional for this call). We could keep them as positional, but still do `hil network grant-access ...`, which is nice in that it mirrors the structure of the client library.

hil/cli/network.py
xuhang57: +1 on `net-id`

examples/dbinit.py
xuhang57: maybe we should do `nic-name`

zenhack:
Quoting Lucas H. Xu (2018-03-14 22:11:36)

>    maybe we should do nic-name

Can't, because it's a python variable name. Fine though, as it's not
part of the CLI's output.

hil/cli/network.py
naved001: okay. I had done that in a previous commit and we decided to change because we had to make `grant-access` and `revoke-access` a subcommand under `network`. Just want to make sure that this is what you are suggesting before I do that again.
https://github.com/CCI-MOC/hil/pull/969/commits/5595dfabf6feb1512df722a09a40004689a2f153?diff=split

hil/commands/admin.py
```diff
@@ -1,16 +1,96 @@
...
+    while True:
+        # Empty the journal until it's empty; then delay so we don't tight
+        # loop.
+        while deferred.apply_networking():
+            pass
+        sleep(sleep_time)
+
+
+@manager.option('-p', '--port', type=IntRange(0, 2**16-1), default=5000)
+def dev_server(port):
```
Izhmash: Having `dev_server` be an action that is not a verb feels a bit weird to me.  Perhaps it could be `run_dev_server`?

hil/cli/network.py
```diff
@@ -0,0 +1,81 @@
...
+
+@click.group()
+def network():
+    """Commands related to network"""
+
+
+@network.command(name='create', short_help='Create a new network')
+@click.argument('network')
+@click.argument('owner')
+@click.option('--access')
```
Izhmash: Maybe here we could do something like 
```
@click.option('--access', help='Project nodes that can access this network.')

That way, `TEXT` would also have a helpful prompt.
```
Found this from a thread here https://stackoverflow.com/questions/39359420/type-and-default-input-value-of-a-click-option-in-help-option

hil/commands/admin.py
naved001: It's gonna be exposed to developers and not end users, which is why I didn't think much of it. But yeah, I'll change that.

hil/cli/network.py
zenhack: If you're talking about this thread:

https://github.com/CCI-MOC/hil/pull/969#discussion_r173029097

The concern there was different, as you had a group for one command.

naved001:
Note to myself: Need to update the docs too since a lot of commands would have changed.

xuhang57:
@naved001 It's up to you whether to have a CLI guide along with the docs or having two separate files. 👍

naved001:
@xuhang57 I mean in the installation docs for developers and some place else we have a bunch of hil commands. That would need to be updated.

naved001:
@zenhack @Izhmash @xuhang57 
I think I have addressed your comments, it's ready for another pass.

hil/commands/admin.py
```diff
@@ -1,16 +1,96 @@
...
+    while True:
+        # Empty the journal until it's empty; then delay so we don't tight
+        # loop.
+        while deferred.apply_networking():
+            pass
+        sleep(sleep_time)
+
+
+@manager.option('-p', '--port', type=IntRange(0, 2**16-1), default=5000)
+def run_dev_server(port):
```
Izhmash: Are we still making `-` our default delimiter for commands? If so, we might as well remove the underscores.

hil/cli/network.py
```diff
@@ -0,0 +1,80 @@
...
+def network():
+    """Commands related to network"""
+    global C
+    C = setup_http_client()
+
+
+@network.command(name='create', short_help='Create a new network')
+@click.argument('network')
+@click.argument('owner')
+@click.option('--access', help='Projects that can access this network. '
```
Izhmash: The `help` makes it look a lot better :+1:

hil/commands/admin.py
zenhack: Again, this is a Python variable name.

hil/cli/network.py
```diff
@@ -0,0 +1,80 @@
...
+from hil.cli.client_setup import setup_http_client
+
+C = None
+
+
+@click.group()
+def network():
+    """Commands related to network"""
+    global C
+    C = setup_http_client()
```
zenhack: Maybe move this logic to cli.main (and centralize the C variable)?

hil/cli/switch.py
```diff
@@ -0,0 +1,48 @@
...
+@click.argument('switch')
+@click.argument('subtype')
+@click.argument('args', nargs=-1)
+def switch_register(switch, subtype, args):
+    """Register a switch with name <switch> and <subtype> <args>
+    eg. hil switch_register mock03 mock mockhost01 mockuser01 mockpass01
+    """
+    # hopefullt the switch_register client call will be done before the cli.the
+    # args is a tuple that can have unlimited number of arguments.
+    pass
```
zenhack: 1. You don't need `pass` if you've got a docstring.
2. The comment has a couple typos.
3. I can't figure out what the comment even means; could you clarify it?

hil/commands/admin.py
```diff
@@ -1,16 +1,96 @@
...
+    This must be run on the HIL API server, with access to hil.cfg and the
+    database. It will create a user named <username> with password
+    <password>, who will have administrator privileges.
+
+    This command should only be used for bootstrapping the system; once you
+    have an initial admin, you can (and should) create additional users via
+    the API.
+    """
+    if not config.cfg.has_option('extensions', 'hil.ext.auth.database'):
+        sys.exit("'make_inital_admin' is only valid with the database auth"
```
zenhack: The name of the command is wrong here. I know it wasn't right before, but would you fix it while you're in there?

hil/commands/admin.py
zenhack: Ah, right, this also ends up being the name of the command. Yeah, let's override this if we can.

Izhmash:
So I just learned that click automatically supports tab auto-completion.  Do we want to add some documentation on how to allow it? All you need to do for HIL is add `eval "$(_HIL_COMPLETE=source hil)"` to `.bashrc`

hil/cli/node.py
```diff
@@ -0,0 +1,182 @@
...
+        sys.stdout.write('All nodes %s\t:    %s\n' % (len(q), " ".join(q)))
+    else:
+        sys.stdout.write('Free nodes %s\t:   %s\n' % (len(q), " ".join(q)))
+
+
+@node.command(name='show')
+@click.argument('node')
+def node_show(node):
+    """Show node information"""
+    q = C.node.show(node)
```
xuhang57: q here works fine since it only exists within this scope. idk. just something thoughts

hil/cli/switch.py
xuhang57: once `switch_register` is merged, I believe Naved will clean this up and finish this function accordingly.

hil/commands/admin.py
```diff
@@ -1,16 +1,96 @@
...
+import logging
+from click import IntRange
+
 manager = Manager(app)
+
 manager.add_command('db', db.command)
 manager.add_command('migrate-ipmi-info', MigrateIpmiInfo())
 
 
+@manager.command
```
xuhang57: could do:
`@manager.command(name='serve-networks')`

hil/commands/admin.py
```diff
@@ -1,16 +1,96 @@
...
+    # (via `rest_call`), though we don't use it directly:
+    # pylint: disable=unused-variable
+    from hil import api
+    server.init()
+    migrations.check_db_schema()
+    server.stop_orphan_consoles()
+    rest.serve(port, debug=debug)
+
+
+@manager.command
```
xuhang57: @manager.command(name='create-admin-user')

hil/commands/admin.py
naved001: I tried that too, but didn't work. Apparently I have to put it in a class which inherits from `flask_script.Command` and then I can do it. Will push it soon.

hil/cli/node.py
```diff
@@ -0,0 +1,182 @@
...
+    global C
+    C = setup_http_client()
+
+
+@node.command(name='list')
+@click.argument('free', type=click.Choice(['free', 'all']), required=True)
+def nodes_list(free):
+    """List all nodes or free nodes"""
+    q = C.node.list(free)
+    if free == 'all':
```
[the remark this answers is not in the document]
Izhmash: If I run the code without typing free or all, I get this `Error: Missing argument "free".  Choose from free, all.`

Izhmash:
#950 is merged now, so it's no longer blocking `switch_register`

hil/cli/node.py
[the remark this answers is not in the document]
naved001: yeah, I made it a positional argument with a required value.I could probably make it an option with a required value too or a default. Whatever you guys think is more user friendly. Maybe I should call the `free` variable something better, any suggestions?

hil/cli/misc.py
```diff
@@ -0,0 +1,19 @@
+"""Miscellaneous commands go here"""
+import click
+from hil.cli.client_setup import setup_http_client
+
+C = None
```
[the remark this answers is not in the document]
naved001: yeah, I'll get to it when I am done with the other major changes.

hil/cli/switch.py
naved001: yeah, I just left a placeholder with a confusing comment; my bad. now I'll update it to use the client library.

hil/cli/node.py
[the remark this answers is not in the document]
Izhmash: I def agree it should be required, @xuhang57 was your comment suggesting that it isn't here?

Hmm a few ideas: "selection", "pool", "group"

hil/commands/admin.py
naved001: fo sho

hil/cli/node.py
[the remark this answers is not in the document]
naved001: I like `pool`. Yeah, it's required and there's no default behaviour for that argument.

hil/commands/admin.py
naved001: done

naved001:
@jeremyfreudberg I couldn't figure out how to always show optional arguments in usage (without typing --help). One alternative is to make those 2 optional arguments as positional arguments that aren't required. 
```
(.venv) naved:~/click/hil$hil network create
Usage: hil network create [OPTIONS] NETWORK OWNER [ACCESS] [NET_ID]
```
so if they want to skip access, they can type `''` for that. I think this is how the current cli handles this.

I think this is the only call that has 2 optional arguments.

naved001:
@zenhack @xuhang57 @Izhmash addressed all your comments. Still need to add file for tab completion. And now it's passing travis checks too.

hil/cli/network.py
naved001: I initialize a client object in `client_setup.py` and then the modules import it. Does that make it better? 
Since I import those modules in `cli.py`, by making the modules import client object from `cli.py` I got worried that I may run into issues with circular imports. Wasn't sure about it so I took the safer route.

xuhang57:
testing CLI is difficult. I don't know how we should approach this.

hil/cli/network.py
zenhack: Fair enough. You could do what we've done with `hil.flaskapp`, but I don't feel strongly.

naved001:
@zenhack @xuhang57 

I have added some integration tests for the CLI. It's just one function doing a bunch of things.
* I could test all the remaining cli commands to it, if it doesn't feel like an overkill.
* Split that big function into multiple ones, but that would require me to use different objects in different functions because they could be run in parallel.

I'll wait for reviewers before I spend more time on that.

naved001:
I feel like I should still include the calls for headnode functionality since it is still a part of the codebase.

naved001:
@zenhack 
could you review this? Thanks!

Izhmash:
I'm happy with how this is now. After Travis is all set it'll get my +1

naved001:
just noticed that in `docs/USING.rst` the curl commands with json body need to set the content-type to application/json which can be specified by `-H "Content-Type: application/json"`. Don't know if I should include it in this PR since I am touching the file here, but then this PR is big as it is. I'll do whatever the reviewers are comfortable with. LMK.

xuhang57:
I would recommend you just update that. It is CLI related so.

docs/INSTALL.rst
```diff
@@ -503,11 +503,18 @@ Create a file ``client_env`` with following entries::
 To get started with HIL from your home dir do the following::
 
     $ source client_env
-    $ hil list_nodes all
+    $ hil list nodes all
 
 If you get an empty list ``[]`` as output then congratulations !!
 At this point, you should have a functional HIL service running!
 
+To enable tab completion, copy this file `scripts/hil-complete.sh` to some location
```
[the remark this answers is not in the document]
naved001: done

tests/integration/cli.py
```diff
@@ -8,32 +9,87 @@
...
+SWITCH2 = 'mock-switch2'
+PORT = 'gi1/0/1'
+
+
+def hil(*args):
+    """Convenience function that calls the hil command line tool with
+    the given arguments.
+    """
+    args = " ".join(args)
+    args = args.split(" ")
```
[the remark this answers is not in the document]
naved001: Just for the convenience, so that I could quickly type the commands in the tests.
`hil('node metadata add', NODE1, 'metadata-label metadata-value')` vs `hil('node', 'metadata', 'add', NODE1, 'metadata-label', 'metadata-value')`. If you think it's not the right way to do it, I don't mind changing it.

zenhack:
Quoting Naved Ansari (2018-04-02 09:06:56)

>    Just for the convenience, so that I could quickly type the commands in
>    the tests.
>    hil('node metadata add', NODE1, 'metadata-label metadata-value') vs
>    hil('node', 'metadata', 'add', NODE1, 'metadata-label',
>    'metadata-value'). If you think it's not the right way to do it, I
>    don't mind changing it.

Yeah, I think it would be better to stick to one command argument per
function argument.

naved001:
> I think it would be better to stick to one command argument per function argument.

@zenhack okay, changed it.

zenhack:
LGTM, merged.

Issue #946, opened 2018-01-30
Enforce unique contraints on object names/labels in the database.


naved001:
We use functions like [_assert_absent](https://github.com/CCI-MOC/hil/blob/master/hil/api.py#L1273) to enforce this in the API.
* I think we should put this constraint on the database itself, because one could, when writing tests, directly add objects to the database and could lead to mistakes if the devs aren't careful enough. Example: [This method](https://github.com/CCI-MOC/hil/blob/master/tests/unit/deferred.py#L168) in tests for the networking server adds a new node with the same name every time a new nic is created. This made it kinda difficult to debug stuff.
  - I'll do some clean up in that file after #929 is merged (to avoid conflicts).

* Once we do that, we could just check SQLAlchemy's integrityerror rather than using _assert_absent. I don't know why we do this to begin with, is there a good reason?

@zenhack thoughts?

zenhack:
We should definitely add a unique constraint everywhere it makes sense.

Re: catching the error from SQLAlchemy, I'm not sure it's even the same across database backends. I seem to remember the postgres backend throwing a ProgrammingError or somesuch in an analogous situation.

Also, I don't want to be over-broad in what errors we're catching -- blindly assuming any IntegrityError is because of the unique constraint violation could end up masking real bugs.

The way I've had this in my head (and if we agree on it we should probably write it down as an official guideline) is to basically write code that should never cause SQLAlchemy to raise an exception. Doing this has the nice advantage that if we get a stack trace with an IntegrityError in it, we don't have to waste time figuring out whether or not that's supposed to happen.

Finally, I don't think just catching the sqlalchemy exception would even be any more concise; we still need to translate it into an APIError to report it to the client, and `_assert_absent`/`_must_find` aren't a lot of code to begin with.

naved001:
Alright, I'm convinced. Let's just add the unique constraint wherever applicable.

xuhang57:
@naved001 ，whenever you have time, we could talk about the implementation details for a bit.

PR #971, opened 2018-03-07
Enforce unique contraints on object names/labels in the database.

xuhang57:
Fix: #946 

Potentially, we should think about whether there are any other constraints.

And this might break lots of tests. Consider this is an on-going implementation

### DIFF
--- a/hil/migrations/versions/264ddaebdfcc_make_labels_unique.py
+++ b/hil/migrations/versions/264ddaebdfcc_make_labels_unique.py
@@ -7,7 +7,7 @@ Create Date: 2018-03-12 11:15:09.729850
 """
 
 from alembic import op
-#import sqlalchemy as sa
+# import sqlalchemy as sa
 
 
 # revision identifiers, used by Alembic.
--- a/hil/migrations/versions/264ddaebdfcc_make_labels_unique.py
+++ b/hil/migrations/versions/264ddaebdfcc_make_labels_unique.py
@@ -7,7 +7,7 @@ Create Date: 2018-03-12 11:15:09.729850
 """
 
 from alembic import op
-import sqlalchemy as sa
+#import sqlalchemy as sa
 
 
 # revision identifiers, used by Alembic.
--- a/hil/migrations/versions/264ddaebdfcc_make_labels_unique.py
+++ b/hil/migrations/versions/264ddaebdfcc_make_labels_unique.py
@@ -1,0 +1,32 @@
+"""make labels unique
+
+Revision ID: 264ddaebdfcc
+Revises: 89ff8a6d72b2
+Create Date: 2018-03-12 11:15:09.729850
+
+"""
+
+from alembic import op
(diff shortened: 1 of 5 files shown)


xuhang57:
Some help would be much appreciated. 

I am suspecting that because this is called many time:

https://github.com/CCI-MOC/hil/blob/45f218d272a4da8cd21694bcf1de0783a165b756/tests/unit/api/auth.py#L612

so the node name is not unique in the CI database?

zenhack:
Could you also write a migration script? Other than that, LGTM.

naved001:
project names, switch names, network names are unique too.

hil/migrations/versions/264ddaebdfcc_make_labels_unique.py
```diff
@@ -0,0 +1,32 @@
...
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '264ddaebdfcc'
+down_revision = '89ff8a6d72b2'
+branch_labels = ('hil')
```
zenhack: You'll want a trailing comma here; this is supposed to be a tuple.

tests/unit/migrations/flask.sql
```diff
@@ -287,7 +287,7 @@ ALTER SEQUENCE nic_id_seq OWNED BY nic.id;
 
 CREATE TABLE node (
     id integer NOT NULL,
-    label character varying NOT NULL,
+    label character varying NOT NULL UNIQUE,
```
xuhang57: Naved and Jeremy suggested that this is not the place that does the migration tests. And the question now is how we can increase the test coverage? Or, where do we do the migration tests?

zenhack: We shouldn't **ever** be modifying this file; the whole point of the tests is to verify that we can cleanly upgrade from an old database schema; this is a snapshot of a database we should be able to upgrade from. What you've done here is manually modify the old schema.

The actual tests are in `hil/migrations.py`.

The missing coverage in `upgrade` is incorrect; we're definitely running that in the migration tests (and it seems to be erroneously missing from all of the other scripts as well). downgrade() is really untested, and we basically haven't done testing with downgrading databases at all.

My inclination is to:

1. Figure out why the upgrade function doesn't look like it's covered
2. Open a separate issue about the fact that we're not testing downgrade() anywhere. Honestly, I'm not sure it's worth the effort to properly support database downgrades, and we've only been adding the function at all because alembic expects it. But we should discuss the issue.

I'm fine opening separate issues for both of these and just overriding the coveralls warning for now.

naved001:
> Honestly, I'm not sure it's worth the effort to properly support database downgrades, and we've only been adding the function at all because alembic expects it.

I agree.

zenhack:
I'm happy. If everyone's in agreement re: coveralls we can just merge.

naved001:
Cool, merging then.

PR #972, opened 2018-03-08
Move show_console to the client library

xuhang57:
Simply move `show_console` to the client library.

Relates to: https://github.com/CCI-MOC/hil/issues/924

### DIFF
--- a/hil/api.py
+++ b/hil/api.py
@@ -1289,7 +1289,7 @@ def show_console(nodename):
     if log is None:
         raise errors.NotFoundError(
             'The console log for %s does not exist.' % nodename)
-    return json.dumps(log)
+    return log
 
 
 @rest_call('PUT', '/node/<nodename>/console', Schema({'nodename': basestring}))
--- a/hil/api.py
+++ b/hil/api.py
@@ -1289,7 +1289,7 @@ def show_console(nodename):
     if log is None:
         raise errors.NotFoundError(
             'The console log for %s does not exist.' % nodename)
-    return log
+    return json.dumps(log)
 
 
 @rest_call('PUT', '/node/<nodename>/console', Schema({'nodename': basestring}))
--- a/hil/cli.py
+++ b/hil/cli.py
@@ -760,6 +760,7 @@ def show_console(node):
     """Display console log for <node>"""
     C.node.show_console(node)
 
+
 @cmd
 def start_console(node):
     """Start logging console output from <node>"""
--- a/hil/cli.py
+++ b/hil/cli.py
@@ -758,7 +758,7 @@ def list_headnode_images():
 @cmd
 def show_console(node):
     """Display console log for <node>"""
-    print C.node.show_console(node)
+    print (C.node.show_console(node))
 
 
 @cmd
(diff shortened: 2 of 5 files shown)


xuhang57:
Hmm, I am still getting a 

```
FailedAPICallException: The console log for node-01 does not exist.
```

I started the console before I show it. Did I miss anything here?

zenhack:
If it's using the mock obm driver, it looks like actually showing the console isn't implemented (`hil/ext/obmd/mock.py`). If not for the fact that we're looking at migrating stuff to obmd soonish, I might think it was worth stubbing more of this out and maybe putting some logic in there to just e.g. set a boolean when start/stop console are called, and check it in get_console, returning ether None or a string depending. That may actually not be a whole lot of work, in which case it might make sense to do it anyway, but it also seems sensible to just say screw it, and revisit the coverage of the console related calls when we re-do that API.

naved001:
I am happy with this PR, this seems simple enough to me. We can either update the mock driver to handle get_console or revisit it later (meaning not have the client lib test for it). I am fine either ways.
Alternatively, can we use the ipmi driver with dry_run set to true?

xuhang57:
Thanks, guys!

naved001:
Turns out the client library was using the ipmi driver rather than the mock driver. Any reason? I changed it to use the mock driver and put some logic like you suggested it @zenhack

naved001:
On a slightly related note, I don't understand why the client library is registering all sorts of drivers when it's not using those. What's the intention here?

hil/ext/obm/mock.py
```diff
@@ -50,22 +52,30 @@ def require_legal_bootdev(self, dev):
     def set_bootdev(self, dev):
         return
 
-    @no_dry_run
     def start_console(self):
+        global CONSOLE
+        CONSOLE = True
```
zenhack: The all-caps convention normally indicates something that's a constant, so it shouldn't be used for a variable that you're modifying.

It would also be better for this to operate on a per-obm level, as opposed to having start/stop console toggle *all* obms. Maybe have a look at how the mock switch driver does this (it has a dictionary).

naved001:
@zenhack done.

I'll let @xuhang57 merge master into this (or rebase) after #978 is in.

hil/api.py
```diff
@@ -1289,7 +1289,7 @@ def show_console(nodename):
     if log is None:
         raise errors.NotFoundError(
             'The console log for %s does not exist.' % nodename)
-    return log
+    return json.dumps(log)
```
zenhack: I don't think casting this to JSON is necessarily the right thing here. Once we move this to obmd we'll want to expose the raw byte stream in the client, so maybe we should just make that tweak now.

hil/cli.py
```diff
@@ -758,8 +758,7 @@ def list_headnode_images():
 @cmd
 def show_console(node):
     """Display console log for <node>"""
-    url = object_url('node', node, 'console')
-    do_get(url)
+    print C.node.show_console(node)
```
zenhack: Let's start putting parens around our print statements; this is a python 2 vs. 3 compat thing.

xuhang57: sure, and Naved has to modify this in the new cli anyways I believe.

hil/api.py
xuhang57: should we clean up in another PR, since we have lots of places are using `json.dumps`

zenhack:
I'm happy.

Issue #975, opened 2018-03-12
Flask 0.10 -> 0.11?


xuhang57:
Not sure whether we want to bother ourselves to upgrade the flask from `0.10` to `0.11` (current stable version is `0.12`). And it seems to be a not-so-easy upgrade, at least, flask adds a new cli and we may need to change our migrations process as well (by using the new flask cli). Thoughts?

![screen shot 2018-03-12 at 10 06 31](https://user-images.githubusercontent.com/7245860/37288358-73635140-25dd-11e8-87eb-e2c0e0767810.png)

zenhack:
It doesn't sound like they're actually dropping Flask-Script yet; I think we could upgrade flask without having to switch to their CLI tool at the same time.

The api for the new thing also isn't dramatically different; I don't think porting we be a ton of work. The big question is how to we wrangle Flask-Migrations?

xuhang57:
True, but are you saying that upgrading to `0.11` will break the Flask-Migrations? Or break our current setup about it?

I haven't figured everything out yet. According to https://flask-migrate.readthedocs.io/en/latest/
We should be able to do migrations with the flask CLI (we have to construct it for sure) and it seems like it is gonna take some effort.

zenhack:
@xuhang57, we're already using Flask-Script, as Flask-Migrate has built-in support for it: https://flask-script.readthedocs.io/en/latest/

Upgrading Flask shouldn't break anything; we can keep right-on using Flask-Script for the migration script, and it probably makes sense to do so until Flask-Migrate supports the thing that's been added to Flask core.

The warning on the Flask-Script page is aimed at potential new users, telling them that Flask-Script is *redundant* with functionality now in Flask core, but it's not *broken* afaik.

PR #977, opened 2018-03-16
Upgrade Flask from 0.10.1 -> 0.11.1 -> 0.12.1

xuhang57:
Currently, the most stable release is 0.12. (http://flask.pocoo.org/docs/0.12/changelog/)

And not sure whether we should jump to 0.12 or we should go from 0.10 -> 0.11 ->0.12

Relates: https://github.com/CCI-MOC/hil/issues/975


naved001:
it's odd that the obmd migration tests are failing :/

zenhack:
I suspect the test failures are due to the change in https://github.com/CCI-MOC/obmd/pull/10; we need to add the DB info to the config file we're generating there.

I'd say let's upgrade straight to 0.12.

xuhang57:
Coverage is decreased because we never test the `setup.py` file? 

I don't have the hil `pypi` creds but should we add a test for testing `pip install hil`?
Not sure whether adding this in `travis` would make more sense?

xuhang57:
The coveralls actually shows there is no change in coverage percentage. However, the check fails here on GitHub. I will restart the build and see whether it fixes.

Issue #571, opened 2016-04-11
REST API should be versioned


henn:
There should be a version number in the REST API.

We can start with v0

shwsun:
@henn what about we set the REST API v0.1 when we ship the HIL project, before the SoCC ddl?

xuhang57:
Do we just get to choose whatever number as a starting point as we want? @naved001

zenhack:
I would basically follow semver: https://semver.org/

With the exception that:

* We don't need to bother bumping the version with changes to the v0 api
* The v0 api doesn't need to be "older" than the stable apis; we can treat it as development ground rather than "pre" stabilization.
* The package versions don't have to coinside with the API versions.
* There's no patch part of the number, since we're just talking about an API.

xuhang57:
@zenhack 👍 , sounds a good plan. I have no problem with that, will send a PR soon.

xuhang57:
@zenhack  on a second thought, this is not a small task as I thought before. If we have the versioning, let say, `1.1`, does that mean we need to change all the rest auth url? 

For example, change `/node/<node>` to `/node/v1.1/<node>`?

zenhack:
My thinking was to start we'd just prefix everything with `/v0`. Existing clients will basically "just work" if we add `/v0` onto the endpoint they're using; at most we might want to namespce the client library so users are specifically opting in to the version.

So we'd change `/node/<node>` to /v0/node/<node>`.

xuhang57:
That should work well. But I don't know how to work on this at the moment.
so leave this issue open for now.

xuhang57:
@zenhack let's say we manually add the prefix to all rest calls. What if we want to bump the version to v0.2? Then we have to manually update every one of them.

Is there some kind of variable mechanism we could use? 
For example

version = "v0"

`version + /node/<node>`
`version + /network/<network>`
`version + /project/<project>`

zenhack:
Let me mull this over a bit; I have some half-baked ideas about how to deal with this in the server-side code, but need to think about it a bit more.

xuhang57:
What OpenStack does: https://docs.openstack.org/nova/latest/contributor/microversions.html

zenhack:
Yeah, the mechanism there seems conceptually equivalent, though I like putting it in the path better because it makes it impossible not to specify a version.

My question was more about how to implement the interface, in terms of python's routing and such.

[...]

PR #983, opened 2018-03-22
Discuss: API should be versioned

xuhang57:
See details in: https://github.com/CCI-MOC/hil/issues/571

Basically, we should discuss whether we should get started for API versioning through an approach this patch purposes. 

And to test: 

1. pull this PR
2. install hil and configure hil and start hil server
3. use:
```bash
$ hil list_nodes free
or
$ curl -H "Content-Type: application/json" -X GET $HIL_ENDPOINT/nodes/free
```
And for testing request methods that are not `GET`:
for example,
```bash
$ hil project_create test-project
or
$ curl -X put http://127.0.0.1:5000/project/mockproject01 -u hil:hil
```

### DIFF
--- a/hil/rest.py
+++ b/hil/rest.py
@@ -163,7 +163,7 @@ def rest_call(methods, path, schema, dont_log=()):
         * A tuple, whose first element is a string (the response body), and
           whose second is an integer (the status code).
     """
-    
+
     def register(f):
         """Return value from rest call; this decorates the function itself."""
 
--- a/hil/rest.py
+++ b/hil/rest.py
@@ -89,7 +89,7 @@ class ValidationError(APIError):
     """An exception indicating that the body of the request was invalid."""
 
 
-def rest_call(methods, path, schema, dont_log=()):
+def rest_call(methods, path, schema, version={'version': 'v0'}, dont_log=()):
     """A decorator which registers an http mapping to a python api call.
 
     `rest_call` makes no modifications to the function itself, though the
@@ -176,13 +176,13 @@ def rest_call(methods, path, schema, dont_log=()):
 
         app.add_url_rule(path,
                          f.__name__,
-                         _rest_wrapper(f, schema, dont_log),
+                         _rest_wrapper(f, version, schema, dont_log),
                          methods=meths)
         return f
     return register
 
 
-def _do_validation(schema, kwargs):
+def _do_validation(version, schema, kwargs):
     """Validate the current request against `schema`.
 
     `schema` should be a schema as passed to `rest_call`.
@@ -215,6 +215,9 @@ def _do_validation(schema, kwargs):
                 raise ValidationError("Empty parameter specified")
             else:
                 final_kwargs[key] = value
+        print version
+        if version['version'] != 'v0':
+            raise ValidationError("API version does not match")
     else:
         # Methods other than GET can use path and body arguments
         if flask.request.data != '':
@@ -243,9 +246,12 @@ def _do_validation(schema, kwargs):
         # which, while fairly clear and helpful, is obviously
         # talking about python types, which is gross.
         raise validation_error
+    print version
+    if version['version'] != "v0":
+        raise validation_error
 
 
-def _rest_wrapper(f, schema, dont_log):
+def _rest_wrapper(f, version, schema, dont_log):
     """Return a wrapper around `f` that does the following:
 
     * Validate the current request against the schema.
@@ -260,7 +266,7 @@ def _rest_wrapper(f, schema, dont_log):
 
     def wrapper(**kwargs):
         """The wrapper described above."""
-        kwargs = _do_validation(schema, kwargs)
+        kwargs = _do_validation(version, schema, kwargs)
 
         censored_kwargs = kwargs.copy()
         for argname in dont_log:
--- a/hil/rest.py
+++ b/hil/rest.py
@@ -22,6 +22,7 @@ from uuid import uuid4
 from hil import auth
 
 local = flask.g
+VERSION = 'v0'
 
 
 class _RequestInfo(object):
@@ -195,7 +196,6 @@ def _do_validation(version, schema, kwargs):
     If the schema does not validate, this will raise an instance of
     `ValidationError`.
     """
-
     final_kwargs = {}
 
     if flask.request.method == "GET":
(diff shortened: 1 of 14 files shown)


hil/rest.py
```diff
@@ -231,6 +235,10 @@ def _do_validation(schema, kwargs):
             raise validation_error
         final_kwargs[k] = kwargs[k]
 
+    print version
+    if version['version'] != VERSION:
```
xuhang57: I haven't figured out why we cannot put the check for the version after the `schema.validate`

If we put it after, it will never check the version.

hil/rest.py
```diff
@@ -215,6 +215,10 @@ def _do_validation(schema, kwargs):
                 raise ValidationError("Empty parameter specified")
             else:
                 final_kwargs[key] = value
+
+        print version
```
xuhang57: for testing purpose. Will delete once we are sure about this.

xuhang57:
thanks Naved for sending this to me:
 https://stackoverflow.com/questions/389169/best-practices-for-api-versioning

I believe it worths us all to take a look 👍 
@naved001 @Izhmash @zenhack

zenhack:
So, I'm increasingly skeptical of the notion that REST is even a good idea for our use cases, and most of the arguments around "good REST design" seem 90% aesthetic, which for a tool that's supposed to be very low-level feels rather bikeshed. Early on I spent a lot of personal energy fussing about the way we do URIs and requests bodies etc. so that it felt "natural" in that pardigm, but the paradigm basically puts for thinking about URIs as resources, documents, and the use of the API as shunting data around, which is an awkward, clumsy way to model what HIL actually does. In retrospect, I think with something more RPC-ish we would have wasted a lot less time bikesheding the mapping of our design to HTTP.

There *are* important bits of thinking about an HTTP mapping; http verbs have specific semantics around things like caching, side effects, etc. But I've grown weary of arguments that basically amount to "it fits the metaphor better." Metaphors are there to help us understand something, not as a prescription.

From a very pragmatic perspective, I think putting the URL in the version is a good idea, becuase it forces the user to actually specify the version they're depending on. The arguments against it in that stack overflow thread are, frankly, coming more from a dogmatic view about "what REST should be" rather than any kind of serious thought about the problem. The REST metaphor makes sense in some cases, but I'm alergic to paradigms, especially as a way to guide pragmatic design decisions. Historically "good object oriented design" has been used to justify a lot of really crappy techincal decisions; see <https://plus.google.com/+RobPikeTheHuman/posts/hoJdanihKwb> for a good description. I worry REST has become the same thing to an extent.

We basically cargo-culted openstack when we started this project, which is universally a terribly idea that I've been pushing back on since very early on. Sometimes they do stuff that's sensible, but it's hit or miss, which is why I get a bit annoyed any time "openstack does this" comes up as an argument for something.

I guess what I'm getting at is I'd urge people to think about the concrete. This isn't a call to necessarily ignore aesthetics; a good developer experience is something that I care deeply about. From the closed tag on that post:

> Many good questions generate some degree of opinion based on expert experience, but answers to this question will tend to be almost entirely based on opinions, rather than facts, references, or specific expertise.

Sorry for the rant. I guess I feel like the a lot of the details of how to do this are pretty bikeshed, and I'm a bit weary of the discussion.

I also feel like a detailed design is premature; we haven't even talked about when we're going to start stabalizing the API, or what sort of guarantees we want to have. All I really want to do right now is tack `'/v0'` on to the front of our URLs so that existing tools are very clearly marked as pulling from the unstable API.

xuhang57:
We don't have a detailed plan of what to do with the API versioning so it is not bikeshed yet. Thanks so much for the input.

Maybe we should spend some time discussing and writing some specs first so we can actually implement this. And I agree it is still premature for having a detailed plan for now. 

In other words, now I do think putting a `/v0` on to the front is a good idea. I will update this PR accordingly.

And as a side note, I don't know whether we should even go back to the RPC world. REST solves the problems and it just works. GraghQL might be the future and I honestly don't know how it will go. Admittedly HIL is obviously not like Facebook that has tons of nodes/trees/graph. But GraphQL does allow you to query the exact data you are looking for in one request and I believe it is very powerful.

Again, putting `/v0` should give us what we need for now.

zenhack:
Ok, so sounds like we're agreed on the immediate term: just add `/v0` to the front of the URLs.

Re: rpc, I think our current API design is perfectly functional and it doesn't seem worth the effort to port everything over to something else. My only complaint is that there's a bit of a disconnect between how people tend to talk/think about it (basically procedure names) and the notion of "REST," and with each change to the API there's been some overhead in discussing how to map things to HTTP, which in retrospect I don't think really provided any value. An rpc design would have obviated those discussions.  But again, I don't think it's a big enough deal to change course.

Re: GraphQL/other speculation: none of these approaches are universally applicable. They each have their use cases, but none of them are "the future" in any general sense. A lot of the ideas behind (all of) these designs have waxed and waned in mindshare over the years, resurging in new forms. Best to avoid the fads and think about actual requirements.

zenhack:
This looks good to me; once the obmd tests are fixed assuming Travis passes I'm happy.

I'm working on debugging the Go stuff now.

naved001:
Couple of things:

* Could you document this somewhere? `docs/USING.rst`, `docs/rest_api.md` are the few places I think where we talk about the URI.
* rebase this on top of master to get the CI fix

xuhang57:
I rebased it. But these commits are still listed separately.

naved001:
hmmm, there's a bunch of test failures.

zenhack:
Ah, yeah, looks like the `/v0` needs to be added in a bunch of places in the tests as well.

xuhang57:
I am thinking of how we can create a global `/v0` in the tests so that we only need to update one place. Maybe we should even put the version in one settings file?

xuhang57:
I added the `/v0` to a few places in the tests. However, it seems we are still having some problems with user operation testing. Still trying to figure out why.

zenhack:
@xuhang57, looks like the client library is inserting an extra slash, which is causing the server to return a 404, since it can't match the path `/v0//projects`.

naved001:
the commit history is really weird :/ You should probably rebase this and pick the commits you want, or idk maybe squash it.

xuhang57:
okay, it is because in the `client/project.py` and `client/switch.py`, both of the `list` function have the extra `/` in the `self.object_url`.

xuhang57:
I am going to send a new PR. I messed up with the commit history

PR #987, opened 2018-04-03
Add release notes section to UPGRADING document.

### LINKED ISSUE #865: Create release note documenting upgrades with haas -> hil name change
For the 0.2 release, we need to document that:

- [ ] Since the systemd script is created manually, the old one needs to manually be deleted and the new one copied in.
- [ ]  If wanting to keep the previous `haas` user, then the script needs to be modified to reflect the different username.
- [ ] The `/var/lib/haas` and `/etc/haas.cfg` entries need to be moved (or at least symlinked)
- [ ] You should remove the "haas" version, since it will be a different set of scripts: `pip uninstall haas`
- [ ] Re-copy `hil.wsgi` and update apache's wsgi.conf entry to point to it.
- [ ] Update any scripts that have env vars (like `HAAS_ENDPOINT`) to their HIL_ varieties.

zenhack:
Fixes #865 

Note that the 0.3 section is not necessarily complete; I added stuff I could think of off the top of my head, and skimmed the closed issues on the 0.3 milestone (where I didn't find anything that seemed like it needed mentioning).

### DIFF
--- a/docs/UPGRADING.rst
+++ b/docs/UPGRADING.rst
@@ -47,7 +47,7 @@ Release Notes
   undetected you may need to fix them before the servers will start again
 * The interface to the ``hil`` command line tool has changed substantially;
   If you have custom scripts that invoke it they will likely need to be
-  modified.
+  modified. Use ``hil --help`` to explore the new interface.
 * URLs for API calls are now prefixed with a version number, which for the
   current (unstable) API is ``v0``; you may need to update scripts and
   settings accordingly (although the HIL tools themselves should remain
--- a/docs/UPGRADING.rst
+++ b/docs/UPGRADING.rst
@@ -53,6 +53,19 @@ Release Notes
   settings accordingly (although the HIL tools themselves should remaining
   internally consistent).
 
+Other changes, which do not require specific action when upgrading:
+
+* All HIL APIs are now wrapped by the client library.
+* Support for a new optional "maintenance pool" feature; see
+  ``docs/maintenance-pool.md``.
+* Some new APIs (see ``docs/rest_api.md`` for details):
+  * Networking actions can now be queried to get their status.
+  * When using the database auth backend, it is possible to list users.
+* Updated APIs:
+  * List networks now shows public networks to regular users.
+  * Some switches now support public key authentication. See
+    ``docs/network-drivers.md`` for details.
+
 0.2
 +++
 
--- a/docs/UPGRADING.rst
+++ b/docs/UPGRADING.rst
@@ -3,7 +3,8 @@ Upgrading HIL
 
 This file describes the procedure for upgrading HIL to a new version.
 
-1. Read the release notes, as they may contain version-specific information.
+1. Read the release notes for the particular release, below, which will cover
+   any version-specific information.
 2. Stop the HIL services (webserver and HIL network daemon). For exmaple::
 
      $ systemctl stop httpd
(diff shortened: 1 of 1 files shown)


docs/UPGRADING.rst
```diff
@@ -34,3 +35,38 @@ This file describes the procedure for upgrading HIL to a new version.
 
      $ systemctl start httpd
      $ systemctl start hil_network
+
+Release Notes
+-------------
+
+0.3 (Upcoming)
```
xuhang57: Upcoming is a bit vague to me. Maybe having an actual date?

docs/UPGRADING.rst
```diff
@@ -34,3 +35,38 @@ This file describes the procedure for upgrading HIL to a new version.
...
+++++++++++++++
+
+* Validation checks for the configuration file have been added; be advised
+  that if there were errors in your ``hil.cfg`` that were previously
+  undetected you may need to fix them before the servers will start again
+* The interface to the ``hil`` command line tool has changed substantially;
```
xuhang57: maybe we need to have an updated doc so we can point the readers to the relevant documentation?

docs/UPGRADING.rst
naved001: we don't really have an actual date. it's just as soon as we are done with all our issues marked with 0.3 milestone.

naved001:
Some more things that might be useful to put in the changelog. 
* All HIL APIs now have client library support. 
* Support for maintenance pool.
* Some new APIs
  - Networking actions can now be queried to get their status. 
  - List users when using database backend
* Updated APIs
  - List networks now shows public networks to regular users.
* Some switches now support public key authentication. See `docs/network-drivers.md` for details.

docs/UPGRADING.rst
xuhang57: sure, am working on the version PR, which is the last PR for 0.3 milestones I believe

naved001:
I am happy.

docs/UPGRADING.rst
naved001: I think just mentioning `--help` should be enough. We could probably, in the future, add some documentation about the CLI (even a man page for HIL `man hil`).

docs/UPGRADING.rst
```diff
@@ -34,3 +35,51 @@ This file describes the procedure for upgrading HIL to a new version.
...
+  If you have custom scripts that invoke it they will likely need to be
+  modified.
+* URLs for API calls are now prefixed with a version number, which for the
+  current (unstable) API is ``v0``; you may need to update scripts and
+  settings accordingly (although the HIL tools themselves should remaining
```
naved001: typo: should **remain** internally consistent

docs/UPGRADING.rst
```diff
@@ -34,3 +35,51 @@ This file describes the procedure for upgrading HIL to a new version.
...
+* Updated APIs:
+  * List networks now shows public networks to regular users.
+  * Some switches now support public key authentication. See
+    ``docs/network-drivers.md`` for details.
+
+0.2
++++
+
+HaaS was renamed to HIL in this release. Accordingly, there are several
+changess that need to be made when upgrading:
```
naved001: typo: changess -> changes

zenhack:
@naved, fixed.

naved001:
not sure if we should merge it now, or after the last PR for 0.3 is merged?

zenhack:
I don't see any reason not to do so now.

Quoting Naved Ansari (2018-04-03 17:47:13)

>    not sure if we should merge it now, or after the last PR for 0.3 is
>    merged?

PR #988, opened 2018-04-04
API should be versioned

xuhang57:
See details in: #571 & https://github.com/CCI-MOC/hil/pull/983

Basically, we should discuss whether we should get started for API versioning through an approach this patch purposes.

### DIFF
--- a/tests/integration/keystone.py
+++ b/tests/integration/keystone.py
@@ -247,7 +247,7 @@ def test_admin_call(keystone_projects, user_info):
     sess = _get_keystone_session(username=user_info['name'],
                                  password=user_info['password'],
                                  project_name=user_info['project_name'])
-    resp = _do_get(sess, 'admin-only')
+    resp = _do_get(sess, 'v0/admin-only')
     if user_info['admin']:
         assert resp.status_code < 300, (
             "Status code for admin-only call should be successful for admin."
@@ -270,7 +270,7 @@ def test_project_call(keystone_projects, caller_info, project_name):
     sess = _get_keystone_session(username=caller_info['name'],
                                  password=caller_info['password'],
                                  project_name=caller_info['project_name'])
-    resp = _do_get(sess, 'project-only/' + project_name)
+    resp = _do_get(sess, 'v0/project-only/' + project_name)
     if caller_info['admin'] or caller_info['project_name'] == project_name:
         assert 200 <= resp.status_code < 300, (
             "Status code for project-only call should be successful for "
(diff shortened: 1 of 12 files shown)


zenhack:
Still a few more failures. making progress though.

xuhang57:
A bit confused on how Keystone Auth works in HIL. Not sure why the integration tests are failing.

Izhmash:
It looks like the versioning broke the API, I'm seeing 404s on Travis. I don't think it has to do directly with keystone auth.

xuhang57:
hmm, right, I haven't figured out which part of the test missing the `/v0` prefix yet

jeremyfreudberg:
In tests/integration/keystone.py you are registering some dummy endpoints as a text fixture (in  the `extra_apis` fixture, specifically) with `rest_call` (which has been modified to insert the "/v0").

However, `_do_get` in this file does not prepend the "/v0", so various tests fail with a 404

xuhang57:
@jeremyfreudberg thanks, good catch!

tests/integration/keystone.py
```diff
@@ -247,7 +247,7 @@ def test_admin_call(keystone_projects, user_info):
     sess = _get_keystone_session(username=user_info['name'],
                                  password=user_info['password'],
                                  project_name=user_info['project_name'])
-    resp = _do_get(sess, 'admin-only')
+    resp = _do_get(sess, 'v0/admin-only')
```
jeremyfreudberg: wouldn't it make more sense, at least in terms of planning for the future, to make this change inside of `_do_get` itself, instead of editing each call to it?

xuhang57: hmm, doing this would allow us to test different versions with different functions. And maybe some functions we want to label them as stable first.

jeremyfreudberg: we can leave it as you have it

naved001:
Looks good to me, just a couple of requests:

1. Could you update the curl calls here? https://github.com/CCI-MOC/hil/blob/master/docs/USING.rst
2. I think it makes sense to mention the current API version in `docs/rest_api.md` (maybe in the overview section).

zenhack:
This was the last thing on the 0.3 milestone; shall we tag a release?

naved001:
Yeah, go for it.

zenhack:
OK, tagged. I pushed one more commit just removing the "upcoming" label (which is tagged as v0.3) straight to master; hopefully that's cool.

naved001:
Awesome,  Thanks! Yay for 0.3!

PR #989, opened 2018-04-07
Test fixes

zenhack:
I hit a few problems trying to run the test locally; here are the fixes.


zenhack:
@naved001, it was one of the `test_bool_validation_*` in `tests/unit/config.py`; str.lower choked on the bool that it fished out of the config object.

The dry run stuff cropped up when I added the `isinstance` assert; those tests tripped it.

zenhack:
Merging, since this has 2 approvals.

PR #990, opened 2018-04-07
Fix lax schema for node_register api call.

zenhack:
This is built on top of #989, so only the last commit is really relevant. See the commit message. Let's merge #989 first, and then I can rebase.


zenhack:
Rebased, now that #989 is merged.

PR #997, opened 2018-04-18
Fix new pycodestyle errors

naved001:
This has @zenhack 's commit in it, and I added a commit to address the switch regexes. I ran the native networks deployment tests for both the switches and those passed. Though, the vlan_networks tests failed due to some assertions errors, looking into it right now; but I don't think these are related to this since even without these changes those failed.

I tested the powerconnect for the `_console.py` changes

### DIFF
--- a/hil/api.py
+++ b/hil/api.py
@@ -348,7 +348,7 @@ def node_connect_network(node, nic, network, channel=None):
     """
 
     def _have_attachment(nic, query):
-        """Return whether there are any attachments matching ``query`` for ``nic``.
+        """Return if there are any attachments matching ``query`` for ``nic``.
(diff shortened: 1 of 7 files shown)


zenhack:
I want to wait until you figure out what's wrong with the deployment tests, but otherwise LGTM.

naved001:
It was a couple of things:
* the vlans I was using in my testsuite.cfg didn't exist on the switch (for the dellnos9), so that caused some failures. 
* the switchports under test weren't clean from the beginning. 

I re ran both network deployment tests again on a dellnos9 switch, a powerconnect switch, and a cisco nexus switch. All passed.

Issue #921, opened 2017-12-13
`get_port_networks()` doesn't read inactive vlans (on the powerconnect switch).


naved001:
```
Trunking VLANs Enabled: 1511
                        1512 (Inactive)
``` 
VS
```
Trunking VLANs Enabled: 1511-1512
```
This was causing deployment tests to fail on the powerconnect switch.
From the comment [here](https://github.com/CCI-MOC/hil/blob/master/hil/ext/switches/_dell_base.py#L59-L61) it's not clear if this is the intended behavior or not.

zenhack:
From the code it looks like the issue may actually be that it's split on to the next line.

naved001:
Since we are telling hil admins to create all VLANs that HIL will be using, do we still need to worry about reading inactive VLANs? I guess not, if you agree you can close this.

zenhack:
Per the above, I don't think `(Inactive) ` is what's causing the problem; I suspect it's the line break.

naved001:
> I suspect it's the line break.

I think so too. So I think we should update the method to read the inactive VLANs too because those VLANs **are allowed** on that switchport and may get activated in the future.

naved001:
#998 fixes this.

PR #998, opened 2018-04-19
Read inactive vlans

Izhmash:
- Addresses #921
- Effects PowerConnect, Nexus, and N3000 switches.
- `get_port_networks()` now returns every VLAN in the ranges rather than the ones at the end of the ranges.
- I am releasing this now for review to get feedback before writing the deployment tests to test more than two trunking VLANs

### DIFF
--- a/hil/ext/switches/_dell_base.py
+++ b/hil/ext/switches/_dell_base.py
@@ -2,6 +2,7 @@
 
 import re
 import logging
+from hil.ext.switches.common import parse_vlans
 
 from hil.ext.switches import _console
 
@@ -35,37 +36,37 @@ class _BaseSession(_console.Session):
         self.disable_vlan(vlan_id)
         self._sendline('sw trunk native vlan none')
 
+
     def get_port_networks(self, ports):
-        num_re = re.compile(r'(\d+)')
         port_configs = self._port_configs(ports)
-        result = {}
+        badchars = ' (Inactive)'
+        network_list = []
         for k, v in port_configs.iteritems():
-            native = v['Trunking Native Mode VLAN'].strip()
-            match = re.match(num_re, native)
-            if match:
-                # We need to call groups to get the part of the string that
-                # actually matched, because it could include some junk on the
-                # end, e.g. "100 (Inactive)".
-                num_str = match.groups()[0]
-                native = int(num_str)
+            non_natives = ''
+            non_native_list = []
+            # Get native vlan then remove junk if native not None
+            native_vlan = v['Trunking Native Mode VLAN'].strip()
+            if (native_vlan != 'none'):
+                native_vlan = ''.join(c for c in native_vlan
+                                      if c not in badchars)
             else:
-                native = None
-            networks = []
-            range_str = v['Trunking VLANs Enabled']
-            for range_str in v['Trunking VLANs Enabled'].split(','):
-                for num_str in range_str.split('-'):
-                    num_str = num_str.strip()
-                    match = re.match(num_re, num_str)
-                    if match:
-                        # There may be other tokens in the output, e.g.
-                        # the string "(Inactive)" somteimtes appears.
-                        # We should only use the value if it's an actual number
-                        num_str = match.groups()[0]
-                        networks.append(('vlan/%s' % num_str, int(num_str)))
-            if native is not None:
-                networks.append(('vlan/native', native))
-            result[k] = networks
-        return result
+                native_vlan = None
+            # Get other vlans and parse out junk if not None
+            trunk_vlans = v['Trunking VLANs Enabled'].strip()
+            if (trunk_vlans != 'none'):
+                non_natives = ''.join(c for c in trunk_vlans
+                                      if c not in badchars)
+                non_natives = non_natives.split('\r\n')
+                non_natives = ','.join(non_natives)
+                non_native_list = parse_vlans(non_natives)
+            else:
+                non_natives = None
+            if native_vlan is not None:
+                network_list.append(('vlan/native', native_vlan))
+            for v in (non_native_list):
+                if v != native_vlan:
+                    network_list.append(('vlan/%s' % v, int(v)))
+        return network_list
 
     def disable_port(self):
         self._sendline('sw trunk allowed vlan none')
--- a/hil/ext/switches/_dell_base.py
+++ b/hil/ext/switches/_dell_base.py
@@ -36,7 +36,6 @@ class _BaseSession(_console.Session):
         self.disable_vlan(vlan_id)
         self._sendline('sw trunk native vlan none')
 
-
     def get_port_networks(self, ports):
         '''Returns every trunking VLAN and native VLAN from a port config.
         Example: Port 1 has 'Trunking Native Mode VLAN': ' 3 (Inactive)',
--- a/hil/ext/switches/_dell_base.py
+++ b/hil/ext/switches/_dell_base.py
@@ -36,33 +36,43 @@ class _BaseSession(_console.Session):
         self.disable_vlan(vlan_id)
         self._sendline('sw trunk native vlan none')
 
-
     def get_port_networks(self, ports):
+        '''Returns every trunking VLAN and native VLAN from a port config.
+        Example: Port 1 has 'Trunking Native Mode VLAN': ' 3 (Inactive)',
+        'Trunking VLANs Enabled': 'none'
+        Port 2 has 'Trunking Native Mode VLAN': 'none',
+        'Trunking VLANs Enabled': ' 10, 2000'
+        Port 3 has 'Trunking Native Mode VLAN': 'none',
+        'Trunking VLANs Enabled': ' 23 - 25'
+
+        Return: [('vlan/native', '3'), ('vlan/10', 10), ('vlan/2000', 2000),
+                 ('vlan/23', 23), ('vlan/24', 24), ('vlan/25', 25)]
+        '''
         port_configs = self._port_configs(ports)
         badchars = ' (Inactive)'
         network_list = []
+        # iterate through the port configurations
         for k, v in port_configs.iteritems():
             non_natives = ''
             non_native_list = []
-            # Get native vlan then remove junk if native not None
+            # get native vlan then remove junk if native is not None
             native_vlan = v['Trunking Native Mode VLAN'].strip()
-            if (native_vlan != 'none'):
+            if native_vlan != 'none':
                 native_vlan = ''.join(c for c in native_vlan
                                       if c not in badchars)
+                network_list.append(('vlan/native', native_vlan))
             else:
                 native_vlan = None
-            # Get other vlans and parse out junk if not None
+            # get other vlans and parse out junk if there are any
             trunk_vlans = v['Trunking VLANs Enabled'].strip()
-            if (trunk_vlans != 'none'):
+            if trunk_vlans != 'none':
+                # remove ' (Inactive)' if it's there; want all VLANS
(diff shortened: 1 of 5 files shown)


hil/ext/switches/n3000.py
```diff
@@ -44,14 +41,13 @@ class DellN3000(Switch):
...
-        Schema({
+        schema.Schema({
             'username': basestring,
             'hostname': basestring,
             'password': basestring,
-            'dummy_vlan': And(Use(int),
-                              lambda v: 0 < v <= 4093,
-                              Use(str)),
+            'dummy_vlan': schema.And(schema.Use(int),
+                                     lambda v: 0 < v and v <= 4093,
```
zenhack: Why get rid of the `0 < v <= 4093` shorthand?

Also, while I don't necessary have an objection to qualifying all of the variables from the `schema` module, I don't understand how it's related to this patch?

Izhmash: I'll clean this PR up, I'm not sure how this change snuck in there.

Izhmash: Actually this should be it for style-related things, the rest should be functional. I'm not sure why the diff is appearing a bit strangely in `_dell_base`

hil/ext/switches/_dell_base.py
```diff
@@ -35,37 +36,37 @@ def disable_native(self, vlan_id):
...
-                # We need to call groups to get the part of the string that
-                # actually matched, because it could include some junk on the
-                # end, e.g. "100 (Inactive)".
-                num_str = match.groups()[0]
-                native = int(num_str)
+            non_natives = ''
+            non_native_list = []
+            # Get native vlan then remove junk if native not None
+            native_vlan = v['Trunking Native Mode VLAN'].strip()
+            if (native_vlan != 'none'):
```
zenhack: parens aren't necessary.

hil/ext/switches/_dell_base.py
```diff
@@ -35,37 +36,37 @@ def disable_native(self, vlan_id):
...
+                native_vlan = ''.join(c for c in native_vlan
+                                      if c not in badchars)
```
zenhack: This is somewhat subjective, but I personally find python's if expression really hard to read. Maybe just use filter and a lambda?

hil/ext/switches/_dell_base.py
```diff
@@ -35,37 +36,37 @@ def disable_native(self, vlan_id):
...
-                        num_str = match.groups()[0]
-                        networks.append(('vlan/%s' % num_str, int(num_str)))
-            if native is not None:
-                networks.append(('vlan/native', native))
-            result[k] = networks
-        return result
+                native_vlan = None
+            # Get other vlans and parse out junk if not None
+            trunk_vlans = v['Trunking VLANs Enabled'].strip()
+            if (trunk_vlans != 'none'):
```
zenhack: Same thing; no need for parens in an if.

hil/ext/switches/_dell_base.py
naved001: personally I find filter and lambdas hard to read. I find what @Izhmash has done to be more readable.

Izhmash:
@zenhack I've added some documentation for the Dell switch, if the docs there are good I can add them to the other drivers too.

Overall, my `get_port_networks()` iterates through ports' port configurations and gets every native + non-native trunking VLAN mentioned in the configuration.  The majority of the code here is string parsing, which some logic that branches off depending on if the switch reports VLANs or if it reports 'none'.

naved001:
> I don't understand how this fixes the bug? in #921 the discussion settled on the problem probably being the newline, not the inactive string. I don't understand where the solution is here?

I think @Izhmash ended up rewriting the whole method so that in addition to reading inactive vlans, it also parses the vlan range correctly.

hil/ext/switches/_dell_base.py
```diff
@@ -36,36 +37,46 @@ def disable_native(self, vlan_id):
...
+            trunk_vlans = v['Trunking VLANs Enabled'].strip()
+            if trunk_vlans != 'none':
+                # remove ' (Inactive)' if it's there; want all VLANS
+                non_natives = ''.join(c for c in trunk_vlans
+                                      if c not in badchars)
+                non_natives = non_natives.split('\r\n')
+                # create comma-separated list to use parse_vlans()
+                non_natives = ','.join(non_natives)
+                non_native_list = parse_vlans(non_natives)
+            for v in (non_native_list):
```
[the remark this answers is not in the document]
naved001: haha writing in Go did this to @Izhmash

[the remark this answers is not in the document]
zenhack: I don't understand how? what construct in Go was confusing him?

zenhack:
Ah, okay, I understand what's going on now.

hil/ext/switches/_dell_base.py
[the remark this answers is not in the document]
Izhmash: No worries, just a typo

hil/ext/switches/_dell_base.py
```diff
@@ -36,36 +37,46 @@ def disable_native(self, vlan_id):
...
+        network_list = []
+        # iterate through the port configurations
+        for _, v in port_configs.iteritems():
+            non_natives = ''
+            non_native_list = []
+            # get native vlan then remove junk if native is not None
+            native_vlan = v['Trunking Native Mode VLAN'].strip()
+            if native_vlan != 'none':
+                native_vlan = ''.join(c for c in native_vlan
+                                      if c not in badchars)
```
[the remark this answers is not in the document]
Izhmash: I'll give that a go, that sounds a lot better.

Izhmash:
> You've got a fair amount of copypasta from one driver to another. Ideally we'd factor out the common bits.

The n3000 and the PowerConnect code here is the same, should I move the function I wrote to `_dell_base`?

Izhmash:
> You've got a fair amount of copypasta from one driver to another. Ideally we'd factor out the common bits.
The n3000 function I wrote is the same as the PowerConnect's; I can have them use the same code.  I'll look into factoring out the Nexus code too.

Izhmash:
So I've removed the n3000 port networks codes and created a unified on in `_dell_base`.  The n3000 and PowerConnect are very similar; only small changes in key word ordering and none representations got in theway.

I've tried to make it more readable by separating some of the code out.

Izhmash:
My latest commit allows the deployment tests to run successfully.  Tested for Nexus and PowerConnect.

Izhmash:
N3000 deployment tests working as well.

Izhmash:
I've added a deployment test that tests the discovery of multiple (>2) networks on one port using `get_port_networks`.

hil/ext/switches/_dell_base.py
```diff
@@ -36,37 +37,74 @@ def disable_native(self, vlan_id):
...
+                native_vlan = native_vlan.replace(' (Inactive)', '')
+                if hasattr(self.switch, 'dummy_vlan'):
+                    # ignore if dummy vlan
+                    if int(native_vlan) == int(self.switch.dummy_vlan):
+                        native_vlan = None
+                    else:
+                        network_list.append(('vlan/native', int(native_vlan)))
+                else:
+                    # appropriate to append native vlan to list
+                    network_list.append(('vlan/native', int(native_vlan)))
```
naved001: we could probably collapse the 2 if statements into 1:
```python
if hasattr(self.switch, 'dummy_vlan') and int(native_vlan) == int(self.switch.dummy_vlan):
    native_vlan = None
else:
    network_list.append(('vlan/native', int(native_vlan)))
````

that way we don't have 2 repeated else blocks either.

File: hil/ext/switches/_dell_base.py [hunk 8]
```diff
@@ -36,37 +37,74 @@ def disable_native(self, vlan_id):
...
+                    non_native_list.remove(native_vlan)
+                # make final vlan list
+                for v in non_native_list:
+                    # checking for extra whitespace
+                    v = "".join(v.split())
+                    network_list.append(('vlan/%s' % v, int(v)))
+            result[k] = network_list
         return result
 
+    def _make_vlan_list(self, dirty_list):
```
naved001: this method could be outside the class (it doesn't need `self` at all)

Izhmash:
@naved001 I fixed those up.

hil/ext/switches/_dell_base.py
```diff
@@ -36,35 +37,63 @@ def disable_native(self, vlan_id):
...
+        '''Returns every trunking VLAN and native VLAN from a port config.
+        Example: Port 1 has 'Trunking Native Mode VLAN': ' 3 (Inactive)',
+        'Trunking VLANs Enabled': 'none'
+        Port 2 has 'Trunking Native Mode VLAN': 'none',
+        'Trunking VLANs Enabled': ' 10, 2000'
+        Port 3 has 'Trunking Native Mode VLAN': 'none',
+        'Trunking VLANs Enabled': ' 23 - 25'
+
+        Return:
+        {<hil.model.Port object at 0xDEADBEEF0123>: [('vlan/native', '3')],
```
naved001: deadbeef

tests/deployment/multi_networks.py
```diff
@@ -0,0 +1,152 @@
...
+class TestMultiNets(NetworkTest):
+    """NetworkTest using multiple tagged vlan networks on a port."""
+
+    def test_multi_networks(self):
+        """Do a bunch of network operations on the switch, and verify things
+        along the way.
+
+        The above is super vague; unfortunately the setup operations are very
+        slow, so it makes a huge difference to do everything in one pass. See
+        the comments in-line to understand exactly what is being tested.
```
zenhack: We've now got this exact docstring copied in 3 places. I'm worried about how much copypasta I've been seing in patches to the test suite lately...

tests/deployment/multi_networks.py
```diff
@@ -0,0 +1,152 @@
...
+            """
+
+            nodes = self.collect_nodes()
+
+            # create 5 networks
+            network_create_simple('net-0', 'anvil-nextgen')
+            network_create_simple('net-1', 'anvil-nextgen')
+            network_create_simple('net-2', 'anvil-nextgen')
+            network_create_simple('net-3', 'anvil-nextgen')
+            network_create_simple('net-4', 'anvil-nextgen')
```
zenhack: This seems like it should be  a loop (same below).

Izhmash:
@zenhack I think I addressed your comments.  Also I'd like to test this again since it's been a little while, I'll put in a comment when that's done

tests/deployment/multi_networks.py
```diff
@@ -63,12 +63,11 @@ def create_multi_nets():
 
             nodes = self.collect_nodes()
 
+            deferred.apply_networking()
```
naved001: this shouldn't be here.

Izhmash: oops I'll remove that

Izhmash:
Deployment tests are still passing

tests/deployment/multi_networks.py
```diff
@@ -0,0 +1,147 @@
...
+                                     'with_request_context',
+                                     'site_layout')
+
+
+class TestMultiNets(NetworkTest):
+    """NetworkTest using multiple tagged vlan networks on a port."""
+
+    def test_multi_networks(self):
+        """A suite of tests to ensure that multiple networks can relate
+        to a single port. Previously, there was an issue with
```
zenhack: Would be good to add a reference to the specific Github issue.

Izhmash: Should be all set now

Issue #973, opened 2018-03-09
Get rid of the mechanism of saving switch config files to startup


naved001:
See discussion here https://github.com/CCI-MOC/hil/pull/964#issuecomment-371244248

Our plan is to:
* Require that the startup config has no networks on HIL controlled switchports.
* Provide an API to deploy an entire switch's configuration from HIL database.

Issue #970, opened 2018-03-07
Turn off/on ports in brocade switch (or disable switchport)


naved001:
Ports on a brocade switch alway default to a native VLAN 1. We should either disable the switchport (`no switchport`) until a native vlan is set, or just shut off the port completely. 

I have mentioned this on a comment somewhere, but I am opening this issue so that I don't forget.

naved001:
This was the case on brocade vdx 6740 running 5.0.1, but now that @radonm upgraded the switches to 7.0 there's a new trunk mode called `trunk-no-default-native` which as the name suggests will have no default vlan in trunk mode.

Along with the switch models we should document what version of OS did we test the driver against.

PR #1000, opened 2018-04-19
Refactor dellnos9 and brocade switch drivers

naved001:
Just some things that I think could be changed. There's info in the commit message.

Deployment tests passed for both switches.

Modify port looks quite similar, but there are some differences. I could make those changes and push modify port to the super class if we are going to have switches that are unlike the console switches. Other than that, we could put it in `switches.common` and then add the method from there to the switch class.

I can also remove saving-to-switch from this PR which would help simplify modify port.

### DIFF
--- a/hil/ext/switches/_unconsole.py
+++ b/hil/ext/switches/_unconsole.py
@@ -1,0 +1,104 @@
+"""Common functionality for switches that aren't session oriented and have
+an API to talk to"""
+
+import logging
+import re
+import requests
+
+from hil.errors import SwitchError
+from hil.model import SwitchSession
+from hil.network_allocator import get_network_allocator
+
+_CHANNEL_RE = re.compile(r'vlan/(\d+)')
+logger = logging.getLogger(__name__)
+
+
+class Session(SwitchSession):
+    """Common base class for sessions in switches that are using an API"""
+
+    def modify_port(self, port, channel, new_network):
+        """This implements modify port that vlan centric switches can use"""
+        if channel == 'vlan/native':
+            if new_network is None:
+                self._remove_native_vlan(port)
+                self._port_shutdown(port)
+            else:
+                self._set_native_vlan(port, new_network)
+
+        else:
+            match = re.match(r'vlan/(\d+)', channel)
+            assert match is not None, "Malformed channel: No VLAN ID found"
+            vlan_id = match.groups()[0]
+            legal = get_network_allocator(). \
+                is_legal_channel_for(channel, vlan_id)
+            assert legal, "Invalid VLAN ID"
+
+            if new_network is None:
+                self._remove_vlan_from_trunk(port, vlan_id)
+            else:
+                assert new_network == vlan_id
+                self._add_vlan_to_trunk(port, vlan_id)
+
+    def revert_port(self, port):
+        self._remove_all_vlans_from_trunk(port)
+        if self._get_native_vlan(port) is not None:
+            self._remove_native_vlan(port)
+        self._port_shutdown(port)
+
+    def disconnect(self):
+        pass
+
+    def get_port_networks(self, ports):
+        response = {}
+        for port in ports:
+            native = self._get_native_vlan(port.label)
+            if native is not None:
+                response[port] = [native]
+            else:
+                response[port] = []
+            response[port] += self._get_vlans(port.label)
+        return response
+
+    def _make_request(self, method, url, data=None,
+                      acceptable_error_codes=()):
+        """This can make the http request for you.
+        Also accetps a list of acceptable error codes if you need."""
+
+        r = requests.request(method, url, data=data, auth=self._auth)
+        if r.status_code >= 400 and \
+           r.status_code not in acceptable_error_codes:
+            logger.error('Bad Request to switch. '
+                         'Response: %s and '
+                         'Reason: %s', r.text, r.reason)
+            raise SwitchError('Bad Request to switch. '
+                              'Response: %s and '
+                              'Reason: %s', r.text, r.reason)
+        return r
+
+    @property
+    def _auth(self):
+        return self.username, self.password
+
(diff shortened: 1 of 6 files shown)


naved001:
I am thinking I might as well make a superclass for non-console like drivers.

hil/ext/switches/common.py
```diff
@@ -53,3 +56,36 @@ def parse_vlans(raw_vlans):
...
+            self._set_native_vlan(port, new_network)
+    else:
+        match = re.match(r'vlan/(\d+)', channel)
+
+        if match is None:
+            raise SwitchError("Malformed channel: No VLAN ID found")
+
+        vlan_id = match.groups()[0]
+        # we already do these checks in the API, are we being extra careful
+        # by doing it here or is this redundant?
```
zenhack: I think it's definitely a good idea to do the extra sanity check, but given that it should never fail, I'd use an `assert`.

hil/ext/switches/common.py
```diff
@@ -53,3 +56,36 @@ def parse_vlans(raw_vlans):
             vlan_list.append(num_str)
 
     return vlan_list
+
+
+def modify_port_common(self, port, channel, new_network):
```
zenhack: Rather than making this a stand-alone function and assigning it in the class definitions, it would be more idiomatic for it to be a mixin class.

hil/ext/switches/common.py
```diff
@@ -53,3 +56,36 @@ def parse_vlans(raw_vlans):
...
+    """This implements modify port method that non-console like drivers can use
```
zenhack: Rather than "non-console", I might suggest mentioning that it is vlan-specific (we still don't have any other channel types, but this is the constraint).

hil/ext/switches/common.py
```diff
@@ -53,3 +56,36 @@ def parse_vlans(raw_vlans):
...
+        # we already do these checks in the API, are we being extra careful
+        # by doing it here or is this redundant?
+        legal = get_network_allocator(). \
+            is_legal_channel_for(channel, vlan_id)
+
+        if not legal:
+            raise SwitchError("Invalid VLAN ID")
+
+        if new_network is None:
+            self._remove_vlan_from_trunk(port, vlan_id)
```
zenhack: we should document which additional methods need to be implemented by `self`.

hil/ext/switches/common.py
naved001: oh right, that's the path I am on right now, didn't know it's called mixin class. Will push it in the next commit. I am factoring out more stuff.

tests/unit/ext/switches/brocade.py
```diff
@@ -3,12 +3,31 @@
...
+    config_merge({
+        'extensions': {
+            'hil.ext.network_allocators.null': None,
+            'hil.ext.network_allocators.vlan_pool': '',
+        },
+        'hil.ext.network_allocators.vlan_pool': {
+            'vlans': '100-105',
+        },
+    })
+    config.load_extensions()
```
zenhack: What changed to make this necessary?

hil/ext/switches/_unconsole.py
```diff
@@ -0,0 +1,104 @@
+"""Common functionality for switches that aren't session oriented and have
+an API to talk to"""
```
zenhack: Should document that this is specific to the vlan_pools allocator.

hil/ext/switches/_unconsole.py
```diff
@@ -0,0 +1,104 @@
...
+                              'Response: %s and '
+                              'Reason: %s', r.text, r.reason)
+        return r
+
+    @property
+    def _auth(self):
+        return self.username, self.password
+
+    def _remove_native_vlan(self, interface):
+        assert False
```
zenhack: Should probably add a message here, like the `assert False` statements in other classes.

tests/unit/ext/switches/brocade.py
naved001: in modify port I am using `get_network_allocator().is_legal_channel_for()` to assert that vlan_id is valid for channel. I remember [you suggested](https://github.com/CCI-MOC/hil/pull/866#discussion_r137414606) during the review of the dellnos9 driver. 

So, without this change, if there's no network allocator then it wouldn't work.

hil/ext/switches/_unconsole.py
naved001: yeah, good point.

naved001:
@zenhack: Should I do #973 and/or #970 while am at it, or in separate PRs? Besides that, is there anything else that you think can be factored out?

tests/unit/ext/switches/brocade.py
zenhack: Aha, makes sense.

zenhack:
Let's keep this one self contained; those issues can be addressed in separate prs.

naved001:
To make `modify_port` and `revert_port` generic, I removed the the calls to `save_running_config` in dellnos9.

naved001:
> Missing docstrings

Done. Wondering why didn't pylint complain about it given many of these methods aren't declared in the `SwitchSession()` class in `hil/model.py`

zenhack:
LGTM

naved001:
@zenhack I have one doubt here; If I wanted to use this new module with the openVswitch driver it will contain the _make_request method which it doesnt need. Is that okay in theory for that class to have that attribute?

zenhack:
It shouldn't *break* anything. That said it may be a bit cleaner to have it be a stand-alone function (and make auth a parameter, rather than a separate property). I'm fine either way.

naved001:
Just rebased this and fixed conflicts.

zenhack:
Cool. Can someone do a second review on this?

naved001:
@izhmash was a bit busy this week with other things. He'll be back tomorrow
though.

PR #1002, opened 2018-04-26
Implement the first couple obmd-related api calls.

zenhack:
Just enabling/disabling the obm.

### DIFF
--- a/ci/obmd/run_integration_tests.sh
+++ b/ci/obmd/run_integration_tests.sh
@@ -1,0 +1,14 @@
+#!/usr/bin/env sh
+set -ex
+
+# Set environment variables so we use the newer version of Go, rather than
+# the 1.7 that's installed in Travis by default:
+export GOROOT=/usr/lib/go-1.10
+export PATH="$HOME/go/bin:/usr/lib/go-1.10/bin:$PATH"
+
+# Just in case, forget any cached command locations:
+hash -r
+
+go get -v github.com/CCI-MOC/obmd
+
+py.test --cov=hil --cov-append tests/integration/obmd.py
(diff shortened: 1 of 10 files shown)


ci/obmd/run_integration_tests.sh
```diff
@@ -0,0 +1,14 @@
...
+
+# Set environment variables so we use the newer version of Go, rather than
+# the 1.7 that's installed in Travis by default:
+export GOROOT=/usr/lib/go-1.10
+export PATH="$HOME/go/bin:/usr/lib/go-1.10/bin:$PATH"
+
+# Just in case, forget any cached command locations:
+hash -r
+
+go get -v github.com/CCI-MOC/hil
```
naved001: you wanna go get CCI-MOC/obmd

zenhack: D'oh!

zenhack:
@naved001, all set.

hil/api.py
```diff
@@ -246,6 +248,23 @@ def node_register(node, obmd, **kwargs):
     db.session.commit()
 
 
+@rest_call('PUT', '/node/<node>/obm', Schema({
+    'node': basestring,
+    'enabled': bool,
+}))
+def node_enable_disable_obm(node, enabled):
```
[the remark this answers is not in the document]
zenhack: I suppose we *could* hide these operations as an implementation detail behind every call that used them, but that runs somewhat counter to the usual design decision of having the client deal with managing resources, so consistency at least argues in favor of making this explicit.

It's also worth noting that the token does get exposed to the client (see the redirects added in #1007), so I guess it at least feels a little weird twiddling it when we're also giving it to the client, albiet somewhat implicitly.

Izhmash:
I will have time to review this tomorrow.

PR #1004, opened 2018-04-26
Updates documentation about node_register 

### LINKED ISSUE #1003: Documentation for node_register
#991 introduced new parameters required for registering a node. 
Looks like updating the [obm-drivers](../../hil/blob/master/docs/obm-drivers.md) got missed out. 
Needs updating.

SahilTikale:
Fixes #1003 
@naved001 
only one reviewer required. short PR.

### DIFF
--- a/docs/obm-drivers.md
+++ b/docs/obm-drivers.md
@@ -45,12 +45,14 @@ The type field for the IPMI driver has the value::
 
 IPMI driver requires three additional feilds
 to be able to communicate with the IPMI subsystem of a server.
-These feilds are `host`, `user`, and `password`
+These feilds are `host`, `user`, and `password`.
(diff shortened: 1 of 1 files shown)


naved001:
Trailing whitespace causing failures:

You can put this in your `.vimrc`: `autocmd BufWritePre * %s/\s\+$//e`  I got this from here: http://vim.wikia.com/wiki/Remove_unwanted_spaces

zenhack:
Having this in the driver section doesn't make a lot of sense, because it's not at all driver specific. But we're gutting the entire driver infrastructure soon, so I'm not going to worry about it. When the dust is settled this whole section will be gone and we'll just describe those parameters in the docs for node_register.

PR #1008, opened 2018-05-02
Make travis a little faster

naved001:
This PR does a few things to address #992 (I think we can do more stuff)

* It moves the client tests to the integration tests directory.
* Integration and deployment tests are run in a separate build (with only postgres)|
* Unit tests are run with both, sqlite and postgres.
* Linters are run along with unit tests and where db=sqlite 
* We only setup apache and keystone for the integration tests.

Build finishes in under 6 minutes for unit tests & linters and 9 for the rest of the integration/deployment tests.

I think it's time we rearrange the CI folder a little, but haven't thought exactly how to.


naved001:
I can't figure out how to put the `addons` under `env`. So it's still installing `postgresql: "9.3"`, `apache2` and `libapache2-mod-wsgi` for all `env`

zenhack:
Coveralls is reporting a signficant drop in coverage; looks like it doesn't think the client lib is being covered anymore?

ci/apache/run_integration_tests.sh
```diff
@@ -11,6 +11,7 @@ cd $TRAVIS_BUILD_DIR
 
 # Test commands
 py.test --cov=hil --cov-append tests/integration/cli.py
+py.test --cov=hil --cov-append tests/integration/client.py
```
naved001: this should have taken care of coverage for the client lib tests. Don't know what's wrong with it :/ will have to think about it.

naved001:
Searching for "coverage dropped in coveralls" takes me to open github issues from other people :/

ci/run_tests.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+if [ $TEST_SUITE = integration ]
+then
```
zenhack: Style nitpick: all of the other shell scripts put `then` on the same line.

ci/run_tests.sh
```diff
@@ -0,0 +1,13 @@
...
+    sh -e ci/apache/run_integration_tests.sh
```
zenhack: Something odd going on with the indentation here;.

ci/run_tests.sh
```diff
@@ -0,0 +1,13 @@
...
+	sh -e ci/keystone/run_integration_tests.sh
+	sh -e ci/deployment-mock-networks/run_integration_tests.sh
+else
+	sh -e ci/pycodestyle_tests.sh
+	sh -e ci/pylint_tests.sh
+	sh -e ci/no_trailing_whitespace.sh
+	sh -e ci/run_unit_tests.sh
+fi
```
zenhack: Looks like you're missing a trailing newline.

naved001:
@chemistry-sourabh @xuhang57 You guys know what's up with coveralls? I just moved my tests to a different folder (`tests/unit/client.py` to `tests/integration/client.py`), and coveralls think that the client library is not being tested.

ci/apache/run_integration_tests.sh
```diff
@@ -10,7 +10,8 @@ hil-admin create-admin-user $HIL_USERNAME $HIL_PASSWORD
 cd $TRAVIS_BUILD_DIR
 
 # Test commands
-py.test --cov=hil --cov-append tests/integration/cli.py
+py.test --cov=hil --cov-append tests/integration/cli.py \
```
Izhmash: I don't think this is the correct way to add the test coverage to coveralls, since CLI reports 0 coverage as well.

Izhmash: So rather than Coveralls, I think it's pytest-cov's fault.  The coverage reported in Travis comes from that. That's where we should start.

xuhang57:
hmm, I don't have a clue yet. Something wrong with the client library tests for sure.

ci/run_unit_tests.sh
```diff
@@ -3,12 +3,12 @@
 # On SQLite we run the tests in parallel.
 # This speeds things up, but is currently only safe for sqlite, since it uses
 # an in memory (and  therefore per-process) database.
+# submit coverage report only for postgres builds
 
 if [ $DB = sqlite ]; then
     extra_flags='-n auto'
```
zenhack: Noticed this because of it's proximity to the code you added -- the indentation here doesn't match the convention in the rest of the file, and other shell scripts.

naved001:
I am really puzzled why there is no coverage for integration tests (or any of the postgres builds). I can't reproduce this locally (running integration tests with postgres gives me expected coverage report).

zenhack:
So, I have a thought:

* So far, we haven't had any organized effort to act on the coverage information; coveralls is just keeping us posted on coverage.
* I don't think it's reports have really gotten us to write any more tests -- we seem to be keeping on top of coverage at least not getting worse as we do new stuff.
* We've definitely spent a decent chunk of time debugging coveralls, and thus far it hasn't really had any benefit.

I'm half tempted to suggest just junking coveralls. I basically trust us to insist on writing tests for new code, and if we're going to work on improving existing coverage, given that I'm not worried about regressions, I'd be fine with people just posting comments showing coverage before and after. There's only so much fighting with tooling that makes sense.

zenhack:
LGTM

PR #1009, opened 2018-05-03
Rework test http client to handle external redirects.

zenhack:
I ran into this problem when working on the obmd stuff, and it's a self-contained enough thing to submit as a separate pr. See the commit message and the docstrings for full details.

I expect coveralls will complain about decreasing test coverage, since I've added code that isn't excercised except on my WIP obmd branch.

PR #1010, opened 2018-05-04
Fix a bug that was silencing errors from the server.

zenhack:
I caught this while working on OBMd stuff, which raises another question: right now OBMd doesn't follow the same error reporting convention as the HIL api server.

### DIFF
--- a/tests/unit/client_regression.py
+++ b/tests/unit/client_regression.py
@@ -13,10 +13,9 @@ from hil import config, rest
 from hil.client.base import FailedAPICallException
 from hil.client.client import Client
 from hil.test_common import HybridHTTPClient, fail_on_log_warnings, \
-    fresh_database, server_init, config_testsuite
+    server_init, config_testsuite
 
 fail_on_log_warnings = pytest.fixture(fail_on_log_warnings)
-fresh_database = pytest.fixture(fresh_database)
 server_init = pytest.fixture(server_init)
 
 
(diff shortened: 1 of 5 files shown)


naved001:
If it's simple enough, can we write a unit test for this?

zenhack:
Hm, I think that's worth doing, but thinking about how it's a little fiddly. I'd like to wait until #1009 is merged, since I'll need to use the test client, and don't really want to deal with potential conflicts. I'll probably end up moving that into test_common, and have the test for this in a separate file from the existing client (integration) tests. I think I'll tag this one as "waiting on other change."

zenhack:
Added a test.

zenhack:
(Per the docstring at the top of the file, we'll want to merge your thing and then rename it).

tests/unit/client_regression.py
```diff
@@ -0,0 +1,56 @@
...
+from schema import Schema
+
+from hil import config, rest
+from hil.client.base import FailedAPICallException
+from hil.client.client import Client
+from hil.test_common import HybridHTTPClient, fail_on_log_warnings, \
+    fresh_database, server_init, config_testsuite
+
+fail_on_log_warnings = pytest.fixture(fail_on_log_warnings)
+fresh_database = pytest.fixture(fresh_database)
```
naved001: why is this fixture required?

zenhack: It isn't, I just add the "standard" fixtures without really thinking about it -- I've removed it.

Izhmash:
@zenhack we'll need to fix the merge conflicts here

zenhack:
Rebased. I also did the rename suggested in the docstring, now @naved001's pr is merged.

naved001:
Cool, merging.

PR #1013, opened 2018-05-15
Optimize postgres to make it run slightly faster

naved001:
This patch modifies the postgres configuration file and then restarts the
postgres server.

Locally this made a significant difference, but in Travis it only seems to save a minute or so. I don't know if it's worth it, but we can discuss if we want this or not.

Anybody want to test this locally too (so that we can be sure that there isn't something else going on in my local environment).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,7 @@ env:
 install:
   # A dependency might use new pip syntax; upgrade to prevent breakage. See #771
   - pip install --upgrade pip
+  - sh -e ci/optimize_postgres.sh
   - sh -e ci/hil_install.sh
   - sh -e ci/apache/install.sh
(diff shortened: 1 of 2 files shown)


Izhmash:
I didn't see a significant speedup running these tests locally on my laptop, but I think it's a good idea to reduce overhead regardless

naved001:
> I wonder what impact putting the database in /tmp would have; this would prevent it from needing to hit the disk at all.

Should we look into that before merging this?

zenhack:
I think we can leave it as future work.

Izhmash:
Just need a rebase here.

naved001:
rebased

zenhack:
Looks like there's a cert error on the openstack servers.

PR #1015, opened 2018-05-15
Install obmd only when running integration tests

naved001:
* doing apt-get update and then dowloading go-lang (~100MB) takes up a minute
or so.

Edit:
Actually, it takes way more than a minute. From the build of this PR:
"Fetched 108 MB in 5min 26s (330 kB/s)"

### DIFF
--- a/ci/obmd/install.sh
+++ b/ci/obmd/install.sh
@@ -1,6 +1,11 @@
 #!/usr/bin/env sh
 set -ex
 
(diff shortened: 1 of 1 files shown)


zenhack:
LGTM, but as a sanity check I want to wait until #1007 is merged, in case this interacts badly with that. I don't expect it to, but...

naved001:
closing this in favor of #1019

PR #1018, opened 2018-05-22
Use github mirror instead of git.o.o

jeremyfreudberg:
Just a test... not in the mood to learn how to run Travis without making a PR

### DIFF
--- a/ci/keystone/keystone.sh
+++ b/ci/keystone/keystone.sh
@@ -4,10 +4,13 @@
 #
 #   http://docs.openstack.org/developer/keystone/developing.html
 
+# NOTE: We use the GitHub mirror of OpenStack's respositories rather than the
+# official git.openstack.org host, due to some known certificate problems.
+
 # tunable variables:
 
 # So we can avoid repeatedly downloading the repo when developing this script:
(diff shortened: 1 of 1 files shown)


Izhmash:
For future reference if this ever breaks, `pip install -U requests[security]` also fixes the certificate issue.

Izhmash:
Note: this fixes the current Travis certificate error.

zenhack:
What does the `[securty]` extra actually contain? If the problem is a software support thing on our end, just specifying that in our config might be a better solution.

Izhmash:
@zenhack  I'm not sure exactly what that contains, but in a quick test I found it also works.  @jeremyfreudberg 's solution works fine though since it doesn't require any more installing.

jeremyfreudberg:
@zenhack `[security]` adds a few extra packages, see here: https://github.com/requests/requests/blob/master/setup.py#L99

the solution that i have provided here would be prioritizing speed

zenhack:
Ok. Could you add a comment mentioning why we're pulling from github? Should document the fact that this is purposeful, rather than just an arbitrary choice.

naved001:
@zenhack there's a comment now, you can merge this if you are satisfied.

zenhack:
Works for me, merged.

PR #1019, opened 2018-05-31
Download obmd binary rather than installing it everytime the CI is run

naved001:
* This saves us time when running our CI

See https://github.com/CCI-MOC/obmd/issues/21


zenhack:
Holy crap, Travis hasn't been that fast in *forever*. LGTM, merging.

PR #1020, opened 2018-05-31
A couple test related improvements.

zenhack:
See the individual commit messages.

I don't like the way the file names stutter (e.g. `unit/client_unit.py`), but py.test really doesn't do a good job of namespacing anything, so...


hil/test_common.py
```diff
@@ -437,7 +438,21 @@ def obmd_cfg():
...
         """
-        portno = 8833
+        # Make the port number a function of which worker we're running on,
+        # so parallel tests don't try to use the same port. We derive the
+        # port number by stripping out all non-digit characters from the
+        # worker id, and then adding the result to a base. If the worker id
+        # isn't set, we're probably not running in parallel, so just use
+        # the base.
+        portno = 8800
+        worker = os.getenv('PYTEST_XDIST_WORKER')
```
naved001: Out of curiosity, how did you figure out that this variable refers to gw0/1/2.. in the output? `plugins: requests-mock-1.5.0, xdist-1.22.2, forked-0.2, cov-2.5.1
gw0 C / gw1 C / gw2 C / gw3 C` Could you point me to the documentation because I couldn't find

naved001:
oh thanks, I also searched the variable name and couldn't find it. need to look carefully; *smh*

naved001:
@Izhmash ping

hil/test_common.py
```diff
@@ -437,7 +438,21 @@ def obmd_cfg():
...
+        if worker is not None:
+            portno += + int(''.join([
```
jeremyfreudberg: is that stray `+` intentional? (it doesn't break anything, but does it have some purpose that I'm not seeing?)

Izhmash:
I'm good with all this once jeremy's question is addressed.

hil/test_common.py
zenhack: No, just a typo -- I've removed it.

zenhack:
Cool, this has two signoffs, merging.

PR #1021, opened 2018-06-01
use postgres 9.6 in CI

naved001:
There are some under the hood performance improvements that we might benefit from in our CI. 
https://www.postgresql.org/about/featurematrix/#performance

Travis images don't have postgres 10, though we can manually install it but it's way too much work for no benefit imo, I did it here https://github.com/CCI-MOC/hil/compare/master...naved001:use-postgres-10
All the manual setup causes it to be slower than the already available postgres 9.X. You can see the travis builds for that here https://travis-ci.org/naved001/hil/builds/386740666

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ cache:
   - pip
 
 addons:
-  postgresql: "9.3"
+  postgresql: "9.6"
   apt:
     packages:
       - apache2


zenhack:
Hm, The running time is actually a bit slower... the last run:

https://travis-ci.org/CCI-MOC/hil/builds/386365104

Vs. yours:

https://travis-ci.org/CCI-MOC/hil/builds/386810705?utm_source=github_status&utm_medium=notification

Not terribly surprising; they're probably not tuning for running in a VM with durabilty features turned off, so I'd expect it to be a crapshoot from one release to the next.

zenhack:
(I also don't know what we can expect from Travis re: consistency of performance -- we're probably sharing the machine with other runs, so it may be slower just because it's the middle of the day instead of late at night).

naved001:
Well, then we might as well close this then and come to back upgrading postgres in the future.

Btw, do you think you could port your dummy_verify fixture to work with `tests/unit/ext/auth/database.py` They take around 40 seconds to run. I blindly used the fixture you wrote and that made no difference. Do you think it's worth looking into it? (it might require more thought)

naved001:
my bad, I am actually noticing a difference of ~10 seconds on my machines;

PR #1022, opened 2018-06-01
Use dummy_verify fixture for running database auth unit tests

naved001:
With the fixture it ran in 30 seconds and without it 40 seconds consistently on my local machine. We may not notice in travis because the variance between different travis runs is probably greater than 10 seconds. So do you guys think it's worth doing it or not @zenhack @Izhmash ?


zenhack:
I'd rather not muck with this in the tests for the backend itself; this is disabling some of the functionality we should be testing. The problem this was originally trying to solve was a difference of half an hour, for tests that were conceptually unrelated. I think we should just put up with the 10 seconds.

naved001:
alright, cool.

PR #1023, opened 2018-06-01
Add power on api call.

zenhack:
...now that obmd has this, we should add another redirect.

This is built on top of #1020, which should be merged first.

### DIFF
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -607,6 +607,17 @@ Authorization requirements:
 
 * Access to the project to which `<node>` is assigned (if any) or administrative access.
 
+#### node_power_on
+
+`POST /node/<node>/power_on`
+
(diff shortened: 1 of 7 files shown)


hil/test_common.py
```diff
@@ -437,7 +438,21 @@ def obmd_cfg():
         be used as a fixture, but not declared here as such; individual
         modules should declare it as a fixture.
         """
-        portno = 8833
+        # Make the port number a function of which worker we're running on,
```
Izhmash: Why did this need to get added in this PR?  Did an issue come up?

naved001:
putting the label on this because we want to merge #1020 first.

Izhmash:
#1020 merged, all we need is a rebase now.

zenhack:
Rebased.

PR #1026, opened 2018-06-07
Fix typo in psql docs

Izhmash:
Fixes a small typo to match up instructions and commands.

### DIFF
--- a/docs/INSTALL-devel.rst
+++ b/docs/INSTALL-devel.rst
@@ -129,8 +129,8 @@ eg. There has to be a system user `hil` to access database named `hil` as databa
 
 Create database `hil_dev` owned by database role `hil_dev`::
 
-  sudo -i -u hil_dev
-  $ createdb hil
+  $ sudo -i -u hil_dev
+  $ createdb hil_dev
 
 Confirm it created a database named `hil_dev` and it is owned by `hil_dev`::
 


naved001:
On a slightly unrelated note, I think our installation docs are a tad bit confusing to follow. Sometimes even I get tired of doing all that, which is why I wrote some helper scripts which can be found here https://github.com/CCI-MOC/admin-tools/tree/master/hil_installation

In the long term I want to do that using Ansible.

Izhmash:
@naved001  I saw that, that definitely helps too.

zenhack:
LGTM, merging.

zenhack:
@naved001, re: the installation docs in general, while in production you definitely want to be automating things, I very much think it's important to keep "manual" installation within the grasp of mere mortals. If this was ever a property openstack had, it was lost before the MOC even existed, and when the docs were first written, George and I were both pretty adamant about not letting that happen to HIL as well. So I'm all for helping folks automate setup, but I want to be careful to avoid letting the automation mask a process that's fundamentally too complex.

naved001:
>  I very much think it's important to keep "manual" installation within the grasp of mere mortals.

I completely agree with you on this. I might rewrite some part of the docs that I think can be improved.

PR #1028, opened 2018-07-27
Put an upper bound on pytest dependency.

zenhack:
I hit this while digging back in to the OBMd stuff. I suspect it's some problem with namespacing.


zenhack:
*Sigh* looks like there's more bitrot. Will dig into it soon.

zenhack:
Found the problem.

PR #1034, opened 2018-08-31
Move show_console over to obmd

zenhack:
This is the last API call that actually uses the drivertized API support. A subsequent PR will delete the old functionality.

Still futzing with travis; I'm getting an error in the CLI tests that I can't reproduce locally.

### DIFF
--- a/hil/client/node.py
+++ b/hil/client/node.py
@@ -179,8 +179,8 @@ class Node(ClientBase):
         url = self.object_url('node', node, 'console')
         response = self.httpClient.request('GET', url)
         # we don't call check_response here because we want to return the
-        # raw byte stream, rather reading the whole thing in and converting
-        # it to json.
+        # raw byte stream, rather than reading the whole thing in and
+        # converting it to json.
         if 200 <= response.status_code < 300:
             return response.body
         raise FailedAPICallException(error_type=response.status_code,
--- a/hil/client/node.py
+++ b/hil/client/node.py
@@ -168,27 +168,23 @@ class Node(ClientBase):
 
     @check_reserved_chars()
     def show_console(self, node):
-        """Display console log for <node> """
+        """Stream the console for <node>.
+
+        Unlike most of the other API calls, rather than returning a json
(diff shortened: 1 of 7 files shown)


zenhack:
Ok, tests are passing.

naved001:
https://github.com/CCI-MOC/hil/blob/c4057f747bbd956143cc0f4b84c7bd4c60360ac8/hil/model.py#L136
On a related node, how does this line of code even work in our tests. As far as I know we the url to enable the obm (which returns the token) is `obmd_uri/node/<node-name>/token`. But in that method we just do `obmd_uri/path` where path is just token in this case.

This is what I did locally to get it working:
```
         return requests.request(
             method,
-            self.obmd_uri + '/' + path,
+            self.obmd_uri + '/node/' + self.label + path,
             auth=('admin', self.obmd_admin_token),
```
But how are the integration tests passing then

zenhack:
The `obmd_uri` field in the model is the path to *that node* in the obmd api, not the base path to the obmd api. so it includes the `'/node/' + self.label` already. This is per the [spec](https://github.com/CCI-MOC/hil/blob/master/specs/obmd-integration.md).

naved001:
> is the path to that node in the obmd api

ah, I forgot about that bit. I did the registration wrong in that case.

hil/client/node.py
```diff
@@ -168,28 +168,24 @@ def metadata_delete(self, node, label):
...
+        value, this returns an iterator allowing the caller to read the
+        console in a streaming fashion. The iterator's next() method
+        returns the next chunk of data from the console, when it is
+        available.
+        """
         url = self.object_url('node', node, 'console')
         response = self.httpClient.request('GET', url)
         # we don't call check_response here because we want to return the
-        # raw byte stream, rather than converting it to json.
+        # raw byte stream, rather reading the whole thing in and converting
```
ianballou: Rather than *

zenhack: fixed.

Issue #980, opened 2018-03-16
Clarify strategy re: database downgrades


zenhack:
As discussed in #971, alembic technically supports downgrading schema, but while we've made a token effort to write the downgrade functions, we've never really tested it and I wouldn't trust it. We should be more intentional about this, document what usage is actually supported, and maybe actuall rip out the functionality that isn't.

My own inclination is to just not support downgrading at all; it doesn't seem worth the effort.

xuhang57:
this is something that nice to have I believe. 

And if we are using the `flask` CLI, this should be straightforward as
```
flask db downgrade
```

zenhack:
The big thing is figuring out how we're going to go about testing downgrades, and with each upgrade we'd need to think both ways about the change. I worry about the maintenence burden of that and question the utiltiy of the feature, vs. just saying "back up your database if you have any reason to believe you might want to go back.

zenhack:
I'd like to add this to the 0.4 milestone; we've got plans to have a somewhat fiddly upgrade process because of OBMd, so it would be good to have clear docs on what is supported here.

naved001:
Sure.

From your [comment](https://github.com/CCI-MOC/hil/pull/953#issuecomment-365356309)
The second part of the migration strategy will be a part of the 0.4 release. I think we should keep the 0.4 release small just so that we can have support for OBMd as soon as possible. So the main focus of the 0.4 release will be OBMd (and some little things)

I am thinking of removing the 0.4 milestone from issues that aren't trivial.

zenhack:
Sounds good to me; we can defer things to 0.5

Quoting Naved Ansari (2018-04-03 15:29:05)

zenhack:
My thinking is for 0.4 we just write up a quick blurb saying that downgrade isn't really tested or supported and leave it at that. We can consider outright removing the functionality later (maybe before the next time somebody writes a migration script).

PR #1036, opened 2018-09-06
Delete old OBM support.

zenhack:
Still need to update some of the docs, but submitting this. I'm copying @naved001's checklist from #1034 here:

- [x] Remove all old code
- [x] Update the API docs
- [x] Update the config files (sample, and dev-no-hardware).


zenhack:
Docs are updated. Still working out a few kinks, but this is probably ready for review.

naved001:
I guess we should delete the deployment tests for ipmi `tests/deployment/ipmi.py`. Other than that I think you got rid of everything else. I'll do one more pass again.

hil/migrations/versions/02f7e9607e16_delete_legacy_obm_support.py
```diff
@@ -0,0 +1,49 @@
...
+    for driver_table in 'mock_obm', 'ipmi':
+        if engine.dialect.has_table(engine, driver_table):
+            op.drop_table(driver_table)
+
+    op.drop_constraint(u'node_obm_id_fkey', 'node', type_='foreignkey')
+    op.drop_column('node', 'obm_id')
+    op.drop_table('obm')
+
+
+def downgrade():
```
naved001: so we won't officially support downgrading but we'll just let it be here for someone that might need it?

zenhack: I thinks falls under the heading of #980; this is sortof status quo of writing the downgrade function but not really supporting it.

zenhack:
@naved001, I removed the ipmi deployment tests. From the look of them they should have been breaking for a while now (we never added the extra parameters needed to node_register). Might be a good idea to just periodically run the deployment tests to catch bitrot. We normally only run them when touching obviously pertinent functionality, but probably a good idea to do a quick sanity check every now and then.

I'll try to find time to troubleshoot Travis soonish.

tests/integration/client_integration.py
```diff
@@ -1,7 +1,7 @@
 """Integration tests for client library"""
 from hil.flaskapp import app
 from hil.client.base import ClientBase, FailedAPICallException
-from hil.errors import BadArgumentError, UnknownSubtypeError
+from hil.errors import BadArgumentError
```
naved001: looks like we don't use UnknownSubtypeError anywhere now. Let's just delete it while we are at it.

zenhack:
Done.

Quoting Naved Ansari (2018-09-13 09:49:21)

zenhack:
I believe I've fixed the travis issue.

PR #1037, opened 2018-09-17
Add get power status

### LINKED ISSUE #724: Adding a get_node_status to the OBM driver
To make it easy for integrating HIL with provisioning systems like Ironic, MaaS etc we need to add 
a call `get_node_status` to the OBM driver that queries the current state of the node (powered on/powered off/Booted )and reports it as it is.

naved001:
I was looking at stuff that needs to be in the 0.4 release, and noticed that get power_status wasn't in the code. It was pretty straightforward to do, so I just did it. Fixes #724 

Also update the stop_orphan_console method in hil/server.py

### DIFF
--- a/hil/cli/node.py
+++ b/hil/cli/node.py
@@ -157,6 +157,13 @@ def node_power_cycle(node):
     client.node.power_cycle(node)
 
 
+@node_power.command(name='status')
+@click.argument('node')
+def node_power_status(node):
+    """Power cycle <node>"""
+    print client.node.power_status(node)
+
+
 @node.group(name='metadata')
 def node_metadata():
     """Node metadata commands"""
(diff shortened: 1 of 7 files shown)


hil/cli/node.py
```diff
@@ -157,6 +157,13 @@ def node_power_cycle(node):
     client.node.power_cycle(node)
 
 
+@node_power.command(name='status')
+@click.argument('node')
+def node_power_status(node):
+    """Returns node power status"""
+    print client.node.power_status(node)
```
zenhack: Let's put parens around the print() call, for python 3 compat (at some point we should actually make a push on improving that situation, but let's not make it worse).

naved001: cool, done.

PR #1038, opened 2018-09-19
CLI Changes

RobinKaul:
Made a bunch of changes to the CLI output. Main changes are to the network and node outputs.

- Made a json flag for raw outputs for node and network.

- Both the network and node outputs are in a tabular format.

- Need opinion on the node output especially. There are 3 tables: Main, Nic and Metadata. Do I keep the NIC output the way it is now? Or do I just club them into a single table with a blank row for every NIC? Or create a separate table for every NIC?


RobinKaul:
I forgot to mention I have commented out a section of network show and node show as I was experimenting. I kept it deliberately since I was kind of in a dilemma if we would be adding more rows in the future. There is no straightforward way to sort rows in prettytable so I opted to just put in if statements for the particular order. So basically that section would print out the table by fetching data from the raw dictionary.

naved001:
Can you actually paste the sample outputs in a comment here (in a code block so that it renders correctly)?

naved001:
closing this in favor of #1040

PR #1041, opened 2018-09-28
Nic delete

### LINKED ISSUE #1039: Unable to delete a nic after a networking action on it
I can't seem to delete a nic on a node.
```
[naved@kzn-vbmi01res ~]$ hil node nic delete neu-3-9 nic1
[naved@kzn-vbmi01res ~]$ hil node nic delete neu-3-9 nic2
^[[A^[[A[naved@kzn-vbmi01res ~]$ hil node show neu-3-9
project	  :  None
nics	  :  [{u'port': None, u'switch': None, u'macaddr': u'00:0f:53:0d:f7:29', u'networks': {}, u'label': u'nic2'}, {u'port': None, u'switch': None, u'macaddr': u'00:0f:53:0d:f7:28', u'networks': {}, u'label': u'nic1'}]
name	  :  neu-3-9
metadata	  :  {}

```
Curl returns a 500 error. Before running this. 

I then repeated this in my development environment. All you have to do is, register a nic, run a networking operation on it (connect to a network, or run revert port). And then try to delete it.
```
IntegrityError: (sqlite3.IntegrityError) NOT NULL constraint failed: networking_action.nic_id [SQL: u'UPDATE networking_action SET nic_id=? WHERE networking_action.id = ?'] [parameters: (None, 2)] (Background on this error at: http://sqlalche.me/e/gkpj)
INFO:werkzeug:127.0.0.1 - - [19/Sep/2018 17:19:30] "DELETE /v0/node/dell1/nic/nic1 HTTP/1.1" 500 -
```
Probably because we create an entry in the neworking_action table for querying later so we can't just delete the nic because the networking_action table refers to that nic_id.

It was easily fixed by clearing my networking_action table.

naved001:
Fixes #1039 

There was one more issue with nic_delete. I added a check to make sure that we remove any networks before we are allowed to delete a nic. 

One question, administrators can delete a nic on a node that belongs to some project. Should that be allowed? Our usual policy is to disallow deletion of something that is owned by a project (networks, nodes).

### DIFF
--- a/hil/api.py
+++ b/hil/api.py
@@ -360,10 +360,13 @@ def node_delete_nic(node, nic):
     """
 
     get_auth_backend().require_admin()
-    nic = get_child_or_404(get_or_404(model.Node, node), model.Nic, nic)
-    check_pending_action(nic)
-    if nic.attachments:
-        raise errors.BlockedError("Nic is connected to a network")
+    node = get_or_404(model.Node, node)
+    nic = get_child_or_404(node, model.Nic, nic)
+    if node.project:
+        raise errors.BlockedError("Nic is on a node that belongs to a project."
+                                  " Remove node from project and try again")
+    if nic.current_action:
+        db.session.delete(nic.current_action)
     db.session.delete(nic)
     db.session.commit()
 
--- a/hil/api.py
+++ b/hil/api.py
@@ -355,9 +355,11 @@ def node_delete_nic(node, nic):
     """Delete nic with given name from it's node.
 
     If the node or nic does not exist, a NotFoundError will be raised.
+    If there's a pending networking action then a BlockedError will be raised.
     """
     get_auth_backend().require_admin()
     nic = get_child_or_404(get_or_404(model.Node, node), model.Nic, nic)
+    check_pending_action(nic)
     db.session.delete(nic)
     db.session.commit()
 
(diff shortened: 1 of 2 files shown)


zenhack:
Yeah, I think we should refuse to delete a nic on a node that is allocated to a project; if an admin wants to remove a nic, they need to free it first.

naved001:
cool, done.

hil/api.py
```diff
@@ -360,10 +360,13 @@ def node_delete_nic(node, nic):
...
-    check_pending_action(nic)
-    if nic.attachments:
-        raise errors.BlockedError("Nic is connected to a network")
+    node = get_or_404(model.Node, node)
+    nic = get_child_or_404(node, model.Nic, nic)
+    if node.project:
+        raise errors.BlockedError("Nic is on a node that belongs to a project."
+                                  " Remove node from project and try again")
+    if nic.current_action:
+        db.session.delete(nic.current_action)
```
zenhack: We seem to have lost the `PENDING` check?

naved001: We don't need it here anymore, since we need to first detach the node from the project which checks for any pending actions an all nics for that node.

naved001: https://github.com/CCI-MOC/hil/blob/master/hil/api.py#L123-L126

zenhack: Ah, you're right. Ok, fix the docstring above and I'm happy.

naved001: done.

PR #1043, opened 2018-10-05
Add more info in the release notes for v0.4

naved001:


naved001:
@zenhack feel free to push directly to this branch if you think we need to add something more. I'll make the release once this is merged.

docs/UPGRADING.rst
```diff
@@ -55,6 +55,9 @@ Release Notes
   3. Update the HIL software as usual. Before restarting the serivces, remove
      both of the above obm extensions from ``hil.cfg``.
 
+* OpenVSwitch is now supported in HIL.
```
zenhack: I might be more explicit here that the new driver is very much not intended for production use. (I might not even mention it, since it's intended for development).

naved001:
cool, merging this then.

Issue #925, opened 2017-12-15
Make our error types more clear and useful


naved001:
This stems from the discussion [here](https://github.com/CCI-MOC/hil/pull/920#issuecomment-352084439)

Errors we raise in our APIs should be distinct and clear enough so that we know how to process those errors just by looking at the type of it.

zenhack:
Would be good to accumulate a list of cases that need dealing with. The one you mentioned:

* `project_detach_node` raises `BlockedError` if any resource is still attached to the node, but this includes `NetworkingAction`s, where the appropriate behavior is to just wait, whereas with other resources the right behavior is to detach them.

Which also brings up the issue: in the latter case, the client might want to know *what* needs to be detached.

I can think of two general approaches off the top of my head:

* Make more error types, and use different ones for different things. It might make sense to e.g. split `BlockedError` into `AttachedResourceError` and `PendingActionError`, which would solve the specific case you brought up.
* Add more data to the error types. Running with the example above, `AttachedResourceError` could have some extra fields designating the type and identity of the attached resource.

It probably makes sense to combine these.

naved001:
>  It might make sense to e.g. split 

@jeremyfreudberg suggested that we inherit from `BlockedError`. That way we could either catch the specific child exception, or the parent one if we don't care about the exact type.

zenhack:
It's not obvious to me that it makes sense to have these two use cases have a shared ancestor. Can you think of a case where we'd want to handle BlockedError but not care about the difference above?

Also, worth noting that the class hierarchy doesn't get sent across the wire.

naved001:
> Can you think of a case where we'd want to handle BlockedError but not care about the difference above?

Can't think of it now. Maybe in future, but in that case we might as well not do it.

> Also, worth noting that the class hierarchy doesn't get sent across the wire.

Not sure what you mean by that? This code seems to work. 

```python
class ParentError(Exception):
    """Some Parent Error"""


class ChildError1(ParentError):
    """ChildError1"""


class ChildError2(ParentError):
    """different one"""


def raise_error1():
    raise ChildError1('Child1')


def raise_error2():
    raise ChildError2('Child2')

try:
    raise_error1()
except ParentError:
    # this will be printed
    print "Exception handled"
```

zenhack:
Right, but the server sends them to the client as JSON, where the type is just a string, and it doesn't include information about what the superclasses are.

naved001:
[...]

PR #1044, opened 2018-10-10
Split BlockedError to AttachedResourceError and PendingActionError

RobinKaul:
I'm referring to issue #925 . Since there were two suggestions I am referring to the first one that was to replace ```BlockedError``` with a ```AttachedResourceError``` and ```PendingActionError```. Also I noticed that in a lot of places the two new error types wouldn't fit into that particular case so I changed them to ```OBMError```/ ```DuplicateError```, etc. Please let me know if they need to be reverted. 
Also I was kind of swaying between a couple of the error names like when a node/ network is not free. I wasn't sure if they need to be named ```PendingActionError``` or ```AttachedResourceError``` so I named them the latter for now.
One little thing I wanted to point out was that in the comments of the ```node_detach_network``` function in api.py, it mentioned it raises an error if there is a pending network action but then the error isn't raised.

### DIFF
--- a/hil/api.py
+++ b/hil/api.py
@@ -58,9 +58,9 @@ def project_delete(project):
     get_auth_backend().require_admin()
     project = get_or_404(model.Project, project)
     if project.nodes:
-        raise errors.BlockedError("Project has nodes still")
+        raise errors.AttachedResourceError("Project has nodes still")
     if project.networks_created:
-        raise errors.BlockedError("Project still has networks")
+        raise errors.AttachedResourceError("Project still has networks")
     if project.networks_access:
         # FIXME: This is not the user's fault, and they cannot fix it.  The
         # only reason we need to error here is that, with how network access
@@ -71,9 +71,9 @@ def project_delete(project):
         # network accessible to ALL PROJECTS!  Once we use real ACLs, this
         # will not be an issue---instead, the network will be accessible by
         # NO projects.
-        raise errors.BlockedError("Project can still access networks")
+        raise errors.AttachedResourceError("Project can still access networks")
     if project.headnodes:
-        raise errors.BlockedError("Project still has a headnode")
+        raise errors.AttachedResourceError("Project still has a headnode")
     db.session.delete(project)
     db.session.commit()
 
@@ -86,13 +86,13 @@ def project_connect_node(project, node):
 
     If the node or project does not exist, a NotFoundError will be raised.
 
-    If node is already owned by a project, a BlockedError will be raised.
+    If node is already owned by a project, a AttachedResourceError will be raised.
     """
     project = get_or_404(model.Project, project)
     get_auth_backend().require_project_access(project)
     node = get_or_404(model.Node, node)
     if node.project is not None:
-        raise errors.BlockedError("Node is already owned by a project.")
+        raise errors.AttachedResourceError("Node is already owned by a project.")
     project.nodes.append(node)
     db.session.commit()
 
@@ -105,8 +105,9 @@ def project_detach_node(project, node):
 
     If the node or project does not exist, a NotFoundError will be raised.
 
-    If the node has network attachments or pending network actions, a
-    BlockedError will be raised.
+    If the node has network attachments or pending network actions, an
+    AttachedResourceError or a PendingActionError will be raised
+    respectively.
     """
     project = get_or_404(model.Project, project)
     get_auth_backend().require_project_access(project)
@@ -114,16 +115,16 @@ def project_detach_node(project, node):
     if node not in project.nodes:
         raise errors.NotFoundError("Node not in project")
     if node.obm_is_enabled():
-        raise errors.BlockedError("Node's obm is enabled.")
+        raise errors.OBMError("Node's obm is enabled.")
     num_attachments = model.NetworkAttachment.query \
         .filter(model.Nic.owner == node,
                 model.NetworkAttachment.nic_id == model.Nic.id).count()
     if num_attachments != 0:
-        raise errors.BlockedError("Node attached to a network")
+        raise errors.AttachedResourceError("Node attached to a network")
     for nic in node.nics:
         if nic.current_action is not None and \
            nic.current_action.status == 'PENDING':
-            raise errors.BlockedError("Node has pending network actions")
+            raise errors.PendingActionError("Node has pending network actions")
 
     project.nodes.remove(node)
     _maintain(project, node, node.label)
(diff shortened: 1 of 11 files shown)


hil/api.py
```diff
@@ -162,7 +163,7 @@ def network_revoke_project_access(project, network):
     """Remove access to <network> from <project>.
 
     If the project or network does not exist, a NotFoundError will be raised.
-    If the project is the owner of the network a BlockedError will be raised.
+    If the project is the owner of the network a AttachedResourceError will be raised.
```
zenhack: This seems conceptually wrong to me; this issue isn't that something needs detaching, but that something fundamentally *cannot* be detached.

hil/api.py
```diff
@@ -319,11 +320,11 @@ def node_delete(node):
     get_auth_backend().require_admin()
     node = get_or_404(model.Node, node)
     if node.project:
-        raise errors.BlockedError(
+        raise errors.Error(
```
zenhack: I assume this was supposed to be `AttachedResourceError`.

hil/api.py
```diff
@@ -355,15 +356,15 @@ def node_delete_nic(node, nic):
     """Delete nic with given name from it's node.
 
     If the node or nic does not exist, a NotFoundError will be raised.
-    If the nic is on a node that belongs to a project then a BlockedError will
+    If the nic is on a node that belongs to a project then a Error will
```
zenhack: Again I assume `AttachedResourceError`.

hil/api.py
RobinKaul: Yes exactly that's why I said some of the errors didn't quite fit into it so I named them ```AttachedResourceError``` for now. I think I wrote ```AuthorizationError``` previously for this but then had to redo all changes for some reason so I might've put the former. What do you think about ```AuthorizationError``` though?

hil/api.py
RobinKaul: I don't know how I skipped these two this but I've corrected them now. Will push in the future commits.

hil/api.py
zenhack: `AuthorizationError` isn't right either; it isn't a permissions issue. I think `BadArgumentError` is the closest thing we currently have. I'm not terribly opposed to adding a new type for this sort of thing.

@naved001, @ianballou, interested in your thoughts.

RobinKaul:
@zenhack quick followup about mentioning the new error types in ```UPGRADING.rst```, has someone already posted a change for 0.5 and I can add to that? Or do I create a new PR for version 0.5?
Side note: the formatting for the steps in version 0.4 is wonky. I will correct that too.

zenhack:
There's no entry for 0.5 yet, because this will be the first user-visible change we've merged so far (@naved001 tagged 0.4 just the other day). So you'll have to add a new section to the document.

hil/api.py
naved001: I agree with @zenhack, `BadArgumentError` is the closest thing we have. This is the only case we have right now where we have the concept of ACLs. So I think `BadArgumentError` with a message explaining why the arguments are bad (cannot revoke access from project since project is the owner) is good enough.
If we were to create a new type we could call it `IllegalOperationError` since it's not legal to for a project to not have access to the network it created. And I think it would be generic enough to be used in similar situations that we may run into.

ianballou: Yeah I agree that `AttachedResourceError` doesn't work here. That error should alert the user that they need to detach something first before continuing with the command that raised the error.  Here, there's nothing that needs detaching.

I agree with @naved001 that something like `IllegalOperationError` would be good.  That plus a good error string stating why it's illegal.

zenhack: Either of those solutions is fine with me.

hil/api.py
```diff
@@ -58,9 +58,9 @@ def project_delete(project):
     get_auth_backend().require_admin()
     project = get_or_404(model.Project, project)
     if project.nodes:
-        raise errors.BlockedError("Project has nodes still")
+        raise errors.AttachedResourceError("Project has nodes still")
     if project.networks_created:
-        raise errors.BlockedError("Project still has networks")
+        raise errors.AttachedResourceError("Project still has networks")
```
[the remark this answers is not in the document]
zenhack: This seems like exactly what `AttachedResourceError` is for, what am I missing?

hil/api.py
```diff
@@ -424,14 +425,14 @@ def _have_attachment(nic, query):
...
-        raise errors.BlockedError(
+        raise errors.AttachedResourceError(
             "The network is already attached to the nic.")
 
     if channel is None:
         channel = allocator.get_default_channel()
 
     if _have_attachment(nic, model.NetworkAttachment.channel == channel):
-        raise errors.BlockedError("The channel is already in use on the nic.")
+        raise errors.PendingActionError("The channel is already in use on the nic.")
```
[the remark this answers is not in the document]
zenhack: Maybe it's worth distinguishing a third case, `AlreadyInUseError`, which could be used here and also in e.g. `project_connect_node`. `AttachedResourceError` means "you release that resource because things need to be detached first", whereas `AlreadyInUseError` means "You can't acquire this resource because it's already in use." I kinda don't like `BlockedError` because it's rather vague.

hil/api.py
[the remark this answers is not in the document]
naved001: Per your comment [here](https://github.com/CCI-MOC/hil/pull/1044#discussion_r224167833), isn't this the same scenario, where we can't detach this resource to get past this error? (In this case, we would have to delete the resource).

hil/api.py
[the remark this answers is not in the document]
naved001: yeah, I like `AlreadyInUseError`.

hil/api.py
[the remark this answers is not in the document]
zenhack: This seems more analagous to `node_delete`.

hil/api.py
RobinKaul: Okay I'm a little confused here. We already have ```IllegalStateError``` to indicate the "generic" illegal operation. Is it serving any other purpose too? Also, according to the description, it is essentially doing the same thing as ```AttachedResourceError``` unless I'm missing out on something. Should we rename it to ```IllegalOperationError``` and give it a proper definition?

zenhack: `IllegalStateError` isn't quite just any generic illegal op. Right now I think it's only used with headnodes, in cases where somebody does something like try to modify the configuration of the VM while it's running -- it is in an "illegal state." The user could get things working by shutting the VM off first. Whereas with this case, removing the owner of a network is simply not allowed; there's nothing the user can do to make the operation legal.

PR #1045, opened 2018-10-15
Add information about how to run hil and apache with TLS

naved001:
This one needs more scrutiny since the httpd configuration file I have was generated by Certbot, I just added the wsgi bits to it to make it work. Though, my test setup got an A grade here https://www.ssllabs.com/ssltest/analyze.html?d=kzn-hiltest.massopen.cloud&hideResults=on

Other thing is that, I didn't have to specify the load wsgi module option in the conf file and it still works.


zenhack:
I don't know that I think it makes sense to include this in as much detail; setting up TLS with apache is a standard thing that's pretty orthogonal to HIL. Maybe just mention that TLS should be used, and link to relevant documentation elsewhere?

naved001:
Yeah, I see what you mean.

Issue #1049, opened 2018-10-18
CLI client: error message output is not useful if the body of the response does not describe the error.


zenhack:
I ran into this while troubleshooting the ipmi timeout issue that @naved001, brought up: if the HTTP server returns a non-200 status, we raise a FailedAPICallException with the message being the body of the response. In my case, I was getting a 401 with no body, so the command line tool just reported `Error:`, with nothing after that...

we should make it a bit smarter about good error messages.

PR #1050, opened 2018-10-19
Just print whatever the exception is

naved001:
For #1049

### DIFF
--- a/hil/cli/cli.py
+++ b/hil/cli/cli.py
@@ -3,7 +3,6 @@ import click
 import sys
 import pkg_resources
 
-from hil.client.base import FailedAPICallException
 from hil.commands.util import ensure_not_root
 from hil.cli import node, project, network, switch, port, user, misc, headnode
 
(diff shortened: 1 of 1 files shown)


zenhack:
I don't think this actually solves the problem -- we'll end up printing a stack trace, but it still doesn't include any extra info -- besides e.message, the FailedAPICallException only includes an `error_type` field, which is not printed.

Probably we should at least try to fit the HTTP status code in there somewhere.

naved001:
yeah, I don't know why I thought this would work. Also, when check_reserved raises an error it is not handled at all.
I should just not handle all exceptions (except for the ones we know) and let there be a stack trace, so we don't accidentally hide something.

zenhack:
Quoting Naved Ansari (2018-11-02 14:41:15)

>    I should just not handle all exceptions (except for the ones we know)
>    and let there be a stack trace, so we don't accidentally hide
>    something.

That too, though it's a conceptually separate issue from this one.

PR #1053, opened 2018-11-09
Cli pretty 2

### LINKED ISSUE #1047: cli client: list nodes one per line
The output of `hil node list` (either free or all) shows all of the nodes on one line. This isn't terribly readable in a setup where you have a significant number of machines in the pool. Might be better to do one per line.

naved001:
This makes the output of all other HIL commands pretty.
Fix #1047

### DIFF
--- a/hil/cli/helper.py
+++ b/hil/cli/helper.py
@@ -10,7 +10,7 @@ def print_json(raw_output):
     sys.exit(0)
 
 
-def make_table(field_names=[], rows=[]):
+def make_table(field_names, rows):
     """Generate a PrettyTable and return it.
     If there's only field, then it will add the count of items in the header.
     """
--- a/hil/cli/helper.py
+++ b/hil/cli/helper.py
@@ -7,3 +7,13 @@ def print_json(raw_output):
     """Format raw_output as json, print it and exit"""
     print(json.dumps(raw_output))
     sys.exit(0)
+
+def gather_output(items, separator='\n'):
+	"""Takes a list of <items> and returns a single string
+	separated by <separator>
+	"""
+	returnee = ''
+	for item in items:
+		returnee += item + separator
+
+	return returnee
--- a/hil/cli/helper.py
+++ b/hil/cli/helper.py
@@ -14,11 +14,7 @@ def gather_output(items, separator='\n'):
     """Takes a list of <items> and returns a single string
     separated by <separator>
     """
-    returnee = ''
-    for item in items:
-        returnee += item + separator
-
-    return returnee
+    return separator.join(items) + separator
 
 
 def print_status_table(raw_output):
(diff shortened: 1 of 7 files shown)


hil/cli/network.py
```diff
@@ -34,30 +35,81 @@ def network_delete(network):
...
+
+    # Make the table and print it
+    network_show_table = PrettyTable()
+    network_show_table.field_names = ['Field', 'Value']
+    network_show_table.add_row(['Name', raw_output['name']])
+    network_show_table.add_row(['Owner', raw_output['owner']])
+    network_show_table.add_row(['Channels', channels])
+    network_show_table.add_row(['Access', projects])
+    network_show_table.add_row(['Connected Nodes', connected_nodes.rstrip()])
+    print(network_show_table)
```
zenhack: I'm finding this kind of imperative construction less readable than it could be; the comment at the top of this block is a bit of a code smell. With an appropriate helper function that would let us do:

```python
print(make_table(
    field_names=['Field', 'Value'],
    rows=[
        ['Name', raw_output['name']],
        ['Owner', raw_output['owner']],
        # etc
    ]))
```

For the cases where we're currently building up the list of rows in a for loop, we could use a list comprehension.

I don't think it would be too hard to generalize the helpers you've already written to cover this.

ianballou: I think this is a cool idea too.  That way, the tables can all be generalized for future use, ensuring that our CLI looks uniform.

naved001: Done. Let me know what you guys think.

hil/cli/network.py
```diff
@@ -103,13 +96,12 @@ def list_network_attachments(network, project, jsonout):
     if jsonout:
         print_json(raw_output)
 
-    attachments_table = PrettyTable()
-    attachments_table.field_names = ['Node', 'Project', 'Nic', 'Channel']
-    for node, attributes in raw_output.iteritems():
-        attachments_table.add_row([node, attributes['project'],
-                                   attributes['nic'], attributes['channel']])
+    rows = [[node, attributes['project'],
+             attributes['nic'], attributes['channel']]
```
zenhack: This scans really poorly; I'd say put each list element on its own line.

hil/cli/helper.py
```diff
@@ -10,24 +10,14 @@ def print_json(raw_output):
...
-    status_table.field_names = ['Field', 'Value']
-    status_table.add_row(['Status ID', raw_output['status_id']])
-    print(status_table)
-    sys.exit(0)
-
-
-def print_list_table(raw_output, header):
-    """Prints a table with a single column. Also, puts the the counts
-    of items in the column name.
+def make_table(field_names=[], rows=[]):
```
zenhack: General rule of thumb: never use mutable objects as defaults for parameters. I don't think it's actually problematic in this case, but this sets off my spidey senses and I'd rather just write code that obviously doesn't have an issue.

FTR, the problem with using a list as the default arises in cases like this:


```python
def append_and_print(lst=[]):
    lst.append(1)
    print(lst)

>>> append_and_print()
[1]
>>> append_and_print()
[1, 1]
>>> append_and_print()
[1, 1, 1]
>>> append_and_print()
[1, 1, 1, 1]
...
```

The value is evaluated at the time of definition, so if you make modifications to it, the next call to the function uses the modified version...

I think an empty tuple would work in this case.

naved001: > The value is evaluated at the time of definition, so if you make modifications to it, the next call to the function uses the modified version...

oh! thanks for letting me know about this. 

> I think an empty tuple would work in this case.

I don't understand. Where would I use an empty tuple? I can just not use defaults in this case and it will work, right?
```python
def make_table(filed_names, rows):
```

zenhack:
LGTM

ianballou:
I'm reading through this, I'll finish up my review next Tuesday

PR #1056, opened 2019-01-10
Cli macros

naved001:
This make administering the cluster a little easier when I have to forcefully take nodes from users. I have made it admin only at the moment. Might make sense to let regular users run it too.

### DIFF
--- a/hil/cli/helper.py
+++ b/hil/cli/helper.py
@@ -7,8 +7,7 @@ from prettytable import PrettyTable
 try:
     HIL_TIMEOUT = int(os.getenv('HIL_TIMEOUT', 10))
 except ValueError:
-    print("HIL_TIMEOUT is not a number")
-    sys.exit(1)
+    sys.exit("Please set environment variable HIL_TIMEOUT to a number")
 
 
 def print_json(raw_output):
--- a/hil/cli/helper.py
+++ b/hil/cli/helper.py
@@ -1,7 +1,14 @@
 """Helper functions for the CLI live here"""
 import json
 import sys
+import os
 from prettytable import PrettyTable
+
+try:
+    HIL_TIMEOUT = int(os.getenv('HIL_TIMEOUT', 10))
+except ValueError:
+    print("HIL_TIMEOUT is not a number")
+    sys.exit(1)
 
 
 def print_json(raw_output):
--- a/hil/cli/project.py
+++ b/hil/cli/project.py
@@ -1,8 +1,8 @@
 """Commands related to projects are in this module"""
 import click
+import time
 from hil.cli.client_setup import client
-from hil.cli.helper import print_json, make_table
-
+from hil.cli.helper import print_json, make_table, HIL_TIMEOUT
 
 @click.group()
 def project():
@@ -85,13 +85,33 @@ def project_detach_node(project, node, force):
     """Remove <node> from <project>"""
     if force:
         client.node.disable_obm(node)
+        print("Disabled OBM")
 
         node_info = client.node.show(node)
+        statuses = []
         if 'nics' in node_info:
             for n in node_info['nics']:
                 # if any nic is connected to a network, then it has a switch
                 # and port. So call revert port on those.
                 if 'networks' in n:
-                    client.port.detach_nic(n['switch'], n['port'])
+                    response = client.port.port_revert(n['switch'], n['port'])
+                    statuses.append(response['status_id'])
+
+        done = False
+        end_time = time.time() + HIL_TIMEOUT
+        while not done:
+            if time.time() > end_time:
+                raise Exception("Removing networks is taking too long. "
+                                "Set HIL_TIMEOUT to wait longer")
+            done = True
+            for status_id in statuses:
+                response = client.node.show_networking_action(status_id)
+                if response['status'] == 'ERROR':
+                    print("Revert port failed. Check HIL network server")
+                    return
+                elif response['status'] == 'PENDING':
+                    done = False
+                    break
+        print("Succesfuly finished removing all networks from the node.")
 
     client.project.detach(project, node)
--- a/hil/cli/project.py
+++ b/hil/cli/project.py
@@ -92,6 +92,7 @@ def project_detach_node(project, node, force):
         print("Disabled OBM")
         node_info = client.node.show(node)
         statuses = []
+        done = True
         if 'nics' in node_info:
             for n in node_info['nics']:
                 # if any nic is connected to a network, then it has a switch
@@ -101,6 +102,7 @@ def project_detach_node(project, node, force):
                         response = client.port.port_revert(
                             n['switch'], n['port'])
                         statuses.append(response['status_id'])
+                        done = False
                     except KeyError:
                         # This is raised when we try to look for switch and
                         # and port in the output of node.show; which is only
@@ -109,7 +111,6 @@ def project_detach_node(project, node, force):
                                  " to call revert port with. Try running this"
                                  " as a HIL admin.")
 
-        done = False
         end_time = time.time() + HIL_TIMEOUT
         while not done:
             if time.time() > end_time:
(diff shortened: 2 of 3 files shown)


hil/cli/helper.py
```diff
@@ -1,8 +1,15 @@
...
 import json
 import sys
+import os
 from prettytable import PrettyTable
 
+try:
+    HIL_TIMEOUT = int(os.getenv('HIL_TIMEOUT', 10))
+except ValueError:
+    print("HIL_TIMEOUT is not a number")
+    sys.exit(1)
```
zenhack: Maybe just `sys.exit('HIL_TIMEOUT ...')`

hil/cli/project.py
```diff
@@ -80,6 +82,42 @@ def project_connect_node(project, node):
...
+                # if any nic is connected to a network, then it has a switch
+                # and port. So call revert port on those.
+                if 'networks' in n:
+                    try:
+                        response = client.port.port_revert(
+                                                    n['switch'], n['port'])
+                        statuses.append(response['status_id'])
+                    except KeyError:
+                        print("Only admins can force remove nodes.")
+                        sys.exit(1)
```
zenhack: I'm not a huge fan of this; at the very least we should have a comment here explaining why a `KeyError` suggests admin-only, but in a case like this where we're doing inference as to what actually went wrong, I'd rather have the error message reflect what actually happend, and maybe some guesses -- it could get easy for something else to unexpectedly trigger this and have a user be very confused about the admin bit, since it's not actually an auth check that's triggering it.

naved001: okay, I have put in a comment and the exit message has more details.

hil/cli/helper.py
```diff
@@ -1,8 +1,15 @@
 """Helper functions for the CLI live here"""
 import json
 import sys
+import os
 from prettytable import PrettyTable
 
+try:
+    HIL_TIMEOUT = int(os.getenv('HIL_TIMEOUT', 10))
+except ValueError:
+    print("HIL_TIMEOUT is not a number")
```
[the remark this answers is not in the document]
naved001: > Especially if they get it and haven't set HIL_TIMEOUT to anything yet. 

If they don't set HIL_TIMEOUT to anything then it will default to 10 seconds and no error is raised. The message should be printed only when they explicitly set it to something non-integer.

hil/cli/project.py
```diff
@@ -80,6 +82,42 @@ def project_connect_node(project, node):
...
+
+        done = False
+        end_time = time.time() + HIL_TIMEOUT
+        while not done:
+            if time.time() > end_time:
+                raise Exception("Removing networks is taking too long. "
```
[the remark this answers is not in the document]
naved001: I was worried about the case where, let's say, the networking server is not responding, so the action will always be pending. The user wouldn't know why this call is blocked since it will be looping forever; which is why I thought of setting a time limit.

naved001:
> In addition, we should add some user-level documentation about the flag (and the environment variable).

I have put the description for the flag in the CLI itself. Any suggestions for where to document the environment variable? Maybe I could start a document for the CLI since it's growing in size.

Also, looks like the CI is having some other issues with pytest not being happy with xdist plugin...

naved001:
> pytest-cov 2.6.1 has requirement pytest>=3.6, but you'll have pytest 3.5.1 which is incompatible.
pytest-xdist 1.26.0 has requirement pytest>=3.6.0, but you'll have pytest 3.5.1 which is incompatible.

Ah, removing the upper bound on pytest from our setup.py will install version 4.1.1 which works with those plugins, but requires us to not call any fixtures directly.
```
[gw0] linux2 -- Python 2.7.15 /home/naved/hil-all/cli-macros/.venv/bin/python
Fixture "dbauth" called directly. Fixtures are not meant to be called directly,
but are created automatically when test functions request them as parameters.
See https://docs.pytest.org/en/latest/fixture.html for more information about fixtures, and
https://docs.pytest.org/en/latest/deprecations.html#calling-fixtures-directly about how to update your code.
```
I guess time to fix all that?

hil/cli/project.py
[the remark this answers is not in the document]
zenhack: I agree that just having this as a fixed timeout to start is the right thing. If we find it's too short we can make it longer. We could make it configurable, but it's not obvious to me that there are likely use cases for tuning it; let's wait until they actually arise.

zenhack:
Re: pytest, is there a reason we specified `<3.6` in the first place? If they're following semver it should be safe to switch to `<4.0`, and that actually seems like The Right Thing -- if it works I'd say do that for now, and maybe open another issue/pr about updating our dependencies.

It might be good to think about process for bumping these in general; we've kinda only been doing it when something breaks.

naved001:
> is there a reason we specified <3.6 in the first place? 

https://github.com/CCI-MOC/hil/pull/1028/commits/fd68983cad3e5b6adedeba1ef3c531890f140313
 this explains it.

zenhack:
Can you verify that the problem still exists?

naved001:
Still some problems with the integration tests. I see fixture `initial_db` not found instead of the error you mentioned in your commit message earlier. 
https://travis-ci.org/naved001/hil/jobs/480565324

zenhack:
Ok. Can we just pin pytest-cov to an earlier version for now, so we can keep using the current version of pytest?

Also: re CLI docs, I think the flag is good for this PR, but maybe open another issue about overview docs for the CLI, yeah.

zenhack:
I merged #1058; rebase and travis should fix itself.

hil/cli/project.py
```diff
@@ -80,6 +82,48 @@ def project_connect_node(project, node):
...
+              help="Disables obm and removes all networks before removing "
+                   "node from project. Requires admin privileges.")
+def project_detach_node(project, node, force):
     """Remove <node> from <project>"""
+    if force:
+        client.node.disable_obm(node)
+        print("Disabled OBM")
+        node_info = client.node.show(node)
+        statuses = []
+        if 'nics' in node_info:
```
ianballou: I'm trying to follow the logic here.  So if `'nics' in node_info` (line 95) returns false, then it skips to line 112 and sets `done = False`.  Then next it would go into the logic that checks if port revert finished successfully.  However, `statuses` will always be empty following this scenario, causing that section after line 119 to be skipped.
Maybe instead then we could just skip to line 129 if line 95 is false?

ianballou: I'm trying to follow the logic here.  So if `'nics' in node_info` (line 95) returns false, then it skips to line 112 and sets `done = False`.  Then next it would go into the logic that checks if port revert finished successfully.  However, `statuses` will always be empty following this scenario, causing that section after line 119 to be skipped.
Maybe instead then we could just skip to line 129 if line 95 is false?

naved001:
Good catch, I'll do what you suggested!

hil/cli/project.py
naved001: Done. See the latest commit.

hil/cli/project.py
```diff
@@ -80,6 +82,49 @@ def project_connect_node(project, node):
...
+                # and port. So call revert port on those.
+                if 'networks' in n:
+                    try:
+                        response = client.port.port_revert(
+                            n['switch'], n['port'])
+                        statuses.append(response['status_id'])
+                        done = False
+                    except KeyError:
+                        # This is raised when we try to look for switch and
+                        # and port in the output of node.show; which is only
```
ianballou: Two ands here

ianballou: Two 'and's here in the comment

zenhack:
Travis needs to be fixed, but once that's done I'm happy.

naved001:
I fixed the errors in #1061 since those are unrelated to this PR. I'll rebase this PR once that one is merged.

zenhack:
Sounds good; I've approved both prs.

ianballou:
Just need a rebase and then we can merge.

PR #1058, opened 2019-01-17
Dont use the latest version of pytest plugins pytest-cov and pytest-x…

naved001:
…dist since

these aren't compatible with the version of pytest we are using at the moment.


zenhack:
LGTM.

PR #1060, opened 2019-01-18
Correctly pass one row to make_table

naved001:
otherwise pretty table fails to generate the table.


zenhack:
What happens without this?

Also, can you add a regression test?

Also, a readability note, though not really part of this PR: it's super weird and hacky the way we're using passing 'Field' as part of `field_names`; I'd like to make this a little cleaner. That might be as simple as renaming `field_names` to `column_names`, so it reads as less of an abuse of the internals.

ianballou:
When I did a simple test without the square brackets:
```
print(make_table(field_names=['Field', 'Value'],
                 rows=['Potato', 'corn']))
```
I got an error: `Row has incorrect number of values`.  Adding the extra brackets fixed this.

naved001:
> What happens without this?

I think @ianballou  answered that with his example. It was treating 'Potato' and 'Corn' as distinct rows, so it would think 'P', 'o', 't', 'a', 't', 'o' are 6 values that make up one row and hence the error says 6!=2.

> Also, can you add a regression test?

zenhack:
Yeah, I'd say just do the integration test.

naved001:
took me long enough to fix the tests haha. 

@zenhack @ianballou does this look good now?

PR #1063, opened 2020-11-06
Implement method to shut off a port

naved001:
* This method is called by modify_port when the native vlan is removed.
It is also called by revert_port which resets the switchport state.

Doing this prevents the switchport from sitting on default VLAN 1.

The switchport is only turned on only when we need to configure a vlan. 

Closes: https://github.com/CCI-MOC/hil/issues/970

I think it's safer to turn off the switchport entirely than fiddle with `trunk-no-native-vlan` mode which depends on the NOS version.

### DIFF
--- a/hil/ext/switches/brocade.py
+++ b/hil/ext/switches/brocade.py
@@ -12,7 +12,7 @@ from schema import Schema, Optional
 
 from hil.migrations import paths
 from hil.model import db, Switch
-from hil.errors import BadArgumentError
+from hil.errors import BadArgumentError, SwitchError
 from hil.model import BigIntegerType
 from hil.ext.switches.common import check_native_networks, parse_vlans
 from hil.config import core_schema, string_is_bool
@@ -72,8 +72,23 @@ class Brocade(Switch, _vlan_http.Session):
         return []
 
     def _port_shutdown(self, interface):
-        """Shuts down port; unimplemented right now
-        TODO: Implement this and associated methods see #970"""
+        """Shuts down port"""
+        payload = '<shutdown>true</shutdown'
+        # accepting 409 makes it idempotent
+        self._make_request('POST', url, data=payload, acceptable_error_codes=(409,))
+
+    def _is_interface_off(self, interface):
+        """ Returns a boolean that tells the status of a switchport"""
+
+        url = self._construct_url(interface)
+        response = self._make_request('GET', url)
+        shutdown = root.find(self._construct_tag("shutdown"))
+        if shutdown is None:
+            return False
+        elif shutdown.text == "true":
+            return True
+        else
+            raise SwitchError("Could not determine if interface is off")
 
     def _get_mode(self, interface):
         """ Return the mode of an interface.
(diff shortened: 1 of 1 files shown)


hil/ext/switches/brocade.py
```diff
@@ -72,8 +72,23 @@ def get_capabilities(self):
         return []
 
     def _port_shutdown(self, interface):
-        """Shuts down port; unimplemented right now
-        TODO: Implement this and associated methods see #970"""
+        """Shuts down port"""
+        payload = '<shutdown>true</shutdown'
```
knikolla, line 76: missing a closing `>`?

hil/ext/switches/brocade.py
```diff
@@ -72,8 +72,23 @@ def get_capabilities(self):
...
+        """ Returns a boolean that tells the status of a switchport"""
+
+        url = self._construct_url(interface)
+        response = self._make_request('GET', url)
+        shutdown = root.find(self._construct_tag("shutdown"))
+        if shutdown is None:
+            return False
+        elif shutdown.text == "true":
+            return True
+        else
```
knikolla: missing `:`

hil/ext/switches/brocade.py
naved001: yep, thanks for pointing that out.

naved001:
I'll run the deployment tests before merging this just to be sure everything works (and that the other methods work without having to turn on the switchport).

knikolla (approved):

hil/ext/switches/brocade.py
```diff
@@ -100,10 +116,14 @@ def _enable_and_set_mode(self, interface, mode):
             mode: 'access' or 'trunk'
 
         Raises: AssertionError if mode is invalid.
-
         """
-        # Enable switching
         url = self._construct_url(interface)
+
+        # Turn on port; equivalent to running `no shutdown` on the switchport
+        if self._is_interface_off:
```
naved001, line 123: mistake here. didn't call the method correctly and this was calling the deployment tests to fail.

naved001:
With the new commit the deployment tests pass.
```
(.venv) naved-work::~/hil-all/brocade-default-vlan-fix$py.test tests/deployment/*networks.py
=========================================================== test session starts ============================================================
platform linux2 -- Python 2.7.15rc1, pytest-3.5.1, py-1.7.0, pluggy-0.6.0
rootdir: /home/naved/hil-all/brocade-default-vlan-fix, inifile: setup.cfg
plugins: requests-mock-1.5.2, xdist-1.24.1, forked-1.0.1, cov-2.6.0
collected 3 items                                                                                                                          

tests/deployment/multi_networks.py .                                                                                                 [  0%]
tests/deployment/native_networks.py .                                                                                                [  0%]

======================================================== 3 passed in 119.12 seconds ========================================================
```

knikolla:
Is the travis-ci broken?

naved001:
not sure why it's not starting on cci-moc/hil, but on my fork I see this: https://travis-ci.org/github/naved001/hil/jobs/743249119 
we haven't been maintaining this repository for a while now and things have become outdated.

knikolla:
Do you think it is worth investigating? I am okay with merging this without a proper CI if you have run the tests locally.

naved001:
I am gonna go ahead and merge then since the failures are unrelated to the changes I made. 

I did run the deployment tests locally from an old virtualenv so I didn't run into the issues that travis-ci is having.

I am gonna open an issue to know that the CI needs to be fixed, in case we ever come back to HIL.

